PR #40, opened 2016-07-19
DOCS: Add Clear Container with Docker install doc

grahamwhaley:
Add installation notes for installing Docker and Clear Containers into
a Clear Linux distribution.

Note: the ReST style 'attention' boxes do not show as boxes on
githubs markdown web rendering, but show fine when using rst2html

### DIFF
--- a/INSTALL_DOCKER.rst
+++ b/INSTALL_DOCKER.rst
@@ -1,0 +1,259 @@
+.. contents::
+.. sectnum::
+
+``Installing Clear Containers 2.0 into Clear Linux``
+====================================================
+
+Introduction
+------------
+.. attention::
+   This document is a work in progress - the **Attention!** boxes note places where things may be updated as infrastructure develops.
+
+`Clear Containers`_ 2.0 provides an Open Containers Initiative (OCI_) compatible 'runtime' and is installable into Docker_ 1.12-rc4 and later, where OCI_ runtime support is available.
+
+This document details how to install Docker-1.12-rc4 and the necessary parts of `Clear Containers`_  into a `Clear Linux`_ distribution.
+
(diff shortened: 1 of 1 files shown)


INSTALL_DOCKER.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Download, build and install the clr-oci-runtime from source:
+
+  ::
+
+    cor_source=${HOME}/clr-oci-runtime
+    git clone https://github.com/01org/clr-oci-runtime.git $cor_source
+    cd $cor_source
+    autoreconf -fvi
+    bash autogen.sh --disable-cppcheck --disable-valgrind
```
grahamwhaley: I think we leave it as is - and then we'll have to work out the clash for the users who have installed from source.

INSTALL_DOCKER.rst
```diff
@@ -0,0 +1,259 @@
...
+
+      which clr-oci-runtime
+      #typically /usr/bin/clr-oci-runtime
+
+Then edit the Docker_ systemd unit file ExecStart to make `Clear Containers`_ the default runtime.
+
+  ::
+
+    Edit: /usr/lib/systemd/system/docker-upstream.service
+    ExecStart=/usr/bin/dockerd-upstream --add-runtime cor=/usr/bin/clr-oci-runtime--default-runtime=cor -H fd://
```
grahamwhaley: 'simply'? :-)  I think it is just personal preference. For me I actually prefer the 'Edit:' version, as it is clearer to me that we are doing an edit and what it is adding.

jodh-intel: Fair point :)

I've now fixed this in master, but note the missing space on line 202 before "--default-runtime=cor".

grahamwhaley: Just so we get this clear - the action is to *add* arguments to the existing ExecStart line, yes? In which case the text needs to be a little clearer. Are you OK clarifying that @jodh-intel ?

PR #46, opened 2016-07-22
configure.ac: add libnetlink check from iproute2

miguelinux:
Check for libnetlink header and library in configure.ac.
This library is shipped by iproute2 and is needed to
compile network module.

Signed-off-by: Miguel Bernal Marin <miguel.bernal.marin@linux.intel.com>


mcastelino:
@miguelinux thank you

PR #52, opened 2016-07-27
DOCS: Add iproute-dev dependency to build cc-oci-runtime

gorozco1:
* Add a dependency needed to build cc-oci-runtime: iproute2-dev

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -220,6 +220,7 @@ installed on your system:
 - glib
 - json-glib
 - uuid
+- iproute2
 
 Configure Stage
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


sameo:
@gorozco1 Could you explain what brings that dependency in ? Is that for the networking code ?

README.rst
```diff
@@ -220,6 +220,7 @@ installed on your system:
 - glib
 - json-glib
 - uuid
+- iproute-dev
```
jcvenegas: For consistency, you may want to use  iproute instead iproute-dev, above is specified that are development versions

gorozco1:
Just updated the commit @sameo

jcvenegas:
+1

PR #53, opened 2016-07-27
simplify code

devimc:


jcvenegas:
+1, this change depends on network patch, right ?  Otherwise current network hooks will fail and docker cc container  wont  start ?

amshinde:
+1

devimc:
@jcvenegas nop  , CC are working  :)

jcvenegas:
@devimc  great no blockers, thanks for clarify

src/command.c
```diff
@@ -179,8 +190,10 @@ handle_command_stop (const struct subcommand *sub,
 	cc_oci_node_dump(root);
 #endif /*DEBUG*/
 
-	g_node_children_foreach(root, G_TRAVERSE_ALL,
-	    (GNodeForeachFunc)process_config_stop, (gpointer)config);
+	if (! cc_oci_process_config(root, config, stop_spec_handlers)) {
```
sameo: Instead of open coding a tree traversing routine, the typical way of tracking errors with a GLib `*_foreach()` routine is to create a:
```
struct config_cookie {
          gboolean failed
          struct cc_oci_config *config
}
}
```

And passing a `config_cookie` pointer to `g_node_children_foreach()`. Your `GNodeForEachFunc` callback is a wrapper around `process_config_stop()` that only sets `config_cookie->failed` to true when `process_config_stop()` fails.

sameo:
@devimc Please be more specific as to what this PR does. Which part of the code does it simplify ?
Also, it actually does more than simplifying code.

Finally, you're pushing 3 unrelated commits through the same PR. One wrong commit will block the whole PR, so you want to actually send 3 separate PRs.

sameo:
@devimc Regarding the commit messages, they're missing some further details, and the `Fixes` keyword when fixing an exisiting issue.

For example, ` Do not ignore values returned by spec handlers` could be:

```
OCI config parsing should fail when one spec handler fails

When parsing the OCI config file, each section is handled by a specific callback.
We currently do not report any section parsing error meaning we could potentially
start running on top of an invalid OCI config file.

This patch forces a runtime exit whenever we failed to parse an OCI config file section.

Fixes #25

```

grahamwhaley:
Hi @devimc , I see you sent this PR from your master branch. I'd recommend using 'topic branches' for on-going work so we can have multiple in flight, and it should be easier for you to track which work has been merged or not etc.

devimc:
@sameo  @grahamwhaley   I'm agree

PR #57, opened 2016-07-28
Simplify hypervisor tag processing code by tabulating data

devimc:
This patch reduces code lines and makes code simpler and readable.
This patch won't decrease the calls to cc_oci_replace_string.

### DIFF
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -147,6 +147,23 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
 
 	procsock_device = g_strdup_printf ("socket,id=procsock,path=%s,server,nowait", config->state.procsock_path);
 
+	struct special_tag {
+		const gchar* name;
+		const gchar* value;
+	} special_tags[] = {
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Nice cleanup. Can we maybe make the subject line give more context, for example:
  'Simplify hypervisor tag processing code by tablulating data'
or similar - as currently it does not convey why this is a good thing or necessary

grahamwhaley:
Thanks for the update.
@sameo , look good to you? If so, I'll merge.
Whilst here, just a minor nit - if/when we touch this file again then it might be nice to sort the entries in the table alphabetically, just for some extra clarity - unless there is some implicit special ordering magic going on there, in which case it should be commented :-)

sameo:
@grahamwhaley Yes, it looks good. Please go ahead and merge it.

grahamwhaley:
Heh, I see a typo 'tab[l]ulating' in the subject line - my typo copied from my comment. I'll see if I can do a trivial fix to that, and merge anyway.

Issue #17, opened 2016-07-18
No output with docker when run non-interactively


jodh-intel:
If the runtime is configured for docker and `docker run` is specified without `-ti`, no output is currently generated to either the console or the docker log (`docker logs` command).

```
$ docker-upstream run --runtime runc ubuntu echo "hello world"
hello world

$ docker-upstream logs  $(docker-upstream ps -qa | head -1)
hello world

$ docker-upstream run --runtime cor ubuntu echo "hello world"
container 20ed5890250df32cd7c1b24027bcba708c6b350481557205fc881a16e924b27d created

$ docker-upstream logs  $(docker-upstream ps -qa | head -1)
container 20ed5890250df32cd7c1b24027bcba708c6b350481557205fc881a16e924b27d created

$ docker-upstream run -it --runtime runc ubuntu echo "hello world"
hello world

$ docker-upstream logs  $(docker-upstream ps -qa | head -1)
hello world

$ docker-upstream run -it --runtime cor ubuntu echo "hello world"
hello world
```

amshinde:
Docker creates name pipes init-std* under /run/docker/libcontainerd/{container} dir. These pipes are used for reading the container output. containerd-shim creates unnamed pipes (https://github.com/docker/containerd/blob/ef23e7041b27a773da6dfaf6483f018e48c40d79/containerd-shim/process.go#L311)  to pipe the IO from from the runtime to the shim. (https://github.com/docker/containerd/blob/ef23e7041b27a773da6dfaf6483f018e48c40d79/containerd-shim/process.go#L173). The shim copies the output from the unnamed pipes to the named ones for the output to be available for docker.
Runc simply dups the  standard IO when run in non-interactive mode.  The solution for our runtime is to  pipe the output from the qemu process to stdout of the parent process.  This should solve the non-interactive output as well as fix docker attach.

grahamwhaley:
I've not read the code, but your analysis sounds good to me.

devimc:
@amshinde  :+1:

amshinde:
Passed  'stdio,id=charconsole0,signal=off'  to qemu as its virtual console device for qemu to use the stdio inherited from the parent. Following suggestion from devimc, added a delay before  the parent exits. The output from qemu  is shown as long as the parent is alive. Working on a fix.

amshinde:
[...]

Issue #34, opened 2016-07-19
Redirect stdin/out/err to the containerd provided pseudoterminal


sameo:
Rather than programatically redirecting stdin/out/err for the qemu-lite process, we could run it with `-chardev serial path=/dev/pts/XX` where that device is the value passed from containerd.

PR #59, opened 2016-07-28
Fix for non-interactive docker output and docker attach

amshinde:
This is a workaround for using stdio as console device for qemu.
This solution bypasses the shim pipes and writes to docker named
pipes. We should resort to using stdio as console once that is resolved.

### DIFF
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -23,6 +23,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <stdio.h>
+#include <string.h>
+#include <errno.h>
 
 #include <glib.h>
 #include <uuid/uuid.h>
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Hi. Is this related to #17 and/or #34 ?  If so, let's consider referencing them in the commit message (whilst I contemplate if that is the right thing to do for github commits...)

amshinde:
grahamwhaley This is related #17

sameo:
@amshinde @grahamwhaley If it fixes #17, it should appear in the commit message:

```
Fixes #17
```

And github will close the issue with a reference to the commit automatically.

amshinde:
sameo, this solution works. But I'd like to keep the #17 open until the stdio problem with qemu is figured out

devimc:
:+1:

src/hypervisor.c
```diff
@@ -124,23 +126,76 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
...
+				g_critical("Failed to create symlink for input pipe: %s",
+					   strerror (errno));
+				goto out;
+			}
+
+			console_device = g_strdup_printf ("pipe,id=charconsole0,path=%s", config->console);
+
+		} else {
+
+			/* In case the runtime is called standalone without console
```
grahamwhaley: nit pick - this could be a single line comment I think

src/hypervisor.c
```diff
@@ -124,23 +126,76 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
...
+			*/
+
+			/* No console specified, so make the hypervisor create
```
grahamwhaley: nit pick - this comment and the one following it are in different 'styles' - with the alignment/indentation of the '*'s differing. We should stick to one style (normally the style already in use in the file/project).

grahamwhaley:
Hi @amshinde . I just had a couple of comment style nitpicks - if you can fix them up trivially then I'll merge this.
 thanks!

amshinde:
grahamwhaley, I pushed the changes you suggested.

PR #60, opened 2016-07-28
cc-oci-runtime should fail when a spec handler fails

### LINKED ISSUE #49: Do not ignore values returned by spec handlers
```process_config_start``` ignores values returned by spec handlers therefore spec handler's validations are ignored.
If a spec handler returns ```false``` cor must not continue and report the error

### PR DESCRIPTION
**Pull request by devimc**
Currently we have 2 different functions to process
a config using spec handlers: process_config_stop
and process_config_start. This patch merges these 2
functions and returns false when a spec handler fails.

With this patch when a spec handler fails cc-oci-runtime
will report an error and will exit.

fixes #49

### CODE REVIEW

**Comment by sameo** [comment]
@devimc Have you considered implementing [my comment](https://github.com/01org/cc-oci-runtime/pull/53/commits/7f64775a0bbdcee3765e0965568ae4fe046031fe#r72545872) ?

**Comment by devimc** [comment]
@sameo  yes, but we would still have 2 different function to do the same (process configure) and using your approach when a spec handler fails the loop won't stop until end, I think that when a spec handler fails the error must be reported and the runtime must exit

File: src/command.c [hunk 1]
```diff
@@ -18,13 +18,14 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+#include <glib/gstdio.h>
```
sameo: This is an unrelated change, please don't move the header as part of this patch.

src/command.c
```diff
@@ -303,7 +316,7 @@ handle_default_usage (int argc, char *argv[],
 		     (!g_strcmp0 (argv[0], "-h")))) {
 		help = true;
 	}
-				
+
```
sameo: Unrelated whitespace cleanup, please remove from this patch.

src/command.c
```diff
@@ -54,6 +55,16 @@ struct subcommand *subcommands[] =
 	NULL
 };
 
+/**
+ * List of spec handlers used to process config on stop
+ */
+static struct spec_handler* stop_spec_handlers[] = {
```
sameo: It should be `struct spec_handler *stop_spec_handlers[]`

src/oci-config.c
```diff
@@ -159,3 +159,48 @@ cc_oci_config_free (struct cc_oci_config *config)
...
+/*!
+ * find and call the spec handler for each child of GNode
+ *
+ * \param [in] node GNode
+ * \param[in,out] config cc_oci_config struct
+ *
+ * \return \c false if a spec handler fails, else \c true
+ */
+gboolean
+cc_oci_process_config (GNode* root, struct cc_oci_config* config,
```
sameo: The expected code style is `struct foo *a` and not `struct foo* a`. For having a consistent format across the code, I'd appreciate if you could fix this.

src/oci-config.c
```diff
@@ -159,3 +159,48 @@ cc_oci_config_free (struct cc_oci_config *config)
...
+	struct spec_handler** spec_handlers)
+{
+	GNode* node;
+
+	for (node=g_node_first_child(root); node; node=g_node_next_sibling(node)) {
```
sameo: `node = g_node_first_child(root)` and not `node=g_node_first_child(root)`.

src/oci-config.h
```diff
@@ -1,18 +1,18 @@
...
  * as published by the Free Software Foundation; either version 2
  * of the License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
```
sameo: Please remove this serie of whitespace cleanup, it adds noise to the patch and makes it more difficult to read/review.

src/spec_handler.h
```diff
@@ -1,18 +1,18 @@
...
  * as published by the Free Software Foundation; either version 2
  * of the License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
```
sameo: Same here regarding the whitespace cleanups.

sameo:
@devimc I understand your patch is more efficient than using `g_node_children_foreach()` but it open codes something that's implemented in GLib, i.e. an K-ary tree traversal routine. That's my point.

That being said, I like the fact that you're factorizing 2 functions into a single one. So please go ahead and fix the few style nitpicks that I spotted and we'll merge that patch.

sameo:
This one looks good to me now.
@gorozco1 Are you ok with this patch as well ?

gorozco1:
+1

PR #62, opened 2016-07-29
Redirect stdio to the containerd provided pseudoterminal for console …

### LINKED ISSUE #34: Redirect stdin/out/err to the containerd provided pseudoterminal

amshinde:
…case.

Fixes #34. Rather than programatically redirecting stdin/out/err for the qemu
process, run it with -chardev serial path=/dev/pts/XX where that device
is the value passed from containerd.

### DIFF
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -196,7 +196,7 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
 				config->console);
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


devimc:
nice work   :+1:

PR #63, opened 2016-07-29
README: Remove limitation as a workaround has been pushed.

amshinde:


gorozco1:
Why is this being removed? the issue 17 is still open

amshinde:
I wanted to keep it open to track the stdio issue for qemu. Meanwhile I have pushed a workaround. (I need to check if it will make getting the exit status from the container difficult which hasnt been working anyways. I havent looked at it yet.)

With the workaround merged, we no longer have the limitation and can see the correct output in docker logs.

gorozco1:
Thanks for the explaining. 
+1

devimc:
:+1:

PR #65, opened 2016-08-01
travis: resolved dependencies to build cc-oci-runtime

chavafg:
travis: resolved dependencies to build cc-oci-runtime

needed to build glib, json-glib and libcheck from sources
still have some issues with unit tests that need to investigate
and resolve before adding those checks to travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,39 +5,8 @@ language: c
 compiler: gcc
 
 before_script:
-- sudo apt-get -qq install valgrind lcov uuid-dev pkg-config zlib1g-dev libffi-dev gettext libpcre3-dev texinfo gtk-doc-tools cppcheck
-- mkdir cor-dependencies
-- pushd cor-dependencies
-- curl -L -O https://github.com/GNOME/glib/archive/2.49.4.tar.gz
-- tar -xvf 2.49.4.tar.gz
-- pushd glib-2.49.4
-- bash autogen.sh --prefix=/usr
-- make
-- sudo make install
-- popd
-- curl -L -O https://github.com/GNOME/json-glib/archive/1.2.2.tar.gz
-- tar -xvf 1.2.2.tar.gz
-- pushd json-glib-1.2.2
-- bash autogen.sh --prefix=/usr
-- make
-- sudo make install
-- popd
-- curl -L -O https://github.com/libcheck/check/archive/0.10.0.tar.gz
-- tar -xvf 0.10.0.tar.gz
-- pushd check-0.10.0
-- autoreconf --install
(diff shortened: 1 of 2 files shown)


grahamwhaley:
Nice!
 I can see that download and build is our only option as we need newer packages than the base distro provides. That 'download and build' every time niggles me though - from both the 'resource' and time point of view. I wonder if we are able to cache any of that, as discussed here: https://docs.travis-ci.com/user/caching/
Browsing that, sadly it looks like caching the downloads or package installs is not really a benefit - but for the compiles maybe the 'ccache' is worth investigating?

.travis.yml
```diff
@@ -1,5 +1,43 @@
...
 compiler: gcc
 
+before_script:
+- sudo apt-get -qq install valgrind lcov uuid-dev pkg-config zlib1g-dev libffi-dev gettext libpcre3-dev texinfo gtk-doc-tools cppcheck
+- mkdir cor-dependencies
+- pushd cor-dependencies
+- curl -L -O https://github.com/GNOME/glib/archive/2.49.4.tar.gz
+- tar -xvf 2.49.4.tar.gz
+- pushd glib-2.49.4
+- bash autogen.sh --prefix=/usr
```
jodh-intel: This is great, but I'd prefer that we put this scripting into a separate file (say `./setup-travis.sh` or something). 

That way, we can add appropriate comments to explain what we're doing and why and we can also use variables to avoid all the hard-coded versions numbers. Travis does support environment variables, but I think a shell script is still going to be more familiar and flexible than taking the Travis approach.

For example, we could do something like this (although a shell function would be even better):

```
glib_version=2.49.4
curl -L -O "https://github.com/GNOME/glib/archive/${glib_version}.tar.gz"
tar -xvf "${glib_version}.tar.gz"
pushd "glib-${glib_version}"
```

chavafg: Hi, 

I looked at the `ccache` feature and it seems that it is worth only in the case we recompile (make clean; make). So maybe it will not benefit us. 
Good idea about the `.travis-setup.sh` file,  I will change it and resubmit.

grahamwhaley:
OK. I was hoping the ccache line would cache the ccache files from the first time we ever build glib, json-glib and libcheck (actually when we shut the first travis instance down I believe), and then next time we invoke a travis build it will restore the ccache data before doing the download/build - so all subsequent builds will use the ccache data where they can, speeding up the builds. We don't need to do this, and we can definitely start off not doing this. If our travis build times ever become onerous then we could re-visit the idea.

jodh-intel:
LGTM - thanks!

PR #80, opened 2016-08-05
WIP: RFC: Enable exec

jcvenegas:
This patch adds basic functionality for exec command.

- 3 sub command options are added: --console, --process --,detach, --pid-file (options used by docker exec).

- The ip address is saved in statefile (ip address is required to execute commands by ssh).
- Require a container with network (docker run --net=none wont work)

Current implementation does not work with docker exec (need to debug more pipes), but directly by cc-oci-runtime. To test current implmentation is required a mini-os image with openssh-server bundle (clear-containers.img +9610) with ssh key access (ssh-copy-id) and root (permitrootlogin yes).

```
docker-upstream run -ti debian sh
cc-oci-runtime exec $CONTAINER_ID bash
```

### DIFF
--- a/src/command.h
+++ b/src/command.h
@@ -74,6 +74,7 @@ struct start_data {
 	gchar *bundle;
 	gchar *console;
 	gchar *pid_file;
+	gchar *process_json;
 	gboolean detach;
 	gboolean dry_run_mode;
 };
--- a/src/commands/exec.c
+++ b/src/commands/exec.c
@@ -65,7 +65,7 @@ handler_exec (const struct subcommand *sub,
 	g_assert (config);
 
 	if (handle_default_usage (argc, argv, sub->name,
-				&ret, 2, "<cmd> [args]")) {
+				&ret, 1, "<cmd> [args]")) {
 		return ret;
 	}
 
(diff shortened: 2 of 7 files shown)


dlespiau:
So, qemu seems to have a daemon one can run in the guest to do a bunch of commands. One of those commands is guest-exec:

http://git.qemu.org/?p=qemu.git;a=blob;f=qga/qapi-schema.json;h=c21f3084dcafb9039e9139d9d169d268a3806bb1;hb=HEAD#l1009

It seems to me that using qagent in the guest image is a good long term plan, but I'm still new to this virtualisation thingy. On top of guest-exec, some advanced features like freeze/migration need an agent in the quest.

mcastelino:
@jcvenegas  @dlespiau The qemu daemon is preferable to the ssh option. When you run containers with overlay networks, it is possible that there will be no way to reach into the container over ssh from the local host. So an approach that does not need local network access into the container is preferable.

jcvenegas:
I am closing this PR in favor of qemu daemon.  I will send only options patch .. and  tests qemu-ga in a clear-containers image

PR #85, opened 2016-08-06
[OCI] Support running workload as a different user/group

### LINKED ISSUE #20: the user and group settings in config.json are not honoured.
Specifically, `process.user.uid`, `process.user.gid` and `process.additionalGids` are ignored.

See:
-  `clr_oci_create_container_workload()` in [src/oci.c](src/oci.c).
- https://github.com/opencontainers/runtime-spec/blob/master/config.md#user.

amshinde:
The uid/gid passed to the runtime was being ignored. The workload was
being run as root user and group.Use the gid/uid in config.json to lookup
username/groupname and pass that to the workload to run with appropriate
user permissions.

Fixes #20

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -30,6 +30,8 @@
 #include <stdbool.h>
 #include <errno.h>
 #include <sys/types.h>
+#include <stdlib.h>
+
 
 #include <glib.h>
 #include <glib/gstdio.h>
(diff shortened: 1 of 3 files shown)


src/oci.c
```diff
@@ -30,6 +30,8 @@
 #include <stdbool.h>
 #include <errno.h>
 #include <sys/types.h>
+#include <stdlib.h>
+
```
grahamwhaley: extra empty line added?

src/oci.c
```diff
@@ -516,13 +518,178 @@ cc_oci_attach (struct cc_oci_config *config,
...
+	size_t     len = 0;
+	gchar      *eptr;
+	gboolean   done = false;
+
+	fp = g_fopen (password_path, "r");
+	if ( fp == NULL) {
+		return NULL;
+	}
+
+	while ( getline (&line , &len, fp) != -1) {
```
grahamwhaley: Are we able to use fgetpwent() and related functions here?

src/oci.c
```diff
@@ -516,13 +518,178 @@ cc_oci_attach (struct cc_oci_config *config,
...
+	size_t     len = 0;
+	gboolean   done = false;
+	gchar      *eptr;
+
+	fp = fopen (group_path, "r");
+	if ( fp == NULL) {
+		return NULL;
+	}
+
+	while ( getline (&line , &len, fp) != -1) {
```
grahamwhaley: Similar to pw - could we use getgrent() and family here?

src/oci.c
```diff
@@ -516,13 +518,178 @@ cc_oci_attach (struct cc_oci_config *config,
...
+lookup_user_name(gchar *password_path, uid_t userid)
+{
+	FILE       *fp = NULL;
+	uid_t      uid;
+	gchar      *matched_uname = NULL;
+	char       *line = NULL;
+	size_t     len = 0;
+	gchar      *eptr;
+	gboolean   done = false;
+
```
jodh-intel: Missing test on password_path:

```

if (! password_path) {
  return NULL;
}
```

src/oci.c
```diff
@@ -516,13 +518,178 @@ cc_oci_attach (struct cc_oci_config *config,
...
+lookup_group_name(gchar *group_path, gid_t gid_to_match)
+{
+	FILE       *fp;
+	gid_t      gid;
+	gchar      *matched_gname = NULL;
+	gchar      *line = NULL;
+	size_t     len = 0;
+	gboolean   done = false;
+	gchar      *eptr;
+
```
jodh-intel: Again, missing check on group_path param.

src/oci.c
```diff
@@ -516,13 +518,178 @@ cc_oci_attach (struct cc_oci_config *config,
...
+ */
+
+private void
+lookup_user_and_group_name(struct cc_oci_config *config)
+{
+	gchar   *password_path = NULL;
+	gchar   *group_path = NULL;
+	gchar   *username;
+	gchar   *groupname;
+
```
jodh-intel: Missing check on `config` arg.

src/oci.c
```diff
@@ -538,6 +705,11 @@ cc_oci_create_container_workload (struct cc_oci_config *config)
 	g_autofree gchar   *cwd = NULL;
 	gboolean           ret = false;
 	gchar             **args = NULL;
+	gchar             *quoted_workload_cmd = NULL;
```
jodh-intel: These (and other gchars in this patch) could also be "g_autofree gchar *foo = NULL" to simplify the code (no need to call `g_free()` / `g_free_if_set()`).

src/oci.c
```diff
@@ -595,6 +769,14 @@ cc_oci_create_container_workload (struct cc_oci_config *config)
 		goto out;
 	}
 
+	quoted_workload_cmd = g_shell_quote (workload_cmdline);
+	sg_command = g_strdup_printf ("sg %s -c %s",
```
jodh-intel: The bad news is that the `sg` command is not always available. For example, the popular `busybox` docker image only provides `su`.

We have 2 main options here:

- Require `sg` if `gid != 0` and we can't find the `sg` command, fail the request and thus fail the container start.
  That's a safe approach but not very friendly since the user won't know if we can honour their request until the last moment.

- Create a static binary (must be static) that we make available in the workload chroot and have that handle uid, gid and cwd changes before having it exec the required workload. I already have code to do this fwiw so will send that through. This overall approach isn't idea either, but may be the best pragmatic solution since:
  - it does not require the workload image contain particular userspace binaries.
  - it guarantees that we can always honour the request.

amshinde:
Closing this PR. Going with another approach.

PR #87, opened 2016-08-08
A few small fixes

dlespiau:
A few fixes I did while starting with the runtime, hopefully nothing to contentious.


grahamwhaley:
lgtm. Thanks for clearing up the rc4 bits - I'll take that off my list now ;-)  I'll leave @jodh-intel to check the autoconf bits (but to my untrained eye they looked sensible)

configure.ac
```diff
@@ -225,3 +230,32 @@ AC_ARG_WITH([cc-image],
...
 
 AC_OUTPUT
+
+dnl === Summary ===============================================================
+
+AS_IF([test x"$disable_tests" = xno], [enable_tests=yes], [enable_tests=no])
+AS_IF([test x"$disable_functional_test" = xno],
+      [enable_functional_tests=yes], [enable_functional_tests=no])
+
+echo ""
```
jodh-intel: Rather than raw `echo` calls, I'd use `AC_MSG_RESULT`.

For example, https://github.com/jamesodhunt/procenv/blob/master/configure.ac#L212.

dlespiau:
Pushed an updated version using AC_OUTPUT()

jodh-intel:
LGTM.

PR #89, opened 2016-08-08
build: Bump the glib requirement to 2.46

dlespiau:
When testing the build with the minimum requirements we declare, I
discovered G_FILE_MONITOR_WATCH_MOVES wasn't properly documented and was
introduced during the 2.45 cycle. The minimum requirement for glib is
then 2.46.

The glib documentation patch is on its way:

  https://bugzilla.gnome.org/show_bug.cgi?id=769630


grahamwhaley:
lgtm
Thanks for chasing this upstream as well!

jodh-intel:
LGTM.

PR #91, opened 2016-08-08
Travis setup improvements

dlespiau:
I looked a bit at travis, here's the first resulting series: various small cleanups and shaves 30s-ish of the build.

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -21,14 +21,13 @@ set -e -x
 
 glib_version=2.46.2
 json_glib_version=1.2.2
-check_version=0.10.0
 
 gnome_dl=https://download.gnome.org/sources
 
 # Install required dependencies to build
 # glib, json-glib, check and cc-oci-runtime
 sudo apt-get -qq install valgrind lcov uuid-dev pkg-config \
-  zlib1g-dev libffi-dev gettext libpcre3-dev cppcheck
+  zlib1g-dev libffi-dev gettext libpcre3-dev cppcheck check
 
 mkdir cor-dependencies
 pushd cor-dependencies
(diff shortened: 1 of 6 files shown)


.travis-setup.sh
```diff
@@ -17,42 +17,49 @@
 #  along with this program; if not, write to the Free Software
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-glib_version=2.49.4
+set -e -x
+
+glib_version=2.46.2
```
jodh-intel: I'm not sure we want to do this - the minimum version we support is 2.46, but we don't have to build with the absolute minimum do we? Or are we doing so to guarantee that the version in `configure.ac` is still correct? If the latter, we need a comment.

Also, 2.49.4 seems to be the dev version, and 2.46.2 is an old version - how about 2.48.1 (latest stable)? :-)

dlespiau: The intention was that we do not introduce the use of newer symbols/defines without the mandatory pkg-config check in the configure.ac.

It's too easy to use a new function and not bump the dependency in the configure.ac and travis can help us catch that.

jodh-intel: Ok, so if this is a double-check on `configure.ac`, we should auto-generate `.travis-setup.sh` from `.travis-setup.sh.in` using the value of `required_glib_version` to ensure we only define this version once.

dlespiau: we could do something like that. Actually we need to grep configure.ac for the versions as .travis-setup.sh is run before autogen.sh.

However, we'll also catch when someone bumps the version number but forgets to update .travis-setup.sh, so maybe it can wait a bit longer? I'd like to do it once we've also applied the treatment you've done to glib with json-glib and libcheck (so we can retrieve the version numbers in a single way for all 3 deps).

jodh-intel: grepping `confiigure`.ac sounds a bit fragile - maybe we need a `versions.txt` file or something that just contains lines like:

```
required_glib_version=2.46
```

So that we could source that into `configure.ac` and `.travis-setup.sh`.

dlespiau:
Changes:

- Added a new commit to remove the compilation of libcheck now that we can just use the system one (since the version requirement was dropped)
- Implemented the suggestion to have only one place for the glib requirement.

versions.txt
```diff
@@ -0,0 +1 @@
+required_glib_version=2.46
```
jodh-intel: It's probably also worth adding the following from `.travis-setup.sh`:

```
json_glib_version
check_version
```

Also, `versions.txt` will need to be added to `EXTRA_DIST` in `Makefile.am` to ensure it gets included in any release tarballs and can thus still be sourced by `configure.ac`.

dlespiau:
New version addresses the review comments, but for the check version as we don't install check manually anymore with this series.

jodh-intel:
LGTM.

sameo:
Looks good to me.

chavafg:
LGTM

PR #98, opened 2016-08-09
Add travis email notifications

jodh-intel:


grahamwhaley:
lgtm. Would like a +1 from someone else versed in Travis ... @sameo or @dlespiau  ?

dlespiau:
That looks right quickly checking with the documentation. Just a note that we don't need to specify the on_success/on_failure values as they are the default ones, not that it matters.

jodh-intel:
Yep - I added the default values explicitly to ensure we remain in control if Travis decides to change the defaults on the server-side.

PR #99, opened 2016-08-09
Disable global log

### LINKED ISSUE #42: global log is currently enabled by default
This is useful for debug/diagnostics since all runtime operations are logged to `/run/opencontainer/containers/cc-oci-runtime.log`. However, on a busy server that file could grow fast.

We should probably

- Disable the global log by default and document how to configure docker to pass
   "`--global-log /run/opencontainer/containers/cc-oci-runtime.log`" to `containerd`.
   See: [`setup_logging()`](/01org/cc-oci-runtime/blob/master/src/main.c#L245).
- Update `README.rst` on the new behaviour (no global log by default if run stand-alone).

jodh-intel:
Fixes #42.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -42,7 +42,7 @@ src/commands/cc_oci_runtime-version.$(OBJEXT): \
 	 commit_id
 
 data/cc-oci-runtime.sh: data/cc-oci-runtime.sh.in
-	$(SED) \
+	$(AM_V_GEN)$(SED) \
 		-e 's|[@]PACKAGE_NAME[@]|$(PACKAGE_NAME)|g' \
 		-e 's|[@]bindir[@]|$(bindir)|g' \
 		"$<" > "$@"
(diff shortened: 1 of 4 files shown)


grahamwhaley:
lgtm

Makefile.am
```diff
@@ -41,13 +41,20 @@ STORED_COMMIT = $(shell cat commit_id 2>/dev/null)
 src/commands/cc_oci_runtime-version.$(OBJEXT): \
 	 commit_id
 
+data/cc-oci-runtime.sh: data/cc-oci-runtime.sh.in
+	$(SED) \
```
dlespiau: Maybe AM_V_GEN?

dlespiau:
lgtm.

I think I might try to use the docker config file myself instead, and declare multiple configuration (one 'cc' and one 'cc-debug'). Hopefully that'll work as well.

PR #100, opened 2016-08-09
Tests: Fix failures caused by running hypervisor_test.c under mock(1).

jodh-intel:
mock(1) sets up stdin as a non-terminal device which triggers the runtime
to change its behaviour slightly (see commit 7899f891). This causes the
hypervisor tests to fail under mock, but not when run interactively.

### DIFF
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -146,6 +146,11 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
 		goto out;
 	}
 
+	if (! config->bundle_path) {
+		g_critical ("No bundle path");
+		goto out;
+	}
+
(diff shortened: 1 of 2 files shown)


gorozco1:
+1

src/hypervisor.c
```diff
@@ -146,6 +146,11 @@ cc_oci_expand_cmdline (struct cc_oci_config *config,
 		goto out;
 	}
 
+	if (! config->bundle_path) {
```
grahamwhaley: nitpick: Did you sneak this bundle_path check in at the same time as the tty fix? Just it confused me a little as the only change to this file.

devimc:
:+1:

amshinde:
+1

PR #102, opened 2016-08-09
build: Use AM_V_GEN when generating the commit_id file

dlespiau:
Nicer builds are all the rage.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,7 @@ CHECK_DEPS =
 GIT_COMMIT := $(shell git rev-parse HEAD 2>/dev/null)
 
 commit_id:
-	if [ "x$(GIT_COMMIT)" != "x" ]; then \
+	$(AM_V_GEN)if [ "x$(GIT_COMMIT)" != "x" ]; then \
 		echo "Writing commit id"; \
 		echo $(GIT_COMMIT) > commit_id; \
 	else \


dlespiau:
The builds have started failing with:

configure.ac:23: error: Autoconf version 2.69 or higher is required

Weird, as we didn't see that before... and we did test all the PRs

chavafg:
I see that travis went back to ubuntu 12.04 instead of 14.04

dlespiau:
ah! now, let's find out why

grahamwhaley:
good spot @chavafg

dlespiau:
Seems like we managed to catch a window where travis hit a regression and didn't honour the dist variable. Just re-triggered the build and it's now green.

jodh-intel:
LGTM, although the commit message doesn't contain a subsystem in the first line and there isn't really an explanation of the change.

jodh-intel:
Ah - apologies @dlespiau. Regarding my previous comment, I was looking at the comments section, not the commit header log entry.

grahamwhaley:
@dlespiau , amusing though the commit message is :-), I too think we can do a little better with a description/explanation. Could you expand that a bit please, and then we'll merge.  Thanks!

dlespiau:
Fair enough, updated the commit message.

grahamwhaley:
Great, thanks :-)

PR #106, opened 2016-08-10
Miscellaneous fixes

jodh-intel:


src/oci.c
```diff
@@ -241,20 +241,27 @@ cc_oci_kill (struct cc_oci_config *config,
 		config->state.status = last_status;
```
grahamwhaley: It's a shame we can't make the 'kill' block fall out to the error: code as well, as then we'd have a nice clean exit path for the whole func - but I think the extra config file update step makes a reflow awkward eh?

jodh-intel: Yes, that's why. Annoying :(

dlespiau:
lgtm

Issue #69, opened 2016-08-04
Enable 'make check' within Travis


grahamwhaley:
Travis is now doing a build check, but we can enhance that to add a run check with 'make check'.

dlespiau:
#PR 109 is step towards that and runs make check in travis.

However the following things still need consideration:

- The valgrind tests aren't run for some reason

- It doesn't seem possible to use KVM within travis (googled around and asked @sameo that has ran into the same issue with ciao). However, we might be happy enough with launching qemu without KVM support, as we're really only testing our interfaces, not KVM itself (or so I believe). In that case we'd need to:
    - compile qemu-lite
    - fetch a clear image and container kernel
    - prepare an OCI bundle in tests/functional/bundle
    - make check should then run the functional tests

jodh-intel:
This seems like a duplicate of #255 where I investigated running the full test suite under Travis. In summary, we don't think this is possible given the lack of VT-x support in the Travis environment.

Valgrind *is* configured to run, but valgrind's `helgrind` and `drd` tools are explicitly disabled. For an explanation see commit 614fb442cba6102f6de7ca44dd02725c7ebca083. However, there seems to be a
problem running valgrind under Travis (see issue #295).

dlespiau:
ok, it seems we have more precise issues describing the remaining problem(s), let's close that one, we do run make check in travis after all.

PR #109, opened 2016-08-10
Enable make check (unit tests only) and check for warnings in travis

dlespiau:
Another step towards better CI:

- Get rid of warnings and make sure we don't introduce any later on
- Run make check

What's still to do is documented in https://github.com/01org/cc-oci-runtime/issues/69

### DIFF
--- a/src/commands/create.c
+++ b/src/commands/create.c
@@ -22,6 +22,10 @@
 
 extern struct start_data start_data;
 
+/* ignore -pedantic to cast handle_option_console, a function pointer, to a
+ * void* */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
 static GOptionEntry options_create[] =
 {
 	{
@@ -52,6 +56,7 @@ static GOptionEntry options_create[] =
 
 	{NULL}
 };
+#pragma GCC diagnostic pop
 
 static gboolean
 handler_create (const struct subcommand *sub,
--- a/tests/hypervisor_test.c
+++ b/tests/hypervisor_test.c
@@ -524,12 +524,12 @@ START_TEST(test_cc_oci_vm_args_get) {
 	g_strfreev (args);
 
 	/* recreate the args file */
-	ret = g_file_set_contents (args_file, "hello\# comment\n", -1, NULL);
+	ret = g_file_set_contents (args_file, "hello\\# comment\n", -1, NULL);
 	ck_assert (ret);
 
 	ck_assert (cc_oci_vm_args_get (&config, &args));
 
-	ck_assert (! g_strcmp0 (args[0], "hello\# comment"));
+	ck_assert (! g_strcmp0 (args[0], "hello\\# comment"));
 	ck_assert (! args[1]);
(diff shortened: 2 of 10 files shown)


src/commands/create.c
```diff
@@ -22,6 +22,10 @@
 
 extern struct start_data start_data;
 
+/* ignore -pedantic to cast handle_option_console, a function pointer, to a
+ * void* */
+#pragma GCC diagnostic push
```
grahamwhaley: I'd like to hear @jodh-intel opinion on the pragmas. Great that you've nullified the warnings, but in-code GCC specific pragmas are not generally to my liking :-)  If James is OK, then I think we'll probably be fine.

tests/hypervisor_test.c
```diff
@@ -524,12 +524,12 @@ START_TEST(test_cc_oci_vm_args_get) {
 	g_strfreev (args);
 
 	/* recreate the args file */
-	ret = g_file_set_contents (args_file, "hello\# comment\n", -1, NULL);
+	ret = g_file_set_contents (args_file, "hello\\# comment\n", -1, NULL);
```
grahamwhaley: Good catch - but then this hints to me that the test was not really working or maybe testing what it should... @devimc can you have a think about that?

dlespiau:
Don't really want to open the pandora box, but I'd drop pedantic for -Wall and either all or some of the -Wextra warnings. I have yet to find an actually bug -pedantic helped catching and considering we're only targeting linux and x86_64, seems overkill to me.

grahamwhaley:
Nice series, thanks for doing that. Overall it looks good to me. Just a couple of minor comments.

src/commands/create.c
jodh-intel: @grahamwhaley - I agree, not generally a fan of pragmas, but aside from changing `-std=`, I don't know how we can suppress these warnings.

Other thoughts on pragmas / options to suppress:

- will static analysis tools understand them?
- will other compilers understand them?

jodh-intel:
To be clear, I'm +1 too. We might want to revisit the use of `-pedantic` at some point, but for now the ability to have Travis running the tests is more important IMHO. Added to which the pragma is commented and they are easy to `grep` :)

grahamwhaley:
OK, I was about to merge this and let @devimc pick up on the \\# test item above separately - but we now have merge conflicts. @dlespiau , if you could fix that then we'll merge.  Thanks.

dlespiau:
A few changes:

- Rebased on top of master
- Added a couple more commits to compile warning free with clang
- Changed the pragma to ignore "-Wpedantic" instead of "-pedantic" as the latter wasn't working with clang (I usually include a change log in the commit message)

src/commands/create.c
dlespiau: I guess the answers to those questions will have to come with time. Worse than can happen is being ignored though, maybe with a warning.

The good news is that it's now also working with clang.

dlespiau:
Ok, passing again. This time it was and interaction with the parallel build (-j5) that just got added, patch has the details.

Issue #103, opened 2016-08-09
Document runtime code design in DESIGN.rst


jodh-intel:

jodh-intel:
See #122.

jodh-intel:
#122 is now merged, so closing.

PR #120, opened 2016-08-13
Resource leak fix and some code cleanup

amshinde:
The title should be more descriptive.. IMHO

### DIFF
--- a/src/logging.c
+++ b/src/logging.c
@@ -536,25 +536,21 @@ void cc_oci_setup_hypervisor_logs (struct cc_oci_config *config)
 		g_autofree gchar* std_file_path = g_build_path ("/", hypervisor_log_dir,
 			std_file_name, NULL);
 
-		if (std_file_path) {
-			/* creating log file
-			 * i.e: $hypervisor_log_dir/$containerId-hypervidor.stdout
-			 */
-			int std_file_fd = g_creat(std_file_path, CC_OCI_LOGFILE_MODE);
+		/* creating log file
+		 * i.e: $hypervisor_log_dir/$containerId-hypervidor.stdout
+		 */
+		int std_file_fd = g_creat(std_file_path, CC_OCI_LOGFILE_MODE);
 
-			if (std_file_fd < 0) {
-				g_critical("failed to create file: %s", std_file_path);
-			} else {
-				/* redirecting stdout/stderr to a file */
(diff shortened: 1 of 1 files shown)


jodh-intel:
These two commits should be prefixed by the subsystem first ("logging").

grahamwhaley:
I agree. @amshinde , can you prefix with the subsystem, and also could you expand the PR message a bit more please - 'Fixes' is too generic and short.
@jodh-intel , once the text is correct, how is the code? It lgtm

jodh-intel:
@grahamwhaley - yep, code lgtm.

dlespiau:
Missing the s-o-b tags as well.

For error paths, using the Linux style ones should really be the only way I'm afraid: Chapter 7 of https://www.kernel.org/doc/Documentation/CodingStyle.

grahamwhaley:
Oh, good catch @dlespiau  I knew there would come a day when I didn't spot one of those.
It is also referenced in our https://github.com/01org/cc-oci-runtime/blob/master/CONTRIBUTING.md#certificate-of-origin

grahamwhaley:
And on error paths - yes, that did occur to me when reading the patches. @jodh-intel , do you have thoughts on this - always use the goto:error method for any depth of nest?

jodh-intel:
I'd personally have written this without the `else { ... }` - just test for error, and if found call `g_critical()`, then immediately return. That code is then simpler and hence easier to understand, but also less indented in the non-error case.

Certainly, for >1 error branches, we should prefer `goto error` to handle cleanup and avoid duplicated code. I'm adding that to #103.

amshinde:
grahamwhaley: Missed out on the signed off tag. I have added it to my git alias now so I don't forget :)

grahamwhaley:
Hi. I think these lgtm enough now, thanks for the cleanups.
@amshinde , can you ack that you took note of the exit path error handling recommendations from @dlespiau for the future? :-)

https://www.kernel.org/doc/Documentation/CodingStyle.
http://eli.thegreenplace.net/2009/04/27/using-goto-for-error-handling-in-c

thanks!

amshinde:
@grahamwhaley yes,  took note of it :)

amshinde:
@dlespiau any comments on the reworked patch?

dlespiau:
oh, sorry, I didn't realise it has changed. lgtm.

PR #121, opened 2016-08-15
Make helper script read from cfg file

jodh-intel:


grahamwhaley:
Hi. Some niggles:
 - I'm not a fan of wrapping the commit subject line into the body with '...' - if you can find a way to make those more succinct that would be nice :-)
 - In the middle commit, typo in body - s/that/than/
 - and the Travis failed - I'll leave that one with you.

jodh-intel:
Note that the README update does not add a new section for `cc-oci-runtime.sh.cfg` since it is _hoped_ that Docker will soon provide a daemon option to pass arguments to the runtime (`containerd` already has the facility).

jodh-intel:
Branch updated.

(I'm not sure what you were referring to wrt `...` - that the subject line was too long? I've fixed that. I don't think the commit body text flowed from the subject line though?)

grahamwhaley:
lgtm. Thanks, and Jarvis is happy now :-)  Of course, I can't go re-check the subject line thing after the push. It could be a github view anomaly - if I see something again I'll try to capture it.

PR #126, opened 2016-08-15
DOCS: Improvements to the INSTALL_DOCKER

grahamwhaley:
A range of small improvements to the INSTALL_DOCKER.rst file.
 - remove un-used doc links
 - remove the remaining Attention box
 - add a missing 'sudo'
 - bump the out of date CLR version number
 - Add example outputs for the docker commands
 - clean up 'root' and # promts
 - remove newgrp usage for logout instead

### DIFF
--- a/INSTALL_DOCKER.rst
+++ b/INSTALL_DOCKER.rst
@@ -93,12 +93,7 @@ To enable your user to access both Docker and KVM you will need to add them to t
 
     $ sudo usermod -G kvm,docker -a <USERNAME>
 
-Then run the following commands to add those group ids to your active login session:
-
-  ::
-
-    $ newgrp kvm
-    $ newgrp docker
+For these new group memberships to take effect, log out and log back in again as \<USERNAME\>.
 
 Docker on Clear Linux
 ~~~~~~~~~~~~~~~~~~~~~
--- a/INSTALL_DOCKER.rst
+++ b/INSTALL_DOCKER.rst
@@ -140,15 +140,25 @@ Before we dive into using `Clear Containers`_ it is prudent to do a final sanity
   ::
 
     $ docker ps
+    CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
     $ docker network ls
+    NETWORK ID          NAME                DRIVER              SCOPE
+    028f36f79ca5        bridge              bridge              local
+    8a22faead310        host                host                local
+    960c24b9d9de        none                null                local
     $ docker pull busybox 
-
+    Using default tag: latest
+    latest: Pulling from library/busybox
+    8ddc19f16526: Pull complete 
+    Digest: sha256:a59906e33509d14c036c8678d687bd4eec81ed7c4b8ce907b888c607f6a1e0e6
+    Status: Downloaded newer image for busybox:latest
     $ docker run -it busybox sh
     [    0.063356] systemd[1]: Failed to initialise default hostname
     / # uname -a
      Linux f0098e68456f 4.5.0-49.container #1 SMP Mon Aug 8 20:46:42 UTC 2016 x86_64 GNU/Linux
     / # exit
 
+In the final step above it can be seen that we are running a Clear Container in a hypervisor, as the Linux kernel version 4.5.0-49.container is different from the kernel being used by the underlying host system.
 
 Conclusion
 ----------
--- a/INSTALL_DOCKER.rst
+++ b/INSTALL_DOCKER.rst
@@ -6,9 +6,6 @@
 
 Introduction
 ------------
-.. attention::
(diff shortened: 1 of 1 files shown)


INSTALL_DOCKER.rst
```diff
@@ -35,15 +32,15 @@ If your user does not already have ``sudo`` rights, you should add your user to
 
   ::
 
-    usermod -G wheel -a <USERNAME>
+    $ usermod -G wheel -a <USERNAME>
 
 This change will take effect once you have logged out as root and logged back in as <USERNAME>.
```
jodh-intel: We could refer to `newgrp` here too for consistency with the 2nd `usermod` (`$ sudo usermod -G kvm,docker -a <USERNAME>`)?

jodh-intel:
One nit, but not a blocker. lgtm.

INSTALL_DOCKER.rst
grahamwhaley: Actually @jodh-intel , as you mention it ;-), I was going to add a note or Issue saying that I'm not sure the 'newgrp' is the right way to do this (below) - each newgrp, at least on my F24 box, starts a new sub-shell, so at the end of the whole sequence I need to 'exit' 3 times to get out of my original shell - not great. I was going to ask @gorozco1 if he had any other ideas here - apart from log-out and in again. I don't remember if there is a nice way to re-initialise a login shell etc. Worst case is I may add a note explaining what the `newgrp` is doing.
Oh - and as to the above - I wondered about adding `newgrp` below the above - *but* - the above is done at a root login, and then it asks you to log out and then back in as USERNAME - so you don't need a `newgrp` here I think.

jodh-intel: right. Maybe it would be less confusing all round to remove references to newgrp and just go for the logout/in approach.

INSTALL_DOCKER.rst
```diff
@@ -32,15 +32,15 @@ If your user does not already have ``sudo`` rights, you should add your user to
 
   ::
 
-    usermod -G wheel -a <USERNAME>
+    $ usermod -G wheel -a <USERNAME>
```
gorozco1: this is a command that needs to be done with sudo...

grahamwhaley: not here I think - the instructions note you should be logged in as root I believe?

jodh-intel: +1 on @gorozco1's comment on choice of prompt though. We should use `$` for non-priv commands and `#` for priv ones.

grahamwhaley: Ah, apologies @gorozco1 - my eyes didn't pick up the `#` - yes, good spot - will edit and re-PR. Thanks!

INSTALL_DOCKER.rst
```diff
@@ -35,15 +32,15 @@ If your user does not already have ``sudo`` rights, you should add your user to
...
 
 This change will take effect once you have logged out as root and logged back in as <USERNAME>.
 
 And you will also need to add your user or group to the ``/etc/sudoers`` file, for example:
 
   ::
 
-    visudo
+    $ visudo
```
dlespiau: If we are going to use '#' for root, this one needs a change as well.

dlespiau:
lgtm

grahamwhaley:
OK, new patches pushed to fix '#' for root and reword the 'newgrp' section. Please review. thx.

INSTALL_DOCKER.rst
```diff
@@ -93,7 +93,7 @@ To enable your user to access both Docker and KVM you will need to add them to t
 
     $ sudo usermod -G kvm,docker -a <USERNAME>
 
-Then run the following commands to add those group ids to your active login session:
+The easiest way to get these new group memberships to take effect is to log out and log back in again as \<USERNAME\>. Alternatively, you can run the following commands to add those group ids to your active login session, but be aware each of these commands drops you into a new sub-shell, so you need two extra `exits` at the end of your session to leave your shell:
 
   ::
```
gorozco1: Wouldn't be better to remove completely newgrp commands and just tell the user logout and login again wiht <USERNAME> so the changes take effect?

That way we avoid confusions with the end user... (two ways to do the same)

grahamwhaley: @gorozco1 I'm more than happy to. It will make it simpler, and I had considered already. Anybody else disagree - if not I'll get on it later...

grahamwhaley:
Done. One extra commit to make use of ``root`` more consistent as well.

gorozco1:
+1

grahamwhaley:
Going to merge this - I know it is my own PR - but it has done the review cycle to clean up previous feedback, and some of the maintainers are away/travelling.

PR #132, opened 2016-08-19
docs: containerd seems to do the "start" and "delete" calls

dlespiau:
A close look at containerd(-shim) seems to indicate it's only
responsible for the "create" command, leaving the other ones to
containerd.

shim:
  containerd-shim/process.go	->	create, exec

containerd:
  runtime/process.go		->	start,
  runtime/container.go		->	delete, state, ps, events, ...

To be thorough if anyone wants to double check the above:
containerd-shim/process.go also includes a "delete" call, but it's only
used if "create" fails, so not the normal mode of operation.

I also modified the life time of the shim accordingly, it exits right
after reaping the qemu process, leaving the delete for containerd.


gorozco1:
It seems that containerd-shim: -> create, delete 

not Exec... could someone else validate this?

dlespiau:
So I guess you don't agree with the 4th paragraph of the commit message :)

"delete" is only called in an error path, when "create" is failing: https://github.com/docker/containerd/blob/master/containerd-shim/main.go#L87

In the normal course of actions, containerd-shim reaps the qemu process and exits without calling "delete": https://github.com/docker/containerd/blob/master/containerd-shim/main.go#L105
There is a comment telling exactly what's stated in the commit message "Let containerd take care of calling the runtime delete"/

dlespiau:
As for the "exec" path: https://github.com/docker/containerd/blob/master/containerd-shim/process.go#L130

"exec" also needs the shim to stay around and reap the process exit status.

jodh-intel:
I've confirmed this by observation:

```
containerd-shim    create
containerd-shim    exec
containerd         start
containerd         delete
```

+1.

gorozco1:
Thanks @dlespiau  +1

I guess we can merge this PR then  :)

PR #145, opened 2016-08-19
output: Remove the spurious 'container created' message

### LINKED ISSUE #142: Spurious 'container created' message when running with the cc runtime
```
$ sudo docker run debian true
$ sudo docker run --runtime cc debian true
container 9cf769a030698323d01b9200b1b98c4d554fbe16435d3a6a0127484d7f18a120 created
$
```

amshinde:
Fixes #142

### DIFF
--- a/src/commands/create.c
+++ b/src/commands/create.c
@@ -63,8 +63,6 @@ handler_create (const struct subcommand *sub,
 		struct cc_oci_config *config,
 		int argc, char *argv[])
 {
-	gboolean  ret;
-
 	g_assert (sub);
 	g_assert (config);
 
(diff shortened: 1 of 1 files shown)


dlespiau:
Oh, excellent! lgtm.

dlespiau:
A small not on the subject line prefix, it's usual to put the subsystem in which the change is made, not the "what". In this case 'create:' would be more adequate I think (the change is in the create command).

dlespiau:
Another comment is that we really should try to build up our in tree test suite, especially when fixing bugs, to make sure we won't regress. I think there's a hidden test suite somewhere that may even use docker and I'd love for it to be in tree so developers can also run it easily.

To not delay that PR with that test task (it's be some significant amount of work to add docker based tests to this tree and maybe people will have different opinions about that one), I've created #146.

amshinde:
Sounds good dlespiau. I'll change the subsystem name and resend.

grahamwhaley:
lgtm.
We could (have) changed the PR prefix from 'output' as well, and a little more explanation on why we are changing this as it is different from the 'de-factor runc' (remember, the git commit history is what ultimately counts, not the text in the github issues ;-) But those are not going to stop me merging.

PR #149, opened 2016-08-22
Add pullapprove support

jodh-intel:


dlespiau:
I think PRs shouldn't have merge commits in them but be rebased against master and force pushed instead.

dlespiau:
otherwise, lgtm!

jodh-intel:
Re-pushed with rebase.

grahamwhaley:
lgtm - but oddly seems to have failed the CLANG build. I'm guessing that is a spurious fail, but could you check that for us? (if it is spurious, then that could get annoying...)
I also presume pullapprove does not come into action until we have registered it and let it create its hooks - which I'm also guessing you will do after the merge :-)

jodh-intel:
Yep - the clang error was transient. I've raised #155 to capture that problem. And yes, I'll register the project with pullapprove after this merge...

PR #156, opened 2016-08-23
Process: Fix pullapprove config.

jodh-intel:
The pullapprove.com "approve_regex" value must be a single-quoted
string.


dlespiau:
Still not passing :) sorry for not having caught it. I guess you could also have enabled travis for your fork and published a branch (that's how I test my PRs beforehand).

dlespiau:
as discussed on IRC, let's merge it a bit against the process so we can make progress on this.

jodh-intel:
I think I've fixed the `approve_regex` in `.pullapprove.yml`, but the pullapprove webhook above is still failing since it's applying the version of `.pullapprove.yml` from the `master` branch, *not* the fixed version from y branch.

PR #176, opened 2016-08-26
mount: fix mount when using regular files

### LINKED ISSUE #23: volumes: volume creation and mount fails on clear linux
On clear linux I am unable to launch a container with an attached volume.
This failure is present both with runc and cors

docker-upstream run -it --runtime runc --name webxx -v /webappxx training/webapp python app.py
docker-upstream run -it runc --name webyy -v /webappyy training/webapp python app.py

Both these fail with the error
docker-upstream: An error occurred trying to connect: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create?name=webxx: EOF.

The same command works on ubuntu

jcvenegas:
Fixes #23 
docker -v allows to mount regular files, but cc-oci-runtime assumes that a mount will be a directory.
This patch verify if a mount is done using a regular file. If true it creates a file in the destination target with the same permissions than the original file, this new file is used as mount destination file.

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -21,6 +21,7 @@
 #include <string.h>
 #include <errno.h>
 #include <sys/stat.h>
+#include <fcntl.h>
 
 #include "mount.h"
 #include "common.h"
@@ -134,6 +135,7 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
 		"and flags 0x%lx%s%s";
 
 	int ret;
+	struct stat st;
 
 	if (! m) {
 		return false;
@@ -151,6 +153,22 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
 
 	if (dry_run) {
 		return true;
+	}
+
+	if (stat (m->mnt.mnt_fsname, &st) < 0) {
+		g_critical ("Unable to handle mount file:"
+				"getting file status  %s (%s)",
+				m->mnt.mnt_fsname, strerror (errno));
+		return false;
+	}
+
+	if (S_ISREG(st.st_mode)) {
+		if(creat (m->dest, st.st_mode) < 0 ) {
+			g_critical ("Unable to handle mount file:"
+					"creating file %s (%s)",
+					m->dest, strerror (errno));
+			return false;
+		}
 	}
 
 	ret = mount (m->mnt.mnt_fsname,


src/mount.c
```diff
@@ -152,6 +153,17 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
...
 	}
+	if (g_file_test (m->mnt.mnt_fsname, G_FILE_TEST_IS_REGULAR)) {
+		struct stat st;
+		ret = stat (m->mnt.mnt_fsname, &st);
+		if (! ret) {
+			return false;
+		}
+		ret = creat (m->dest, st.st_mode);
+		if (! ret) {
+			return false;
```
devimc: Maybe you should add some debug log

src/mount.c
```diff
@@ -152,6 +153,17 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
 	if (dry_run) {
 		return true;
 	}
+	if (g_file_test (m->mnt.mnt_fsname, G_FILE_TEST_IS_REGULAR)) {
+		struct stat st;
+		ret = stat (m->mnt.mnt_fsname, &st);
```
jodh-intel: You are essentially calling `stat()` twice here: once explicitly and once via `g_file_test()`.

Since you need to `st.st_mode`, Please can you drop the call to `g_file_test()` and just call `stat()` once instead. You can then perform the check using ` S_ISREG(st.st_mode)`.

src/mount.c
```diff
@@ -152,6 +153,17 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
...
+		if (! ret) {
+			return false;
+		}
+		ret = creat (m->dest, st.st_mode);
```
jodh-intel: I was going to say that there needs to be a corresponding `remove()` call in `cc_oci_perform_unmount()`, but I see that we don't do that for directories either. Hence, this PR is consistent with the existing code.

(I've created issue #178 to handle general mount cleanup).

src/mount.c
jcvenegas: @devimc added messages  as critical

src/mount.c
jcvenegas: @jodh-intel  good catch, thanks! , fixed now

src/mount.c
```diff
@@ -153,6 +155,18 @@ cc_oci_perform_mount (const struct cc_oci_mount *m, gboolean dry_run)
...
 	}
 
+	if (stat (m->mnt.mnt_fsname, &st) < 0) {
+		g_critical ("mount file %s does not exist", m->mnt.mnt_fsname);
+		return false;
+	}
+
+	if (S_ISREG(st.st_mode)) {
+		if(creat (m->dest, st.st_mode) < 0 ) {
+			g_critical ("failed to create mount file: %s", m->dest);
```
jodh-intel: Could you add a `%s` for `strerror (errno)` so the actual `errno` gets logged? You could do that for the `stat()` too. Maybe something like:

```
g_critical ("Unable to handle mount file %s: %s", m->mnt.mnt_fsname, sterror (errno));
```

jcvenegas: sure ... adding errno

jcvenegas: done

jodh-intel:
lgtm.

gorozco1:
lgtm

gorozco1:
lgtm

PR #179, opened 2016-08-30
Always show qemu pclite support

jodh-intel:


jcvenegas:
lgtm

sameo:
LGTM

dlespiau:
if --with-qemu-path is given, but doesn't support pc-lite, we're missing the has_pc_lite=no in that code path.

PR #181, opened 2016-08-30
Tests: Functional tests for docker integration

GabyCT:
These are Basic Acceptance Tests for the docker integration. They run some docker commands.

### DIFF
--- a/tests/integration/docker/test.common
+++ b/tests/integration/docker/test.common
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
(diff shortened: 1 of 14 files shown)


jodh-intel:
Hi @GabyCT - please could you add the standard header to these new files (see https://github.com/01org/cc-oci-runtime/pull/200 :-)

tests/integration/docker/test.common
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+DOCKER_EXE="docker"
+DOCKER_SERVICE="docker-cor"
```
jcvenegas: this service name is default in all packages where cc-oci is packaged ?

jodh-intel: good catch! That should only be `docker-cor` for clr systems.

tests/integration/docker/test.common
```diff
@@ -0,0 +1,26 @@
...
+function startDockerService(){
+	systemctl status "$DOCKER_SERVICE" | grep 'running'
+	if [ "$?" -eq 0 ]; then
+		systemctl restart "$DOCKER_SERVICE"
+	fi
+}
+
+#Checking that default runtime is cor
+function runtimeDocker(){
+    "$DOCKER_EXE" info | grep "Default Runtime: cor"
```
jcvenegas: same as docker service, this runtime id that is registered in dockerd is the same in all obs packages ?

tests/integration/docker/test.common
GabyCT: I reviewed the documentation for other distributions and clearlinux is the only one that is using docker-cor i.e. #systemctl restart docker-cor so one way to solve this is to try to figure out if you are using clearlinux or other Linux distribution. Make a distinction in order to do #systemctl restart docker-cor (for clear) or #systemctl restart docker (other distributions). Also, I verified that when you are running a container in other Linux distribution and you performed a #docker info,  it appears as Default runtime: cor the same as the clearlinux case so we do not have issues in that.

tests/integration/docker/test.common
GabyCT: Yes, we verified an other distributions are using Default runtime: cor.

gorozco1:
lgtm

jodh-intel:
lgtm

PR #202, opened 2016-08-31
Logging: Fix "std_file_fd" fd leak.

jodh-intel:
This fd needs to be closed on success and failure; the former because
it has been dup(2)'d, and the latter because the dup(2) failed but the
fd is still open from the original creat(2).


sameo:
LGTM

jcvenegas:
lgtm

PR #203, opened 2016-08-31
Networking: Move network discovery and configuration to child

mcastelino:
Move the network configuration and the discovery process to
the child. The parent runs the hooks and on completion of the hooks
notifies the child of the hook execution status. The child will launch
the VM if the hooks were executed successfully, or will exit.

This reduces the complexity and the latency in the network setup.

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -452,290 +452,6 @@ cc_oci_vm_netcfg_get (struct cc_oci_config *config,
 }
 
 /*!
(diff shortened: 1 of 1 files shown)


jodh-intel:
LGTM (and love that diffstat: "+36 −344" :-)

sameo:
Same here. LGTM.

PR #204, opened 2016-08-31
Networking: Fix missing break in prefix logic

mcastelino:
Fix missing break in case statement which causes IPv6 prefix
generation failure.

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -363,8 +363,13 @@ subnet_to_prefix(const gint family, const void *const sin_addr) {
 	switch(family) {
 	case AF_INET6:
 		pfix = prefix(addr, 16);
+		break;
 	case AF_INET:
 		pfix = prefix(addr, 4);
+		break;
+	default:
+		g_warning("invalid prefix family %d", family);
+		pfix = 0;
 	}
 
 	return g_strdup_printf("%d", pfix);


gorozco1:
lgtm

src/networking.c
```diff
@@ -363,8 +363,14 @@ subnet_to_prefix(const gint family, const void *const sin_addr) {
 	switch(family) {
 	case AF_INET6:
 		pfix = prefix(addr, 16);
+		break;
 	case AF_INET:
 		pfix = prefix(addr, 4);
+		break;
+	default:
+		g_debug("invalid prefix family %d", family);
```
amshinde: Since this is an error path, would make sense to change this to g_warning

src/networking.c
```diff
@@ -363,8 +363,14 @@ subnet_to_prefix(const gint family, const void *const sin_addr) {
...
+		pfix = 0;
+		break;
```
amshinde: Nitpicking here. This break is not required.

gorozco1: True! good catch

mcastelino: The linux kernel coding style allows for break here and in some cases is useful.

mcastelino:
@gorozco1 @amshinde  updated

jcvenegas:
lgtm

src/networking.c
amshinde: Good to know that.

amshinde:
lgtm

Issue #208, opened 2016-09-01
configure should not resolve real paths for vm.json  variables


jcvenegas:
configure resolve paths for clear containers components :
--with-qemu-path=
--with-cc-kernel=
--with-cc-image=
this files can be symlinks and  it is useful to update components without reconfigure vm.json, so this is not required.

jcvenegas:
Closed by #209

PR #209, opened 2016-09-01
Config: configure must not resolve kernel and container image paths.

jodh-intel:
To avoid having to regenerate the "vm.json" configuration file,
"--with-cc-image=" is often provided with a symlink path [1]. This allows
other mechanisms to update the image and "re-point" the symlink to the
latest image, which avoids having to regenerate "vm.json".

However, configure was resolving kernel and container paths, meaning
that the runtime would not be using the latest image (as specified by
the symlink) and would fail if the release hard-coded in "vm.json" was
removed when it became stale.

Note that the runtime full resolves all paths so that it can log
full details of the artifacts being used for a particular container.

--

[1] - The same logic applies to "--with-cc-kernel=" but note that
     "--with-qemu-path=" is not resolved.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -206,7 +206,7 @@ AC_ARG_WITH([cc-kernel],
         [clear container kernel path])
     ],
     [
-        CONTAINER_KERNEL=$(realpath $with_cc_kernel)
+        CONTAINER_KERNEL=$with_cc_kernel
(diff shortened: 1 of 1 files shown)


jodh-intel:
Fixes #208.

amshinde:
lgtm

jcvenegas:
lgtm

PR #211, opened 2016-09-01
oci: Use process.terminal to check for interactive runs

amshinde:
Currently we check if stdin refers to a terminal using isatty()
Use the process.terminal value from config.json instead.
Docker sets this value to true when run interactively.

Fixes 183


jodh-intel:
lgtm.

sameo:
LGTM

PR #213, opened 2016-09-02
Version: Ensure "--version" and "version" display a commit id.

jodh-intel:
git-archive(1), as used by github.com allows files to be modified
"on-the-fly" as the archive is generated. Hence, configure git-archive
to write the commit id to file "commit_id", which allows the runtime to
display a commit id even when not built from either a release archive
or from a git tree.


grahamwhaley:
lgtm
Nice hack^H^H find there @jodh-intel 
Looks like we failed the Travis - a quick peek shows it might be the bats 'version' test that failed, which could make sense. You want to have a look?

devimc:
:+1:

gorozco1:
lgtm

PR #217, opened 2016-09-06
Process: Fix running standalone.

jodh-intel:
An unintended side-effect of commit 26559c5 was that it broke running
standalone. Although running stand-alone still requires "config.json" to
specify "process.terminal=true", cc_oci_start() was not waiting for the
hypervisor.

This minimal fix [1] is better than what we had before 26559c5 since it
moves the heuristic to the generic cc_oci_config_update() rather than
hard-coding it in cc_oci_start().

---

[1] - An alternative solution would be to add "process.terminal"
      to the state file.

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -1719,6 +1719,11 @@ cc_oci_config_update (struct cc_oci_config *config,
 
 	config->use_socket_console = state->use_socket_console;
 
+	if (config->console && ! config->use_socket_console && isatty (STDIN_FILENO)) {
+		g_debug ("enabling terminal for standalone mode");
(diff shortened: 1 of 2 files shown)


grahamwhaley:
lgtm
I would have been happy with the state file process.terminal fix as well, as that would have kept more in line with what we expect from Docker.

jodh-intel:
I've raised #218 raised to provide a fuller fix, but for now, we need to get standalone mode working again to re-enable the functional tests.

dlespiau:
lgtm. Also we need more testing in travis with soft qemu sooner rather than later :)

Issue #221, opened 2016-09-06
cc-oci-runtime broken after merging PR 215


chavafg:
Unable to launch docker with cc-oci-runtime. 
```
root@containers-ci ~/cc-oci-runtime # docker run -ti ubuntu true
docker: Error response from daemon: containerd: container not started.
```
Seems that commit 3d8911d152dd7541a59c0e97005eeb27ad1b3df0  is the one that broke the package.

cc-oci-runtime logs:
```
2016-09-06T14:02:22.683372Z:12385:cc-oci-runtime:debug:finished namespace setup
2016-09-06T14:02:22.683565Z:12385:cc-oci-runtime:debug:child pid is 12387
2016-09-06T14:02:22.683789Z:12385:cc-oci-runtime:debug:created state file /run/opencontainer/containers/76b5db18410b8975c66afb18bd746e0c084456ff4ed5454ad751614eca0dec62/state.json
2016-09-06T14:02:22.683875Z:12385:cc-oci-runtime:debug:running hook command:
2016-09-06T14:02:22.683919Z:12385:cc-oci-runtime:debug:arg: '/usr/bin/dockerd'
2016-09-06T14:02:22.683957Z:12385:cc-oci-runtime:debug:arg: 'libnetwork-setkey'
2016-09-06T14:02:22.683993Z:12385:cc-oci-runtime:debug:arg: '76b5db18410b8975c66afb18bd746e0c084456ff4ed5454ad751614eca0dec62'
2016-09-06T14:02:22.684034Z:12385:cc-oci-runtime:debug:arg: '513f4e71af9722bc131c31cb6afc624d16241892f38c10bdf2877a15c0b13e8c'
2016-09-06T14:02:22.684563Z:12385:cc-oci-runtime:debug:hook process ('/usr/bin/dockerd') running with pid 12388
2016-09-06T14:02:22.697530Z:12385:cc-oci-runtime:warning:/usr/bin/dockerd: "/usr/bin/dockerd" requires 0 arguments.

2016-09-06T14:02:22.697603Z:12385:cc-oci-runtime:warning:

2016-09-06T14:02:22.697630Z:12385:cc-oci-runtime:warning:Usage: dockerd [OPTIONS]

2016-09-06T14:02:22.698135Z:12385:cc-oci-runtime:debug:Hook pid 12388 ended with exit status 256
2016-09-06T14:02:22.698162Z:12385:cc-oci-runtime:critical:hook process 12388 failed with exit code: 256
[...]

Issue #214, opened 2016-09-02
hook with no arguments fails to run


jodh-intel:
If debug is enabled, the log will contain:

```
A NULL argv[0] was passed through an exec system call.
```

jodh-intel:
The fix for this issue simple - remove the `G_SPAWN_FILE_AND_ARGV_ZERO` flag. However, in adding new tests to `tests/process_test.c` I hit a problem - valgrind's helgrind tool fails. After fairly extensive debugging, it seems that helgrind is happy with glibs thread handling code. Even though the runtime code is not multi-threaded, calling `g_child_watch_add()` ends up calling 
`g_get_worker_context()`, which create a `GThread`. If the call to `g_child_watch_add()` is removed, helgrind is happy. Paring it back even further, I now have a simple glib program that calls `g_thread_new()`. Even without a main loop or a call to `g_thread_join()`, helgrind complains about "`Possible data race`s". But helgrind is happy with a simple corresponding program that just calls `pthread_create()`.

jodh-intel:
Until we can fully diagnose the helgrind/glib issue, we cannot run the new test code. Hence, I've disabled it but added a compiler warning with a pointer to this bug so we don't lose track of the issue.

jodh-intel:
Attaching test programs for pthread and gthread including instructions on how to run helgrind on the binaries.

[cc-oci-runtime-bug-214.zip](https://github.com/01org/cc-oci-runtime/files/454868/cc-oci-runtime-bug-214.zip)

jodh-intel:
Converted the `#warning` into `g_message()` calls to avoid Travis barfing (due to `-Werror`).

jodh-intel:
Fixed on #223.

PR #223, opened 2016-09-07
Process: Re-take on fixing running hooks with no arguments.

jodh-intel:
The fix introduced on 3d8911d since although it supported no-arg hooks,
it did not support running hooks where the argv[0] passed to the hook
was different to the hook path. This is used by docker to handle network
setup where dockerd is exec'd as "libnetwork-setkey".

The correct fix is to re-introduce the use of
G_SPAWN_FILE_AND_ARGV_ZERO and handle the "no-arg" case specially (along
with adding better comments around arg handling and the use of
G_SPAWN_FILE_AND_ARGV_ZERO).

Updated tests to reflect re-introduction of G_SPAWN_FILE_AND_ARGV_ZERO.
Note that these tests are still commented out since they fail under
valgrind's helgrind tool. See github issue #214.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -305,12 +305,32 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 		return false;
 	}
 
+	flags |= G_SPAWN_DO_NOT_REAP_CHILD;
+	flags |= G_SPAWN_CLOEXEC_PIPES;
+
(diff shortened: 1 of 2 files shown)


jodh-intel:
Fixes #221.

grahamwhaley:
lgtm , with a glance over the patch.

dlespiau:
lgtm

PR #228, opened 2016-09-08
Config: Remove bug report email address.

jodh-intel:
The address specified in "configure.ac" is not that useful given that
there is now a mailing list that could be used instead, as specified in
CONTRIBUTING.md. However, that document already explains how we want
bug reports to be handled (github issues) so a bug report address is
redundant.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@
 #
 
 AC_PREREQ([2.69])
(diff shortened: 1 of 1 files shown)


grahamwhaley:
lgtm

dlespiau:
lgtm

PR #229, opened 2016-09-08
Arguments: Fail if privilege checks fail.

jodh-intel:
handle_arguments() was calling cc_oci_get_priv_level() but not returning
false in the failure scenario as it should have.


grahamwhaley:
lgtm

dlespiau:
lgtm

PR #231, opened 2016-09-08
Mount: leave rootfs as it was found

### LINKED ISSUE #178: mount code should leave bundle rootfs as it was found
`cc_oci_handle_mounts()` currently creates the mount destination by calling `g_mkdir_with_parents()`. That works fine, but the problem is the runtime does not know if the destination existed before the call to `g_mkdir_with_parents()`. Further, #23 means that the destination could in fact be a regular file, not a directory.

The rootfs should be left in its original state on container deletion, so the mount code needs to:

1. identify if the destination directory existed prior to the call to `cc_oci_handle_mounts()`.
1. add information to the state flag specifying if the destination already existed.
1. delete the destination if it already existed.

The state file needs to by updated by modifying `cc_oci_mounts_to_json()` so that rather than creating an array of strings (the full path to the mount destination), it creates an array of objects which specify:

```
destination: "/full/path/to/mount/destination"
existed: true|false
```

(We don't need to add the type of the destination to the state file as we can derive that from the existing destination).

`tests/mount_test.c` also needs to be updated for the new statefile mount data and format.

devimc:
with this patch directories created by cc-oci-runtime
when directories or files are mounted will be deleted
once umount is done

fixes #178

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -23,6 +23,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 
+#include <glib/gstdio.h>
+
 #include "mount.h"
 #include "common.h"
 
@@ -99,6 +101,7 @@ cc_oci_mount_free (struct cc_oci_mount *m)
 	g_free_if_set (m->mnt.mnt_dir);
 	g_free_if_set (m->mnt.mnt_type);
 	g_free_if_set (m->mnt.mnt_opts);
+	g_free_if_set (m->directory_created);
 
 	g_free (m);
 }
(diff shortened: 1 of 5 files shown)


chavafg:
test: qa ok

chavafg:
test: qa ok

chavafg:
qa-passed

chavafg:
qa-passed

jodh-intel:
Hi @devimc - the code looks good, but please can you add some tests for the new functionality (atleast to `tests/state_test.c` and ideally also to `tests/mount_test.c`).

devimc:
Hi @jodh-intel  I updated json files to test the code that was added :)

jodh-intel:
Fair enough - I see that we don't already have explicit tests on the mounts, so have raised #248 :)

lgtm.

chavafg:
qa-passed

PR #233, opened 2016-09-08
docs: match README with code implementation

### LINKED ISSUE #232: runtime command line option -d not parsed after [command]
Placing the '-d' option after the [command] on the command line can produce an 'Unknown option -d' error.

According to (implied by) the command line help, and the README.md file, you should be able to place the '-d' debug enable option anywhere on the command line. In reality, if we take the example from the running standalone section of the README.md:

`$ sudo ./cc-oci-runtime --log "$logfile" --log-format json create --bundle "$bundle_dir" --console $(tty) -d --pid-file "$pidfile" "$name"`

this produces the log output of:

`/usr/bin/cc-oci-runtime: create: Unknown option -d`

Thus, it looks like the '-d' is maybe being treated as a sub-command of 'create' rather than a standalone command. It feels like maybe a change in the command line parsing has had some subtle change in which options we handle in which order.

Moving the '-d' to the front of the command line works.

We should either fix the parsing, or fix the README.md and help to match the code.

devimc:
remove -d option from create cmd line

fixes #232

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,7 +131,7 @@ or ``containerd``::
 
 Or, to simulate how ``containerd`` calls the runtime::
 
-    $ sudo ./cc-oci-runtime --log "$logfile" --log-format json create --bundle "$bundle_dir" --console $(tty) -d --pid-file "$pidfile" "$name"
+    $ sudo ./cc-oci-runtime --log "$logfile" --log-format json create --bundle "$bundle_dir" --console $(tty) --pid-file "$pidfile" "$name"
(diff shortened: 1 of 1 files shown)


dlespiau:
lgtm

grahamwhaley:
well, technically '-d' does work, it is in the parsing code, just not in that position on the command line. If you put it as the first argument then it is fine. So, the first question to @jodh-intel may be - would you like us to leave the '-d' in the hand run example here?

jodh-intel:
I'd say remove it.

A general point, we should probably only ever specify the long options (if there is a corresponding one for a short option) in all the docs.

lgtm.

PR #240, opened 2016-09-09
environment: Provide HOME in the default container environment

### LINKED ISSUE #139: HOME isn't set in our default enviroment
We should really provide what docker/runc have by default. Inevitably people will rely on this (say even in their Dockerfile). $HOME may well be treated in a special way because while it's not part of the 'env' section docker generates here:

```
    "env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "HOSTNAME=0d280ea85f3f"
    ],
```

but running env in the container shows it set:

```
$ sudo docker run debian env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=891d2fbed7e8
HOME=/root
```

and we don't:

```
$ sudo docker run --runtime cc debian env
container 3f7d1c16b086a22068493f6749fa975310612292b1497d2a731990b787760072 created
[    0.122995] systemd[1]: Failed to initialise default hostname
HOSTNAME=3f7d1c16b086
OLDPWD=/
TERM=vt220
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
LANG=en_US.UTF-8
PWD=/
```

amshinde:
HOME is provided by default by runc. Users will expect the same behaviour
from our runtime.
We first check if HOME is provided in the config provided by docker.
If not we set it to the home directory of the workload user.

Fixes #139

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -30,6 +30,9 @@
 #include <stdbool.h>
 #include <errno.h>
 #include <sys/types.h>
+#include <stdio.h>
+#include <sys/types.h>
+#include <pwd.h>
 
 #include <glib.h>
 #include <glib/gstdio.h>
@@ -523,6 +526,98 @@ fail1:
 }
 
 /*!
+ * Get the home directory for the workload user
+ *
+ * \param config \ref cc_oci_config.
+ * \param passwd_path Path to the local passwd file
+ *
+ * \return Newly-allocated path string on success, else \c NULL.
+ */
+
+private gchar*
+get_user_home_dir(struct cc_oci_config *config, gchar *passwd_path) {
+	gchar          *user_home = NULL;
+	FILE           *pw_file = NULL;
+	struct passwd  *pw_entry;
+
+	if (! (config && passwd_path)) {
+		return NULL;
+	}
+
+	pw_file = g_fopen (passwd_path, "r");
+	if ( pw_file == NULL) {
+		g_warning("Could not open password file: %s\n", passwd_path);
+		return NULL;
+	}
+
+	while ((pw_entry = fgetpwent(pw_file)) != NULL) {
+		if (pw_entry->pw_uid == config->oci.process.user.uid) {
+			user_home = g_strdup(pw_entry->pw_dir);
+			break;
+		}
+	}
+
+	fclose(pw_file);
+	return user_home;
+}
+
+/*!
+ * Set the HOME environment variable
+ *
+ * \param config \ref cc_oci_config.
+ *
+ * returns early if HOME is present in the environment configuration in \p config
+ */
+private void
+set_env_home(struct cc_oci_config *config)
+{
+	g_autofree gchar *user_home_dir = NULL;
+	g_autofree gchar *passwd_path = NULL;
+
+	if (! (config && config->oci.process.env)) {
+		return;
+	}
+
+	/* Check if HOME is set in the environment config */
+	for (gchar **var = config->oci.process.env; *var != NULL; var++) {
+		if (g_str_has_prefix (*var, "HOME=")) {
+			g_debug("Home is already set in the configuration\n");
+			return;
+		}
+	}
+
+	guint env_len = 1 + g_strv_length(config->oci.process.env);
+	gchar **new_env = g_new0(gchar*, env_len + 1);
+
+	passwd_path = g_strdup_printf ("%s/%s", config->oci.root.path, PASSWD_PATH);
+	user_home_dir = get_user_home_dir(config, passwd_path);
+
+	if (! user_home_dir) {
+		// Fallback to stateless path 
+		g_free(passwd_path);
+		passwd_path = g_strdup_printf ("%s/%s", config->oci.root.path,
+						STATELESS_PASSWD_PATH);
+		user_home_dir = get_user_home_dir(config, passwd_path);
+
+		// If we are not able to retrieve the home dir, set the default as "/"
+		if (! user_home_dir) {
+			user_home_dir = g_strdup("/");
+			g_debug("No HOME found in environment, so setting HOME %s for user %d",
+				user_home_dir, config->oci.process.user.uid);
+		}
+	}
+	new_env[0] = g_strdup_printf("HOME=%s", user_home_dir);
+
+	for (int i = 0; i < env_len-1; i++) {
+		new_env[i+1] = g_strdup(config->oci.process.env[i]);
+	}
+
+	g_strfreev(config->oci.process.env);
+	config->oci.process.env = new_env;
+}
+
+
+/*!
  * Create the containers Clear Linux workload file
  * (\ref CC_OCI_WORKLOAD_FILE).
  *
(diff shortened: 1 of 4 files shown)


src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
 }
 
 /*!
+ * Get the home directory for the workload user
+ *
+ * \param config \ref cc_oci_config.
+ *
+ * \return home dir for user on sucess, "/" by default
```
jodh-intel: This needs to state explicitly that the function returns `NULL` on failure and that the function returns a newly allocated string on success.

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+	gchar          *password_path = NULL;
+	FILE           *pw_file = NULL;
+	struct passwd  *pw_entry;
+
+	if (! config) {
+		goto end;
+	}
+
+	password_path = g_strdup_printf ("%s/%s", config->oci.root.path, "etc/passwd");
+	pw_file = g_fopen (password_path, "r");
```
jodh-intel: This will fail on a `clearlinux` docker image since there is no `/etc/passwd` (stateless). So, `/` will be returned for the root user even though its home directory is (for `clearlinux`) `/root`. This seems reasonable but needs to be documented in the function header I think.

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+	}
+	return user_home;
+}
+
+/*!
+ * Set the HOME environment variable
+ *
+ * \param config \ref cc_oci_config.
+ *
+ * returns early if HOME is present in the env config
```
jodh-intel: "in the env config" is a little vague - can you change it to something like:

```
returns early if HOME is present in the environment configuration in \p config.
```

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+{
+	gchar *user_home_dir = NULL;
+
+	if (! (config && config->oci.process.env)) {
+		return;
+	}
+
+	/* Check if HOME is set in the environment config */
+	gchar **var = config->oci.process.env;
+	while (*var != NULL) {
```
jodh-intel: This might be better as a `for (gchar **var = config->oci.process.env; ...)` loop - `var` would then also be further localised.

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+	}
+
+	guint env_len = 1 + g_strv_length(config->oci.process.env);
+	gchar **new_env = g_new0(gchar*, env_len + 1);
+
+	user_home_dir = get_user_home_dir(config);
+	new_env[0] = g_strdup_printf("HOME=%s", user_home_dir);
+
+	for (int i = 0; i < env_len-1; i++) {
+		new_env[i+1] = strdup(config->oci.process.env[i]);
```
jodh-intel: `g_strdup()`.

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+ * Set the HOME environment variable
+ *
+ * \param config \ref cc_oci_config.
+ *
+ * returns early if HOME is present in the env config
+ */
+private void
+set_env_home(struct cc_oci_config *config)
+{
+	gchar *user_home_dir = NULL;
```
jodh-intel: You could use `g_autofree` to avoid the need for the explicit `g_free()`.

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
+			return;
+		}
+		var++;
+	}
+
+	guint env_len = 1 + g_strv_length(config->oci.process.env);
+	gchar **new_env = g_new0(gchar*, env_len + 1);
+
+	user_home_dir = get_user_home_dir(config);
+	new_env[0] = g_strdup_printf("HOME=%s", user_home_dir);
```
jodh-intel: It would be useful to add a call like the following so we know when this code path is followed:

```
g_debug("No HOME found in environment, so setting %s for user %s", user_home_dir, config->oci.process.user.uid);
```

src/oci.c
```diff
@@ -523,6 +526,90 @@ cc_oci_attach (struct cc_oci_config *config,
...
 
 /*!
+ * Get the home directory for the workload user
+ *
+ * \param config \ref cc_oci_config.
+ *
+ * \return home dir for user on sucess, "/" by default
+ */
+
+private gchar*
```
jodh-intel: `private` is used for functions that would normally be `static` but which we need to run tests on... could you add some tests for these new functions please? :)

jodh-intel:
Hi @amshinde - a few minor issues to deal with.

The main problem though is that there are no tests for this new code. Please could you add new tests to `tests/oci_test.c`.

chavafg:
qa-passed

amshinde:
jodh-intel yes I have some tests written. Will send them out. Just thought would send out the initial code for review first.

jodh-intel:
@amshinde - great, thanks! :)

src/oci.c
amshinde: Thats a good point. I could  fallback to /usr/share/etc/passwd for stateless systems if we are not able to to retrieve the the home dir from /etc/passwd.

src/oci.c
amshinde: This function returns "/" by default. But I think I am going to change this in my next commit.

chavafg:
qa-failed

grahamwhaley:
Hi @chavafg (and @jodh-intel and @dlespiau ) I see we have a qa-failed here. We'll have to see if that is a real fail or a false-negative - but, either way, it makes me think that we should consider if the qa-failed messages can have some way of publicizing the failure reasons or logs etc., otherwise the submitter may be left somewhat in the dark as to why the QA failed.

jodh-intel:
@chavafg: +1 for @grahamwhaley's idea.

chavafg:
qa-passed

amshinde:
qa-failed was a false positive. jodh-intel any review comments?

src/oci.c
```diff
@@ -523,6 +526,98 @@ cc_oci_attach (struct cc_oci_config *config,
...
+
+	pw_file = g_fopen (passwd_path, "r");
+	if ( pw_file == NULL) {
+		g_warning("Could not open password file: %s\n", passwd_path);
+		return NULL;
+	}
+
+	while ((pw_entry = fgetpwent(pw_file)) != NULL) {
+		if (pw_entry->pw_uid == config->oci.process.user.uid) {
+			user_home = strdup(pw_entry->pw_dir);
```
jodh-intel: `g_strdup()`.

jodh-intel:
Hi @amshinde - just one minor nit (happy to approve after that).

dlespiau:
lgtm

grahamwhaley:
lgtm
will merge

grahamwhaley:
but cannot, as we are waiting for the qa tag. And Mexico are on holiday until next week. @jodh-intel , thoughts on how we handle this - we can just wait this week whilst we work out the new qa auto-workflow, or we can disable the qa ack for the minute.

grahamwhaley:
OK, we've worked around the merge block for now - so merging.

PR #241, opened 2016-09-09
Process: Add QA check to pullapprove config.

jodh-intel:
The QA team are now running additional tests which are triggered when
new PRs are raised. The test outcome is marked by adding a special
comment to the PR:

  - a "qa-passed" comment means all the tests passed.
  - a "qa-failed" comment means atleast one test failed.

Updated the pullapprove configuration file to:

- Require an additional approval from named QA members.

- Allow a PR to be explicitly rejected by introducing use of the
  `reject_regex` option containing `qa-failed`
  (if any QA tests fail, the PR cannot be approved).

- Modified the `approve_regex` to add `qa-passed` and changed `Approve`
  to `Approved` to mirror the `reject_regex`'s `rejected` value.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,15 +1,26 @@
 approve_by_comment: true
-approve_regex: '^(LGTM|lgtm|Approve|\+1|:\+1:)'
+approve_regex: '^(LGTM|lgtm|Approved|\+1|:\+1:|qa-passed)'
+reject_regex: '^(Rejected|-1|:-1:|qa-failed)'
 reset_on_push: true
 reset_on_reopened: true
 author_approval: ignored
 reviewers:
-  teams:
-    - clear-containers-intel
-  name: default
-  required: 2
-  conditions:
-    branches:
-      - master 
+  -
+    name: default
+    teams:
+      - clear-containers-intel
+    required: 2
+    conditions:
+      branches:
+        - master
+  -
+    name: qa
+    members:
+      - chavafg
+      - gabyct
+    required: 1
+    conditions:
+      branches:
+        - master
 signed_off_by:
   required: true


jodh-intel:
Since the QA approval mechanism described is entirely automated, there is a chance that problems might occur.

#  Contingencies

## Automated system not available

In the case that the automated system is unavailable, the QA members specified in the pullapprove config file can run the tests manually and add the appropriate QA comment to the PR to satisfy pullapprove.

## Worst-case scenario

If the automated QA system is not available and neither are the QA members, with the agreement of the team, the pullapprove github webhook could be temporarily disabled to allow an urgent PR to be merged.

# Notes

The configuration changes on this PR are somewhat of a compromise since pullapprove does not currently provide as much control as we'd like:

- Ideally, pullapprove would allow custom `approve_regex` and `reject_regex` for each "reviewer" allowing only QA members to trigger a true QA sign-off.

  Currently, the QA members can use *any* of the specified `approve_regex` values to signify QA sign-off (like `lgtm`). Conversely, any member of the default reviewer group could indicate their sign-off with one of the QA comments.

  There is an open pullapprove issue on this - see: https://github.com/pullapprove/support/issues/44

- To avoid having to disable the pullapprove webhook in the worst-case scenario, it would be better to allow a set of named users to also mark the PR with QA sign-off (after they had run the required tests
  themselves). However, pullapprove.com does not specify what happens in the scenario that a particular reviewer is both specified as a user *and* as part of a group in the "reviewers" sequence.

jodh-intel:
Once approved, we need to update https://github.com/01org/cc-oci-runtime/wiki/Release-Process documenting this enhanced process.

dlespiau:
An alternative could be to directly use the status API instead of going through pullapprove: https://developer.github.com/v3/repos/statuses/

grahamwhaley:
Agreed. We are moving very fast here, and a slip will cause a stall. @jodh-intel is it possible to make it a warning (is there a way?) rather than a hard stop?

grahamwhaley:
Yep, I did consider we could hook directly ourselves rather than use pullapprove. It will give us more flexibility, but I presume may take more effort. If we went this route then I'd really like to check how supported and back up our machine handling this is (and that should really go for our current solution as well ;-) )

jodh-intel:
@dlespiau / @grahamwhaley - well, the way to make this advisory is simply not to merge this PR and allow QA to continue to add the comments :)

I agree that we need a high level of confidence in the QA system, but we also need to ask ourselves which scenario we prefer:

1) Merging code that potentially contains major issues which will only be uncovered once they hit `master` (bad imho).
2) Having code that goes into `master` "test-clean" and occasionally having to disable the webhook (trivial) or ask QA to manually add a comment in case of infrastructure issues (trivial).

I'd vote for (2) every time :)

Maybe we can keep this PR open until we can be convinced of the reliability aspect?

grahamwhaley:
I think we are only asking for it to be advisory whilst we check and gain confidence in the stability of the QA system - once we know it is working well, then we make it mandatory.
And in the interim, if you see an advisory failure, you go and make sure you understand why it has failed (is it a code fail, or a QA system fail) before you merge. Having said that....
 I think the issue here is going to be that there will be nothing saying 'wait for the QA before merging', and if the QA is down for instance, and we have not got QA as a blocker to merging, then it is highly likely we will merge things because we forget we have to wait for the QA cycle.
In which case, yes, maybe we should enable this QA system as a merge blocker and see how it goes. I think we have to reserve the right to back it off if we find it is slowing process whilst we settle the system down.

grahamwhaley:
And I should have said - therefore I think I need to give this a:
lgtm
and we see how it goes.

dlespiau:
let's seen how it goes indeed.
lgtm.

PR #243, opened 2016-09-09
cmdline: Clarify where you can place global and command options

### LINKED ISSUE #242: Clarify command line option ordering
Related to #232 , the cor command line help implies you can place any options before or after the [command], but in reality you cannot. We should try to make this clearer in the command line help.
runc is slightly clearer in that it shows:

```USAGE:
   runc [global options] command [command options] [arguments...]
...
GLOBAL OPTIONS:
```

and cor currently shows:

```
Options:
...
Usage:
cc-oci-runtime [options] [command] [options]
```

I have a small PR to follow.

grahamwhaley:
The Usage string inferred that you could place any options any side
of the [command], but what it meant to indiate was that global options
come before the command and command specific options come after the
command. Change the message to make that clearer.

Ideally we would also modifiy the 'Options:' paragraph section in the
help to say 'Global Options:', but that is buried in the source of the
glib g_option source, with no way to override.

Fixes #242

### DIFF
--- a/src/commands/help.c
+++ b/src/commands/help.c
@@ -25,7 +25,7 @@ usage (void)
 {
 	struct subcommand **sub;
 	g_print ("Usage:\n");
-	g_print ("%s [options] [command] [options]\n",
+	g_print ("%s [global options] [command] [command options]\n",
 			PACKAGE_NAME);
 	g_print ("\n");
 	g_print ("Supported commands:\n");


dlespiau:
lgtm

jodh-intel:
Typo in commit msg ("indiate"), but we'll let that go :)

lgtm.

Issue #123, opened 2016-08-15
running functional tests make the host machine lose network configuration


chavafg:
Host machine hangs when executing test 'start then kill (implicit signal)'. 

Steps to reproduce: 
run autogen:
 ./autogen.sh --with-qemu-path=/usr/bin/qemu-lite-system-x86_64 --with-tests-bundle-path=/root/mycontainer/ --with-cc-image=/usr/share/clear-containers/clear-containers.img --with-cc-kernel=/usr/share/clear-containers/vmlinux.container --enable-debug

run make:
 make -j 

run make check:
 make check

============================================================================
Testsuite summary for cc-oci-runtime 0.1
============================================================================
 TOTAL: 22
 PASS:  22
 SKIP:  0
 XFAIL: 0
 FAIL:  0
 XPASS: 0
 ERROR: 0
============================================================================
make[2]: Leaving directory '/root/cc-oci-runtime'
make[1]: Leaving directory '/root/cc-oci-runtime'
 ✓ cor -h
 ✓ cor exec -h
 ✓ cor kill -h
 ✓ cor list -h
 ✓ cor pause -h
 ✓ cor resume -h
 ✓ cor start -h
 ✓ cor state -h
 ✓ cor delete -h
 ✓ cor stop -h
 ✓ cor foo -h
 ✓ kill without container id
 ✗ kill with invalid container id
   (in test file tests/functional/kill.bats, line 27)
     `[[ "${output}" =~ "failed to parse json file:" ]]' failed
   Test logs: /tmp/tmp.6dCLcI3EKY-cor-test
   start then kill (implicit signal)

chavafg:
After trying to reproduce locally,  @jcvenegas found that the issue is that the host machine is loosing the network when running the functional tests.

jodh-intel:
This is also related to #93 - if you run `sudo unshare -n bats tests/functional/kill.bats` most of the kill tests work. Without the `unshare -n`, re-running the tests causes a storm of network errors if you run with `--debug --global-log ...`.

However, we should not require the functional tests to run as root, hence we need a way to disable network setup for this.

@mcastelino - could we just do something like:

```
if (getuid ()) {
   g_critical ("not running as root, so network setup disabled");
} else {
  cc_oci_network_create(config);
}
```

.. and run the tests in a new net ns?

grahamwhaley:
Just pinging @mcastelino for feedback on @jodh-intel suggestion.

jodh-intel:
[...]

PR #254, opened 2016-09-12
State: Make runtime re-read statefile last updated by child.

jodh-intel:
The child process that becomes the hypervisor updates the on-disk state
file, so the parent process (the runtime) must re-read the state file to
reflect the true state.

This fixes the "run" command and the failing functional tests in "start.bats".

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -1205,6 +1205,16 @@ cc_oci_run (struct cc_oci_config *config)
 		return false;
 	}
 
+	/* Update the config object based on the state.
+	 *
+	 * This is required since the child process that becomes the
+	 * hypervisor has now updated the on-disk state file. But the
+	 * parents state object does not reflect the on-disk state.
+	 */
+	if (! cc_oci_config_update (config, state)) {
+		return false;
+	}
+
 	if (! cc_oci_start (config, state)) {
 		return false;
 	}


chavafg:
qa-passed

dlespiau:
Something bugs me: if we had a test not working, how do we have qa-passed comments so far?

grahamwhaley:
@dlespiau does #255 explain at all? If not, @jodh-intel , can you add some info here (or reference an Issue) to clarify Damiens question?
Other than that, the notion looks fine, so:
lgtm

dlespiau:
#255 doesn't explain it: the QA system (from which qa-approved is sent q) != travis, the QA system is supposed to test on real hw with KVM, so I was expecting the functional tests to be run and thus fail up until this PR. But that's not the case.

jodh-intel:
I believe the confusion may come from the fact that the QA tests marked as "functional tests" are actually running `tests/integration/docker/`, but they are not running `tests/functional/` (which is what this PR fixes).

@chavafg - does that sound right?

chavafg:
@jodh-intel, right, we run the tests that are under `test/integration/docker/`. Sorry for the confusion. They are functional tests but we decided to put into integration tests since we are testing the integration with docker. Actually I have disabled the functional-tests of cc-oci-runtime, since they have network issues (https://github.com/01org/cc-oci-runtime/issues/123).

dlespiau:
Thanks, that does explain it and we have a issue that missing coverage opened!

Issue #255, opened 2016-09-13
Run all bundled tests on every PR raised.


jodh-intel:
Currently, Travis only runs the unit tests.

We need to find a way to run *all* the tests under Travis. However, we face a few problems:

- The newest Linux environment available under travis is Ubuntu Trusty (14.04).
- The Travis environment apparently does not provide VT-x support
  (since Google Compute Engine doesn't offer that - see https://github.com/travis-ci/travis-ci/issues/1419).

dlespiau:
Adding to the 2.1 milestone, seems important enough to me we have that working during that time frame. Do yell if unhappy.

jodh-intel:
thanks - I actually meant to do that at creation time but added the 2.1 backlog label erroneously ;)

jodh-intel:
I've got 2 branches that attempt to solve this problem:

- https://github.com/jodh-intel/cc-oci-runtime/tree/run-functional-tests-in-travis

  This one tries to use only what Travis provides, but requires building gcc6 to allow the build of qemu-lite. The problem is that the gcc build take a looong time and generates a lot of output meaning Travis will fail the job.

- https://github.com/jodh-intel/cc-oci-runtime/tree/run-tests-in-travis-under-docker

  This branch uses a custom Docker image based on Fedora 24 that Travis can pull and run the runtime build and tests within. Whilst cleaner than the first branch, the generated build image is large (>1GB) and that may take some time to pull from within the Travis environment.

The problem with both branches though is that neither solve the issue (nor can they) that Travis doesn't offer VT-x support. Since Clear Containers are heavily reliant on kvm facilities (virtio, including 9p support), we cannot run the tests under Travis even using qemu emulation.

The only solution at this point seems to be to use an alternative to Travis (ideally bare metal) for running the functional tests.

jodh-intel:
Adding Salvador to provide QA input.

jodh-intel:
@chavafg - can this now be closed?

chavafg:
Yes, now that we have CI testing (including functional tests) for all our PRs. Closing this issue.

Issue #114, opened 2016-08-11
Find a way to use the travis cache for our dependencies


dlespiau:
We should try to cache the building of glib/json-glib, they take the vast majority of our test time, making the cycle "change -> test" longer than necessary.

Simply enabling the cache (say caching /usr/local/) doesnss't seem to be enough at a glance, we have to detect if we do need to rebuild the cache (there isn't any cache or the versions of the deps we need are different than the one in the cache) or, if not, skip the whole dependency setup.

For that detection I'd simply copy versions.txt to the cache and compare it with the one in tree. If there's no versions.txt in the cache or if they are different, rebuild the cache. Skip rebuilding otherwise.

https://docs.travis-ci.com/user/caching/#Arbitrary-directories

grahamwhaley:
Possibly relevant - I discussed the notion of ccache enabling on #65 . I still suspect that could save us a bunch of compile time?

dlespiau:
Right, that'd be another possibility to look at, but caching the whole result of the "make install" should be even faster!

dlespiau:
Let's garbage collect that one, the travis build is fast enough as is, no need to be overzealous.

PR #256, opened 2016-09-13
Run functional tests in travis

jodh-intel:


grahamwhaley:
lgtm
Adding a whole gcc build - ouch!  Makes me think of #114 again...

jodh-intel:
Indeed. This isn't ideal.

In fact, I'm hitting Travis limits on log size (4 MB), but more frustrating is the time it takes to build gcc 6 in Travis - this, coupled to running qemu with emulation is going to seriously impact the PR turnaround time unless we can find a solution to #114.

jodh-intel:
Fixes #255.

jodh-intel:
Travis is now falling over as it's taking too long to build (gcc), so now looking at the caching options... :)

jodh-intel:
This approach is proving to be non-viable. Closing to allow investigations of a better approach.

PR #259, opened 2016-09-14
Tests: Verify that attaching a volume to a container works correctly

chavafg:
This tests create a docker volume and verify that it is attached to
a clear container correctly.

### DIFF
--- a/tests/integration/docker/volume.bats
+++ b/tests/integration/docker/volume.bats
@@ -25,14 +25,29 @@ setup() {
 	source test.common
 	cleanDockerPs
 	runtimeDocker
+	volName='volume1'
 }
 
-@test "Volume inspect container" {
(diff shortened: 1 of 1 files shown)


jcvenegas:
I wonder if we can also add other scenarios, use '-v' to share a folder and a file

dlespiau:
Yes please! can we have an additional test trying to bind-mount a single file? We recently had a bug in that area and we fixed it without unit test (oops!).

grahamwhaley:
I'm going to:
lgtm
as imo we don't need to block adding this first test whilst we write more or improve it.

dlespiau:
sounds good, I've created a new issue then: https://github.com/01org/cc-oci-runtime/issues/262

lgtm

PR #260, opened 2016-09-14
Tests: Fix failure when running under docker.

jodh-intel:
The test_cc_oci_ns_setup() test asserts the possible errno values when
cc_oci_ns_setup() fails. However, EPERM was missing, which gets returned
when the tests are run as a non-priv user under docker.


grahamwhaley:
lgtm

grahamwhaley:
Going to close and re-open this PR to see if we can get the qa/pullapprove hooks change to take effect...

PR #261, opened 2016-09-15
data: guest-config: Add hyperstart.service as boot dependency

jcvenegas:
Clear containers image will use hyperstart as daemon started by
hypertart.service. This patch adds this service in the boot flow.

Notice that it uses Wants=hyperstart.service in container.target
(as others services and mount units), "if the units
fail to start or cannot be added to the transaction, this has no
impact on the validity of the transaction as a whole"[1]. This means
that if a container image does not provide hyperstart.service or
fails the boot wont fail.

1. https://www.freedesktop.org/software/systemd/man/systemd.unit.html

### DIFF
--- a/data/container.target
+++ b/data/container.target
@@ -10,3 +10,4 @@ Wants=systemd-networkd.service
 Wants=systemd-resolved.service
 Wants=opt-rootfs.mount
 Wants=opt-rootfs-proc.mount
+Wants=hyperstart.service


jcvenegas:
hyperstart  is already packaged for Clear Linux and provides hyperstart.service file but is not a upstream change yet.

data/container-workload.service
```diff
@@ -3,6 +3,7 @@ Description=Container Workload
 After=opt-rootfs-resolv.service
 After=opt-rootfs-proc.mount
 After=opt-rootfs-sys.mount
+After=hyperstart.service
```
dlespiau: This doesn't seem necessary, the workload today doesn't need hyperstart (and in the future this file will probably go away)

jcvenegas: good catch!,  thinking about this we will want that hyperstart start after shared rootfs mount? or the daemon will mount the shared rootfs (and will remove opt-rootfs.mount )?

dlespiau: this needs to be looked at in more details, the STARTPOD/NEWCONTAINER hyperstart commands seem to do some setup in this area, something we'll need to do at "create" time in the runtime. Investigation still needed.

dlespiau:
lgtm

devimc:
:+1:

PR #267, opened 2016-09-16
Docs: Fix outstanding doxygen issues.

jodh-intel:
We really need to run doxygen regularly.

It's also rather surprising that there is no way to make doxygen return
1 if it find files that contain doxygen comments, but which are invalid
/ incomplete as if so we could tie that into the build.

### DIFF
--- a/src/namespace.c
+++ b/src/namespace.c
@@ -124,6 +124,8 @@ out:
  *
  * \param config \ref cc_oci_config.
  *
+ * \return \c true on success, else \c false.
+ *
  * \todo Show the namespace path. For unshare, the strategy should be to
  * call cc_oci_resolve_path (), passing it the value of.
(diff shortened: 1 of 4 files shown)


dlespiau:
lgtm

[Approved with PullApprove](https://pullapprove.com/01org/cc-oci-runtime/pull-request/267/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=01org/cc-oci-runtime)

grahamwhaley:
lgtm
Do we have a doxygen/docs Makefile target? Could we wrap doxygen into a 'doxygen | fgrep -i error | wc' and use that as the exit code or similar maybe?

jodh-intel:
Yeah, it's not ideal, but I think that's the best we can do. I've raised #268 for a new build target.

PR #276, opened 2016-09-21
Tests: Add tests for bind-mounting files and directories

chavafg:
Couple of tests that bind-mount a single file and a
directory to a container.

### DIFF
--- a/tests/integration/docker/volume.bats
+++ b/tests/integration/docker/volume.bats
@@ -23,6 +23,7 @@
 
 setup() {
 	load common
+	load volume
 	cleanDockerPs
 	runtimeDocker
 	volName='volume1'
@@ -42,10 +43,33 @@ setup() {
 	testFile='hello_world'
 	containerPath='/attached_vol'
 	touch "$volPath/$testFile"
-	$DOCKER_EXE run -i -v "$volName":"$containerPath" busybox ls "$containerPath" | grep "$testFile"
-	RC=$?
+	run bash -c "$DOCKER_EXE run -i -v $volName:$containerPath busybox ls $containerPath | grep $testFile"
 	$DOCKER_EXE rm $($DOCKER_EXE ps -qa)
-	[ $RC -eq 0 ]
+	[ $status -eq 0 ]
+}
+
+@test "Volume - bind-mount a directory" {
+	dir_path="$(pwd)/sharedSpace"
+	test_file="foo"
+	containerPath="/root/sharedSpace"
+	mkdir "$dir_path"
+	touch "$dir_path/$test_file"
+	run bash -c "$DOCKER_EXE run -i -v $dir_path:$containerPath busybox ls $containerPath | grep $test_file"
+	rm -r "$dir_path"
+	$DOCKER_EXE rm $($DOCKER_EXE ps -qa)
+	[ $status -eq 0 ]
+
+}
+
+@test "Volume - bind-mount a single file" {
+	test_file="$(pwd)/foo"
+	file_content="bar"
+	echo "$file_content" >  "$test_file"
+	containerPath="/root/foo"
+	run bash -c "$DOCKER_EXE run -i -v $test_file:$containerPath busybox cat $containerPath | grep $file_content"
+	rm "$test_file"
+	$DOCKER_EXE rm $($DOCKER_EXE ps -qa)
+	[ $status -eq 0 ]
 }
 
 @test "Delete the volume" {


tests/integration/docker/volume.bats
```diff
@@ -48,6 +48,31 @@ setup() {
 	[ $RC -eq 0 ]
 }
 
+@test "Volume - bind-mount a directory" {
+	dir_path="$(pwd)/sharedSpace"
+	test_file="foo"
+	containerPath="/root/sharedSpace"
+	mkdir "$dir_path"
```
jodh-intel: If this volume test fails at the `docker run` stage for some reason, and the test is then re-run, it will fail on this `mkdir` as the directory will exist from the previous test run. The best solution would probably be to add a `setup()` and `teardown()` and have `dir_path` handled there to ensure a completely clean test environment for each test run.

tests/integration/docker/volume.bats
```diff
@@ -48,6 +48,31 @@ setup() {
...
+	[ $RC -eq 0 ]
+}
+
+@test "Volume - bind-mount a single file" {
+	test_file="$(pwd)/foo"
+	file_content="bar"
+	echo "$file_content" >  "$test_file"
+	containerPath="/root/foo"
+	$DOCKER_EXE run -i -v "$test_file":"$containerPath" busybox cat "$containerPath" | grep "$file_content"
+	RC=$?
```
jodh-intel: I don't see how this can work as if the `docker run` fails, the test will immediately abort. I think you need to do something like the following to allow `docker run` to potentially fail but capture the return code and continue:

```
{ $DOCKER_EXEC run ... | grep ...; RC=$?; } || true
rm ...
$DOCKER_EXE rm ...
[ $RC -eq 0 ]
```

jodh-intel: ... or more simply, use the bats `run` command:

```
run $DOCKER_EXEC run ... | grep ...
RC=$status
rm ...
$DOCKER_EXE rm ...
[ $RC -eq 0 ]
```

tests/integration/docker/volume.bats
chavafg: what about this option:
```
mkdir "$dir_path"
touch "$dir_path/$test_file"
run $DOCKER_EXE run -i -v "$dir_path":"$containerPath" busybox ls "$containerPath" | grep "$test_file"
rm -r "$dir_path"
$DOCKER_EXE rm $($DOCKER_EXE ps -qa)
[ $status -eq 0 ]

```

I wouldn't like to manage the directory  under `setup()` and `teardown()` since it will be created and deleted for each test case we have in the `volume.bats` file

jodh-intel: Yes, that works too :) I'd suggest adding a comment in the test explaining why you are delaying the test on `$status`.

chavafg: seems like bats `run` command has issues when using it with pipes `|`, need to find another way.

jodh-intel: I'll add the do-not-merge label for now then...

jodh-intel: @chavafg - if you haven't already found it, this syntax will work:

```
run bash -c "cmd1 | cmd"
```

chavafg: thanks @jodh-intel

chavafg:
Updated commit to fix @jodh-intel  comments.

jodh-intel:
@chavafg - I assume this is ready for re-review? Please can you remove the `do-not-merge` label if so.

Assuming this is ready...

lgtm.

chavafg:
can someone else take a look at this? 

Thanks

jodh-intel:
lgtm.

chavafg:
qa-passed

devimc:
:+1: 

[Approved with PullApprove](https://pullapprove.com/01org/cc-oci-runtime/pull-request/276/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=01org/cc-oci-runtime)

dlespiau:
pull approve says it needs an ack from qa?

chavafg:
qa-passed

chavafg:
seems that we need an ack from @GabyCT, maybe because I was the author of the commit and I cannot give an approve on my PR?

GabyCT:
qa-passed

grahamwhaley:
@chavafg ooh, good spot - @jodh-intel , so there is a pullapprove blip for you - one that we can probably live with though.

jodh-intel:
@grahamwhaley - actually, it's not a blib, that's by design to avoid authors approving their own work:

```
author_approval:
    ignored: true
```

To resolve this properly, I think we need a virtual QA user who can add the QA comments and who will never be a PR author.

grahamwhaley:
@jodh-intel yep, a virtual QA user sounds like the right solution. thx

PR #281, opened 2016-09-22
Disable debug by default

jodh-intel:


grahamwhaley:
that's a bit of a mixed bag of vaguely related commits - not just 'disable debug' :-). But, the content:
lgtm

src/main.c
```diff
@@ -410,9 +410,6 @@ main (int argc, char **argv)
 {
 	gboolean ret;
 
-	// FIXME: --debug currently forcibly enabled
-	cc_log_options.enable_debug = true;
```
gorozco1: Can we change to:

`cc_log_options.enable_debug = false; `

amshinde:
lgtm

PR #282, opened 2016-09-22
Mount: fix mount sysfs

devimc:
With this patch sysfs will be mounted on /sys
this change is needed by hyperstart due to it
reads /sys/class/virtio-ports/ in order to find
CTL and TTY channels


jcvenegas:
lgtm , btw hyperstart is started outside the chroot  it should be able to use guest /sys/

gorozco1:
lgtm

dlespiau:
Ah, I knew I saw @amshinde fixed that one before :). Merged.

PR #285, opened 2016-09-22
create: Provide a way to append arguments to the hypervisor arg list

### LINKED ISSUE #219: Provide a more flexible way to build the list of qemu arguments
At the moment, arguments are quite static and trying to add a variable number of arguments is awkward. For networking support we want to be able to provide multiple interfaces based on the docker setup discovery.

That could be useful for more configuration qemu startup, say supporting a non-KVM qemu for travis.

Proposal: change the way we build the list of arguments: first approximation is to pass around an array of strings and have separate subsystems appending to it.

amshinde:
Dynamic arguments can be passed to the hypervisor by adding them to
an array which is appended to the static hypervisor argument list.
The static hypervisor args are constructed as before.

Fixes #219

### DIFF
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -465,18 +465,21 @@ out:
  *
  * \param config \ref cc_oci_config.
  * \param[out] args Command-line to expand.
+ * \param hypervisor_extra_args Additional args to be appended
  *
  * \return \c true on success, else \c false.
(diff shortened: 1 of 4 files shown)


jodh-intel:
lgtm

devimc:
:+1:

PR #288, opened 2016-09-25
build: Make the generated files depend on Makefile

dlespiau:
When we re-run configure, some Makefile variables may have changed as a
result and so we need to re-generate files potentially impacted.


jodh-intel:
lgtm.

devimc:
:+1:

PR #289, opened 2016-09-25
Make the travis output a bit more manageable

dlespiau:
Currently we output the full compilation logs of the dependencies but they don't add much value and make the travis UI sluggish.

I had to debug some issues in the travis build and the above made the job quite a pain, so decided to do something about it.

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -35,6 +35,19 @@ sudo apt-get -qq install valgrind lcov uuid-dev pkg-config \
   zlib1g-dev libffi-dev gettext libpcre3-dev cppcheck \
   libmnl-dev
 
+function compile {
+	name="$1"
+	tarball="$2"
+	directory="$3"
+
+	tar -xvf ${tarball}
+	pushd ${directory}
+	./configure --disable-silent-rules
+	make -j5
+	sudo make install
+	popd
+}
+
 mkdir cor-dependencies
 pushd cor-dependencies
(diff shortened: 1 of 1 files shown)


.travis-setup.sh
```diff
@@ -35,42 +35,46 @@ sudo apt-get -qq install valgrind lcov uuid-dev pkg-config \
   zlib1g-dev libffi-dev gettext libpcre3-dev cppcheck \
   libmnl-dev
 
+function compile {
```
jodh-intel: I did something very similar here:

- https://github.com/jodh-intel/cc-oci-runtime/blob/run-functional-tests-in-travis/.travis-setup.sh#L42

.travis-setup.sh
```diff
@@ -35,42 +35,46 @@ sudo apt-get -qq install valgrind lcov uuid-dev pkg-config \
...
+	name=$1
+	tarball=$2
+	directory=$3
```
jodh-intel: I'd personally always quote variables like this to avoid potential problems in the future wrt spaces. It's unlikely ever to be a problem under Travis with this particular function, but just good practise.

jodh-intel:
Potential quoting issues aside...

lgtm.

jodh-intel:
lgtm

chavafg:
qa-passed

amshinde:
lgtm

PR #294, opened 2016-09-26
Add state tests

### LINKED ISSUE #72: Create new bats test to assert state files
We should create a bats file that checks the state directory contents are as expected (`state.json`, `hypervisor.sock`, etc). We need to check file existence, type, owner, group, perms and where appropriate size (assert non-zero, etc).

jodh-intel:
Fixes #72.


devimc:
lgtm

dlespiau:
lgtm

chavafg:
qa-passed

Issue #287, opened 2016-09-25
Functional tests fail to run when using the default configuration


dlespiau:
When not giving any path to configure about where the test bundle is, it chooses `"$srcdir/tests/functional/bundle"` and that's the value used for the generation of common.bash. Unfortunately, this is a relative path and the tests fails, not being able to copy the bundle config.json:

```
 /bin/cp /home/damien/clear/sources/cc-oci-runtime/tests/functional/data/config-minimal-cc-oci.json ./tests/functional/bundle/config.json
   /bin/cp: cannot create regular file ‘./tests/functional/bundle/config.json’: No such file or directory
```

We can't use $abs_scdir in configure, but resolving the path in configure (for both the default value and the parameter given as path) sounds like the way to go to me.

jodh-intel:
Fixed by #297.

PR #297, opened 2016-09-27
Config: Expand path for automatic bundle-generation.

jodh-intel:
The automatic bundle generation feature (commit 7dd2c5a) contained a
bug where the default automatic bundle path was relative rather than
absolute.

Now, both the default and the path specified by
"--with-auto-bundle-creation-path=" is made absolute.

Note that symbolic links are not expanded, for symmetry with
other paths passed to configure (such as "--with-cc-image=" and
"--with-cc-kernel=" where symlink expansion is expressely undesirable).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -214,9 +214,12 @@ AC_ARG_WITH([auto-bundle-creation-path],
     ],
     [
         # default value
-        auto_bundle_creation_path="$srcdir/tests/functional/bundle"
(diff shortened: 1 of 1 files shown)


jodh-intel:
Fixes #287.

jcvenegas:
lgtm

dlespiau:
lgtm

chavafg:
qa-passed

Issue #327, opened 2016-10-13
functional tests fail intermittently


chavafg:
I am trying to build and test cc-oci-runtime using James' Dockerfile:  https://github.com/jodh-intel/cc-oci-runtime/blob/run-tests-in-travis-under-docker/data/Dockerfile

The `autogen` and `make` commands run successfully, but sometimes make functional-tests failed in this test: `start and kill state`

If you run the tests 10 times, you may be able to reproduce the issue. 

The error shown is: 

```
  ✗ start and kill state
   (from function `run_cmd' in file tests/functional/common.bash, line 293,                                                                                                                                                                  
    in test file tests/functional/state.bats, line 59)                                                                                                                                                                                       
     `run_cmd "$cmd" "0" "$COR_TIMEOUT"' failed                                                                                                                                                                                              
   Running: timeout 5 /root/cc-oci-runtime/tests/functional/../../cc-oci-runtime  --debug --root /tmp/tmp.WD4jIlT9jf-cor-test --global-log /run/cc-oci-runtime/cc-oci-runtime.log --hypervisor-log-dir /tmp/tmp.HNsEBAvhzr-cor-test create  -
-console --bundle /var/lib/oci/bundle tests_id                                                                                                                                                                                               
   2016-10-17T20:52:10.221597Z:3273:cc-oci-runtime:critical:hooks failed, child exiting                                                                                                                                                      
   exit code: 141                                                                                                                                                                                                                            
   expected exit code: 0                                                                                                                                                                                                                     
   See /root/cc-oci-runtime/test_start_and_kill_state.log                                                                                                                                                                                    
   Test logs: /tmp/tmp.WD4jIlT9jf-cor-test                                                                                                                                                                                                   
   Global log file: /run/cc-oci-runtime/cc-oci-runtime.log                                                                                                                                                                                   
   Hypervisor log dir: /tmp/tmp.HNsEBAvhzr-cor-test                                                                                                                                                                                          
   bats status: 0                                                                                                                                                                                                                            
   bats output (lines array contains 0 lines):                                                                                                                                                                                               
   BUNDLE_DIR: /var/lib/oci/bundle                                                                                                                                                                                                           
   ROOTFS_DIR: /var/lib/oci/bundle/rootfs 
```

This is part of the cc-oci-runtime-log:

```
[...]

Issue #155, opened 2016-08-23
Unit tests sometimes fail under Travis using clang


jodh-intel:
The Travis run for #149 failed even though the branch only added a single `.yml` file. Further the failure only occured with `clang` (the `gcc` build was succesful). Restarting the `clang` Travis run resulted in a successful run.

We've seen this issue before so this issue attempts to capture what we know about the problem.

It looks like since I restarted the Travis run, the original logs are not available on the Travis site, but I have local copies.

The test which failed was `sh_vm_test`. Viewing the raw log showed that the problem seemed to be caused by the following unexpected errors:

```
realpath 'QEMU-LITE' failed: No such file or directory
realpath 'CLEAR-CONTAINERS.img' failed: No such file or directory
realpath 'CONTAINER-KERNEL' failed: No such file or directory
```

The failing test is `test_vm_handle_section()`, specifically the test on the file `./tests/data/vm.json`. That test tries to read the three files named above. All 3 files files should exist since they were created by `tests/test_common.c:test_spec_handler()`. That function doesn't abort if the files cannot be created, but it does display and error. But the Travis logs do not contain the file creation errors.

jodh-intel:
Original Travis logs...

[log-clang.txt.gz](https://github.com/01org/cc-oci-runtime/files/431929/log-clang.txt.gz)
[log-gcc.txt.gz](https://github.com/01org/cc-oci-runtime/files/431931/log-gcc.txt.gz)

dlespiau:
sounds like it could be an effect of the -j5 and we need to find a way to express a dependency on the creation of this vm.json file for sh_vm_test?

dlespiau:
I haven't seen it happening since I changed the tests to run serially. Closing for now as it's travis only.

Please do re-open if encountered again.

PR #336, opened 2016-10-20
start: Correct how we determine if the runtime is being run standalone

amshinde:
The correct way of determining if we are being run by conatinerd during
start phase is to check if no tty is attached to stdin. We should'nt
be checking for config.terminal value as this refers to the stdin terminal
for the create phase.
Revert earlier changes introduced with this bug. As a side note, the terminal
value will be stored in the state file in the future to account for
the code that determines the config.terminal value programmatically.

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -1144,7 +1144,7 @@ cc_oci_start (struct cc_oci_config *config,
 	 *
 	 * Do not wait when console is empty.
 	 */
-	if ((config->oci.process.terminal && ! config->detached_mode) &&
+	if ((isatty (STDIN_FILENO) && ! config->detached_mode) &&
 	    !config->use_socket_console) {
 		wait = true;
 	}
@@ -1822,11 +1822,6 @@ cc_oci_config_update (struct cc_oci_config *config,
 	}
 
 	config->use_socket_console = state->use_socket_console;
-
-	if (config->console && ! config->use_socket_console && isatty (STDIN_FILENO)) {
-		g_debug ("enabling terminal for standalone mode");
-		config->oci.process.terminal = true;
-	}
 
 	if (state->vm) {
 		config->vm = state->vm;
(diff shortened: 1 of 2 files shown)


jcvenegas:
+1

grahamwhaley:
@amshinde The theory looks good to me - will let @jodh-intel peer at it. I note that one of the tests failed on the clang travis build - which feels somewhat 'spurious' to me, but we should check it out.

jodh-intel:
That failure is #155 again.

@amshinde - that piece of code has changed quite a bit over time as it has been problematic.

Can you confirm that you've tested this:

- standalone (with an arg to `--console` (and without) and with/without `--detach`).
- under docker (again with and without `--detach`, `--tty`, `--interactive`).

amshinde:
@jodh-intel I have tested this under docker with all the options mentioned  (--detach, --tty, --interactive) and standalone ,with and without --console.
For running standalone, I had to comment out the networking code as I lost connection to my system while running the test. I think it makes sense to disable to networking, while running standalone to not mess up networking using the same check above. I can send out another PR for that.

dlespiau:
lgtm

chavafg:
qa-failed

dlespiau:
@chavafg seems like it's just make distcheck failing (it's been a long time since the last time it did)?

chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
@dlespiau that is correct, there are some functional tests that fail randomly that I still need to check(issue #327).  Moving to qa-passed

dlespiau:
@jcvenegas could you give your +1 again, pullapprove seems to have forgotten you.

jcvenegas:
lgtm

PR #339, opened 2016-10-21
Config: Set sockets=1 for qemu.

jodh-intel:
When qemu is launched with socket > 1 ("-smp" option), performance is
harmed, so "sockets" and "threads" must be 1. The other values can
change (and should ideally match the cpu count on the host system).


jcvenegas:
lgtm

chavafg:
qa-failed

Issue #323, opened 2016-10-11
Test that "make dist" embed all needed files


dlespiau:
We currently run the tests in travis from a git checkout, but it'd be good to include testing that "make dist" generates a tarball that can pass the tests.

Two ways I see to do that:
- make distcheck (but we may not be able to honour DISTCHECK_CONFIGURE_FLAGS in travis)
- run make dist and run the travis tests from the resulting tarball, unpacked.

Issue #322, opened 2016-10-11
make distcheck not working properly


chavafg:
make distcheck cannot find `tests/functional/common.bash`

```
...
47/48 files checked 97% done
Checking ../../src/util.c...
Checking ../../src/util.c: DEBUG...
48/48 files checked 100% done
(information) Cppcheck cannot find all the include files (use --check-config for details)
  GEN      tests/functional/common.bash
/bin/sh: line 1: tests/functional/common.bash: No such file or directory
make[1]: *** [Makefile:4628: tests/functional/common.bash] Error 1
make[1]: Leaving directory '/root/test/cc-oci-runtime/cc-oci-runtime-0.1/_build/sub'
make: *** [Makefile:4379: distcheck] Error 1
```

dlespiau:
So, it's actually the change in how we generate certain files that break make distcheck. It boils down to us using sed to write the genrated files in $(srcdir) but in the general case, $(srcdir) has to be considered like a read-only directory, we always want to write in $(builddir).

This breaks the out of tree build that make distcheck does.

Issue #277, opened 2016-09-22
volume.bats fails running test "Volume - use volume in a container" as non-priv user


jodh-intel:
`make check` `tests/integration/docker/volume.bats` failure:

```
✓ Tag a container
 ✓ Volume - create
 ✓ Volume - inspect
 ✗ Volume - use volume in a container
   (in test file tests/integration/docker/volume.bats, line 44)
     `touch "$volPath/$testFile"' failed
   touch: cannot touch '/var/lib/docker/volumes/volume1/_data/hello_world': Permission denied
 ✓ Delete the volume

22 tests, 1 failure
make: *** [Makefile:4634: docker-tests] Error 1
```

Test needs to be skipped if running as a non-privileged user, or the volume should be created outside `/var/lib/docker/`.

Issue #79, opened 2016-08-05
Image and kernel paths not fully expanded when installing in custom prefix


dlespiau:
When configured with --prefix=home/damien/clear/install:

```
$ cat data/vm.json
{
    "vm": {
        "path": "/home/damien/clear/install/bin/qemu-system-x86_64",
            "image": "${prefix}/var/lib/cc-oci-runtime/data/image/clear-containers.img",
            "kernel": {
                "path": "${prefix}/var/lib/cc-oci-runtime/data/kernel/vmlinux.container",
                "parameters": "root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro  rw rootfstype=ext4 tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k panic=1 console=hvc0 console=hvc1 initcall_debug init=/usr/lib/systemd/systemd systemd.unit=container.target iommu=off quiet systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket systemd.show_status=false cryptomgr.notests"
            }
    }
}
```

Of course, cc-oci-runtime cannot, then, find those files.

PR #344, opened 2016-10-25
Assorted build fixes

dlespiau:
I somewhat doubt we've had make distcheck running recently, but this series is a leap towards it.

The (last?) thing remaining is being able to run all docker tests under a non-root user (already captured in an issue).

There's an awkward thing in having the docker tests part of make check (and so make distcheck): it means we need to run have docker running and configured to use the compiled runtime (problably through make install and having docker configured to take the runtime from the installation prefix)

This series also tweaks the travis build to do out-of-tree builds out of an exploded make dist tarball to add extra tests (for instance that we do distribute all the necessary files in the tarball). That's close to what make distcheck does without actually doing a make distcheck (that can't run in travis).


dlespiau:
Fixes https://github.com/01org/cc-oci-runtime/issues/323

dlespiau:
Fixes https://github.com/01org/cc-oci-runtime/issues/79

dlespiau:
PR updated:

  - Rebased against latest master (was conflicting)
  - Fixed https://github.com/01org/cc-oci-runtime/issues/277
  - make distcheck now passes on my laptop

dlespiau:
Fixes https://github.com/01org/cc-oci-runtime/issues/277
Fixes https://github.com/01org/cc-oci-runtime/issues/322

.travis/travis-setup-test-directory.sh
```diff
@@ -0,0 +1,28 @@
...
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# We run the travis script from an exploded `make dist` tarball to ensure
+# `make dist` has the necessary files to compile and run the tests.
+#
+# Note: we don't use `make distcheck` here as we can't run everything we want
+#       for discheck in travis.
+chronic make dist
```
jodh-intel: Might be worth a comment saying why `chronic(1)` is being used here.

Makefile.am
```diff
@@ -127,7 +130,8 @@ DISTCHECK_CONFIGURE_FLAGS = \
 	--enable-cppcheck \
 	--enable-valgrind \
 	--enable-code-coverage \
-	--enable-functional-tests
+	--enable-functional-tests \
+	--with-systemdsystemunitdir=/tmp/cc-distcheck/systemdunitdir/
```
jodh-intel: Can this directory be made unique?

dlespiau: Actually, I didn't really want to make it unique: it's not needed (ie. not needing several distinct copies of that data) and doing several make distcheck will not end up clobbering /tmp.

jodh-intel: Maybe a comment to that effect would be worthwhile to explain this.

dlespiau: That I can do!

dlespiau:
Updated PR:
  - Rebased against master (was conflicting with recent developments)
  - Put the build steps inside their own shell script to work around limitations in travis
  - Added some comments about the use chronic
  - Added quoting to $tarball_dir. Note that It's not actually a problem and adding quotes may actually fail to alert us that the dist tarball has a path with a space in, which isn't a nice property for a dist tarball. I'm not going to bikeshed the bikeshed though and just added the quotes.
  - Added a comment about the systemd unit dir in make distcheck

grahamwhaley:
Nice series @dlespiau  I think maybe we are waiting for just one or two tiny addends from @jodh-intel feedback, but other than that,
lgtm

dlespiau:
Fixed the "builddi" typo in commit message.

All the feedback from James should have been addressed, hopefully.

.travis.yml
```diff
@@ -22,13 +22,13 @@ env:
     - coverity_scan_run_condition='"$CC" = gcc'
 
 before_script:
-- ./.travis-setup.sh
+- ./.travis/travis-setup.sh
```
jodh-intel: The leading `./` is redundant, but innocuous :-)

jodh-intel:
lgtm

grahamwhaley:
I don't think the leading ./ will block the merge, so:
lgtm

chavafg:
qa-failed

dlespiau:
Looking at the CI dashboard, it shows that make distcheck passes (yey!) but make check doesn't (from time to time). It seems like it's a known flaky functional test:

```
not ok 42 start and kill state
```

jodh-intel:
surely it's :snowflake: ? :smile:

chavafg:
qa-passed

dlespiau:
@jodh-intel I wish!, ok merging that one. Thanks all for the reviews.

Issue #171, opened 2016-08-25
stdout/stderr not properly forwarded to the host


dlespiau:
Because we currently always use a tty for our binaries, we get stderr and stdout interleaved in stdout:

```
$ sudo docker run debian ls -l /etc/resolv.conf /etc/foo
ls: cannot access /etc/foo: No such file or directory
-rw-r--r--. 1 root root 122 Aug 25 14:49 /etc/resolv.conf

$ sudo docker run debian ls -l /etc/resolv.conf /etc/foo 2>/dev/null
-rw-r--r--. 1 root root 122 Aug 25 14:49 /etc/resolv.conf

$ sudo docker run --runtime cc debian ls -l /etc/resolv.conf /etc/foo
ls: cannot access /etc/foo: No such file or directory
-rw-r--r-- 1 root root 122 Aug 25 14:50 /etc/resolv.conf

$ sudo docker run --runtime cc debian ls -l /etc/resolv.conf /etc/foo 2>/dev/null
ls: cannot access /etc/foo: No such file or directory
-rw-r--r-- 1 root root 122 Aug 25 14:50 /etc/resolv.conf
```

devimc:
Fixed CC-2.1

Issue #134, opened 2016-08-19
Size of terminal not updating


dlespiau:
Not resizing the terminal between the two commands, obviously!

```
$ docker run -t debian sh -c 'echo size = \(`tput cols`, `tput lines`\)'
size = (195, 52)
$ docker run --runtime cc -t debian sh -c 'echo size = \(`tput cols`, `tput lines`\)'
size = (80, 24)
```

The size of the terminal is updated within containerd-shim, but doesn't work with the way we hook things. Possibly related to #133.

amyleeland:
Close this issue

dlespiau:
Part of #374. Closing.

Issue #18, opened 2016-07-18
Implement "exec"


jodh-intel:
The `exec` command is blocked on network support. Once networking is available, the function `clr_oci_vm_connect()` needs to be updated to remove the `g_critical ("not implemented yet")` call.

devimc:
I took a look to current implementation of exec and I noticed that it uses ssh to connect to VM, this is a good approach but what will happen with minimal container images (busybox) that has not sshd ?

gorozco1:
good catch @devimc, I think the ssh daemon will be provided by the mini-os when the clear-container is launched.

gorozco1:
@devimc has some updates about this issue

grahamwhaley:
now depends on #302

dlespiau:
As part of this, we need to document that while the exec will work, the things people may use the exec function for (like manual networking setup) may not necessary work because it will be running inside of the VM.

dlespiau:
Alright, time to close that one. Bugs found for exec should be opened as new issues (we've started already).

PR #357, opened 2016-10-28
Tests: Exec functional tests

GabyCT:
These tests are for the exec functionality in docker
and the terminal test will verify that the change
of the terminal size in a container is properly
done

Signed-off-by: Cervantes Tellez, Gabriela gabriela.cervantes.tellez@intel.com

### DIFF
--- a/tests/integration/docker/exec.bats
+++ b/tests/integration/docker/exec.bats
@@ -1,0 +1,70 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#Based on docker commands
+
+setup() {
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "modifying a container with exec" {
+	skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18)"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 50"
+	$DOCKER_EXE ps -a | grep "sleep 50"
+	$DOCKER_EXE exec -d containertest bash -c "echo 'hello world' > file"
+	$DOCKER_EXE exec containertest bash -c "cat /file"
+}
+
+@test "exec a container with privileges" {
+	skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18)"
+	$DOCKER_EXE run -d --name containertest ubuntu bash -c "sleep 30"
+	$DOCKER_EXE exec -i --privileged containertest bash -c "mount -t tmpfs none /mnt"
+	$DOCKER_EXE exec containertest bash -c "df -h | grep "/mnt""
+}
+
+@test "copying file from host to container using exec" {
+	skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18)"
+	content="hello world"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	echo $content | $DOCKER_EXE exec -i containertest bash -c "cat > /home/file.txt"
+	$DOCKER_EXE exec -i containertest bash -c "cat /home/file.txt" | grep "hello world"
+}
+
+@test "stdout forwarded using exec" {
+	skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18 and https://github.com/01org/cc-oci-runtime/issues/171)"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	$DOCKER_EXE exec containertest ls /etc/resolv.conf 2>/dev/null | grep "/etc/resolv.conf" 
+}
+
+@test "stderr forwarded using exec" {
+        skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18 and https://github.com/01org/cc-oci-runtime/issues/171)"
+        $DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+        if $DOCKER_EXE exec containertest ls /etc/foo >/dev/null; then false; else true; fi	
+}
+
+@test "check exit code using exec" {
+	skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18)"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	run $DOCKER_EXE exec containertest bash -c "exit 42"
+	[ "$status" -eq 42 ]
+}


tests/integration/docker/terminal.bats
```diff
@@ -0,0 +1,30 @@
...
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+setup() {
+        source test.common
+        cleanDockerPs
+}
+
+@test "changing terminal size test" {
+        skip "this functionality is not working"
```
jodh-intel: How about adding in a reference to `https://github.com/01org/cc-oci-runtime/issues/134` here so anyone seeing the skip message will know where to look.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,50 @@
...
+
+#Based on docker commands
+
+setup() {
+	source test.common
+	cleanDockerPs
+}
+
+@test "modifying a container with exec" {
+	skip "exec is still not working"
```
jodh-intel: Same comment as below - can you add a link to `https://github.com/01org/cc-oci-runtime/issues/18` here.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
...
+setup() {
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "modifying a container with exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -p 6001:6011 -d ubuntu bash -c "sleep 300"
```
dlespiau: You could drop the port mapping

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bats
```
dlespiau: You use both ubuntu and busybox images, could we just have one (say busybox?) no need to download both.

tests/integration/docker/exec.bats
dlespiau: The "sleep 300" will make the container run for 2:30 minutes. Also because the containers have the same names across the tests, so it's all not very clear what this will do?. Could we start the container in setup() and stop the container in teardown() for every test (using the setup() & teardown() test hooks, https://github.com/sstephenson/bats#setup-and-teardown-pre--and-post-test-hooks)

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
...
+	runtimeDocker
+}
+
+@test "modifying a container with exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -p 6001:6011 -d ubuntu bash -c "sleep 300"
+	$DOCKER_EXE ps -a | grep "Up"
+	$DOCKER_EXE exec -d containertest bash -c "echo 'hello world' > file"
+	$DOCKER_EXE exec containertest bash -c "cat /file"
```
dlespiau: Could you grep for the expected result? that will add the check we can send data to stdout.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
...
+}
+
+@test "exec a container with privileges" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	skip "exec is still not working"
+	$DOCKER_EXE run -d --name containertest busybox
```
dlespiau: you don't execute sleep here, is that a problem?

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
...
+	# The first command will just start the container and give enough time to run an exec
+	# workload meanwhile the second command will be blocked until the main workload
+	# finishes its execution
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 5"
+	$DOCKER_EXE exec containertest bash
+	if [ $? != "0" ]; then
+		exit 1
+	fi
+}
```
dlespiau: missing new line.

tests/integration/docker/terminal.bats
```diff
@@ -0,0 +1,32 @@
...
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "changing terminal size test" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/134
+	skip "this functionality is not working"
+	docker run -it ubuntu bash -c 'stty rows 50; stty cols 50; if [ `tput cols` != "50" ] && [ `tput lines` != "50" ]; then exit 1; fi'
+}
```
dlespiau: missing new line.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,58 @@
...
+	$DOCKER_EXE run -d --name containertest busybox
+	$DOCKER_EXE exec -i --privileged containertest sh -c "mkdir mnt && mount -t tmpfs none /mnt"
+}
+
+@test "run a container with exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	# This test will obtain the exit code from exec workload after "docker run" finishes
+	# The first command will just start the container and give enough time to run an exec
+	# workload meanwhile the second command will be blocked until the main workload
+	# finishes its execution
```
dlespiau: I'm not sure what this tests adds compared to the first one.

dlespiau:
Question: have you run those tests with runc to make sure they pass with it? It'd be helpful to rise the likeliness of those tests working.

GabyCT:
Yes @dlespiau  , I tested them using runc and they are working properly.

dlespiau:
Cool! just checking :)

dlespiau:
Missing tests:

  - Ensure we have the right exit value from docker exec (eg. docker exec /bin/sh -c 'exit 0' and docker exec /bin/sh -c 'exit 42')
  - Ensure stdout/err are properly handled see https://github.com/01org/cc-oci-runtime/issues/171
  - Ensure stdin is correctly sent to the container (write stdin to a file and check the output)

tests/integration/docker/terminal.bats
```diff
@@ -0,0 +1,32 @@
...
+setup() {
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "changing terminal size test" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/134
+	skip "this functionality is not working"
+	docker run -it ubuntu bash -c 'stty rows 50; stty cols 50; if [ `tput cols` != "50" ] && [ `tput lines` != "50" ]; then exit 1; fi'
```
dlespiau: Given that we use DOCKER_EXE everywhere else, might as well use it there.

tests/integration/docker/exec.bats
GabyCT: Fixed, now they are using ubuntu

GabyCT:
@dlespiau, I made all the changes and also I add new exec tests :) Thanks for the feedback

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,75 @@
...
+#Based on docker commands
+
+setup() {
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "modifying a container with exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
```
jodh-intel: Thanks @GabyCT - I'd actually put a reference to the URL in the skip message themselves. For example,

```
skip "exec is still not working (see https://github.com/01org/cc-oci-runtime/issues/18)"
```

That way, when you look at the log files, you can immediately understand why a test is being skipped and get further information if you need it.

dlespiau:
Still missing the stdin test:

```
echo "Hello, World!" | docker run -i --runtime runc busybox sh -c 'cat > /foo; cat /foo'
Hello, World!
```

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,75 @@
...
+	$DOCKER_EXE exec -i containertest bash -c "cat /home/file.txt" | grep "$content"
+	rm -rf file.txt
+}
+
+@test "stdout/stderr properly forwarded using exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	# and also with https://github.com/01org/cc-oci-runtime/issues/171
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	run $($DOCKER_EXE exec containertest ls -l /etc/resolv.conf /etc/foo 2>/dev/null)
```
dlespiau: That doesn't look correct. The goal here to to make sure we send stdout and stderr as distinct streams. We ls both a file that exists (generate the name of the file on stdout) and a file that doesn't (generate an error message on stderr) and we want to pipe either stdout or stderr to /dev/null to make sure we receive both stdout/err as 2 separate streams (and not both on stdout like today).

The correct tests/outputs:

```
$ docker run -i --runtime runc busybox ls -1 /etc/resolv.conf /etc/foo 2>/dev/null
/etc/resolv.conf
```

and 

```
$ docker run -i --runtime runc busybox ls -1 /etc/resolv.conf /etc/foo >/dev/null
ls: /etc/foo: No such file or directory
```

dlespiau:
Oh no, sorry, the stdin test is in there.

tests/integration/docker/terminal.bats
```diff
@@ -0,0 +1,33 @@
...
+setup() {
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "changing terminal size test" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/134
+	skip "this functionality is not working"
+	$DOCKER_EXE run -it ubuntu bash -c 'stty rows 50; stty cols 50; if [ `tput cols` != "50" ] && [ `tput lines` != "50" ]; then exit 1; fi'
```
dlespiau: Huum, what we want to do is to change the host terminal size and check that the container terminal size matches. That will test that we properly communicate the terminal size to the container.

What's done here is change the terminal size from within the container.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,75 @@
...
+	[[ ! "${output}" =~ "No such file or directory" ]]
+}
+
+@test "check exit code using exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	run $DOCKER_EXE exec containertest bash -c "exit 42"
+	echo $status
```
dlespiau: Oops, look like a debug message has been left here.

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,75 @@
...
+	$DOCKER_EXE run -d --name containertest ubuntu bash -c "sleep 30"
+	$DOCKER_EXE exec -i --privileged containertest bash -c "mount -t tmpfs none /mnt"
+	$DOCKER_EXE exec containertest bash -c "df -h | grep "/mnt""
+}
+
+@test "copying file from host to container using exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	skip "exec is still not working"
+	content="hello world"
+	echo "$content" > file.txt
```
dlespiau: Instead writing an intermediate file, could you just pipe $content to docker? That makes the test a bit simpler (fewer steps) and will work even if $PWD isn't writable (make distcheck gives me nightmares)

```
echo $content | docker exec -i...
```

tests/integration/docker/exec.bats
```diff
@@ -0,0 +1,75 @@
...
+@test "stdout/stderr properly forwarded using exec" {
+	# This problem is related with https://github.com/01org/cc-oci-runtime/issues/18
+	# and also with https://github.com/01org/cc-oci-runtime/issues/171
+	skip "exec is still not working"
+	$DOCKER_EXE run --name containertest -d ubuntu bash -c "sleep 30"
+	run $($DOCKER_EXE exec containertest ls -l /etc/resolv.conf /etc/foo 2>/dev/null)
+	[[ ! "${output}" =~ "No such file or directory" ]]
+}
+
+@test "check exit code using exec" {
```
dlespiau: Probably a good idea to put this test up in the chain. Elsewhere we rely  on docker exec returning the exit status, so better check that works early on.

tests/integration/docker/terminal.bats
GabyCT: This test will be changed for cc shim https://github.com/01org/cc-oci-runtime/issues/374

dlespiau:
lgtm

chavafg:
lgtm

chavafg:
qa-failed

PR #362, opened 2016-11-01
Shim implementation for master

amshinde:

PR #365, opened 2016-11-02
launch-shim-using-allocate-io

jcvenegas:
With this PR cc-shim is launched using api command allocateIO  from
proxy.

cc_proxy_cmd_allocate_io: ask proxy for a IO id to be used by
cc-shim and hyperstart for stdio and stderr streams and 
get proxy io fd needed by cc-shim to read streams data.

out-of-band data will be read in cc_proxy_msg_read and
send to cc-shim through cc_proxy_cmd_allocate_io

cc_proxy_hyper_new_container: Create a new container
with an initial workload and IO id's from allocate_io

socket_connection_from_fd:  gets a GSocketConnection from
a fd, GSocketConnection can be used to send files descriptors
between processes

change: cc_oci_fd_set_cloexec -> cc_oci_fd_toggle_cloexec
Allow to enable enable or disable cloexec flag in a fd
removed functions:
cc_oci_create_container_workload

Signed-off-by: Julio Montes julio.montes@intel.com
Signed-off-by: Jose Carlos Venegas Munoz jose.carlos.venegas.munoz@intel.com


src/oci.c
```diff
@@ -1827,3 +1705,51 @@ cc_oci_config_update (struct cc_oci_config *config,
 
 	return true;
 }
+
+/*!
+* Convert the config process to a JSON object.
+*
+* \param config \ref cc_oci_config.
```
jodh-intel: Should be:

```
\param process \ref oci_cfg_process.
```

jodh-intel:
@jcvenegas - Pullapprove has noticed that commit f7861bb is missing a required `Signed-off-by`.

src/process.c
```diff
@@ -562,18 +654,18 @@ cc_shim_launch (struct cc_oci_config *config,
 			goto child_failed;
 		}
 
-		/* +1 for for NULL terminator */
+		/* remove FD_CLOEXEC flags */
+		flags = fcntl(proxy_socket_fd, F_GETFD);
+		flags &= ~FD_CLOEXEC;
+		fcntl(proxy_socket_fd, F_SETFD, flags);
```
jodh-intel: I'd be tempted to rework `cc_oci_fd_set_cloexec()` to be:

```
gboolean
cc_oci_fd_toggle_cloexec (int fd, gboolean set)
```

so that you can call:

```
cc_oci_fd_toggle_cloexec (proxy_socket_fd, false);
cc_oci_fd_toggle_cloexec (proxy_io_fd, false);
```

src/process.c
```diff
@@ -478,6 +483,45 @@ cc_oci_vm_netcfg_get (struct cc_oci_config *config,
...
 
 /*!
+ * Create a socket connection from a fd.
+ *
+ * \param fd int.
+ *
+ * \return a GSocketConnection on success, else NULL.
+ */
+static GSocketConnection *
+socket_connection_from_fd (int fd)
```
jodh-intel: Needs a check for `fd >= 0`.

src/process.c
```diff
@@ -487,18 +531,22 @@ cc_oci_vm_netcfg_get (struct cc_oci_config *config,
...
+ *   send the proxy IO fd to child before \ref CC_OCI_SHIM is launched.
  *
  * \return \c true on success, else \c false.
  */
 static gboolean
 cc_shim_launch (struct cc_oci_config *config,
 		int *child_err_fd,
-		int *shim_args_fd)
+		int *shim_args_fd,
+		int *shim_socket_fd)
```
jodh-intel: Need to add `shim_socket_fd` to the args test below which currently checks `child_err_fd` and `shim_args_fd`.

src/proxy.c
```diff
@@ -881,3 +1089,141 @@ cc_proxy_hyper_pod_create (struct cc_oci_config *config)
...
+			config->oci.process.stderr_stream);
+
+	/* initial workload from config */
+	for (gchar** p = config->oci.process.args; p && *p; p++) {
+		json_array_add_string_element (args, *p);
+	}
+
+	for (gchar** p = config->oci.process.env; p && *p; p++) {
+		JsonObject *env_var = json_object_new ();
+		char *var = g_strdup(*p);
```
jodh-intel: You could use change `char *var` to `g_autofree gchar *var` to avoid the explicit `g_free()`.

src/util.c
```diff
@@ -559,6 +559,15 @@ cc_oci_enable_networking (void)
 	return enable;
 }
 
+/*!
+ * Convert the value stored in buffer to little endian
+ *
+ * \param buf Buffer storing the big endian value
```
jodh-intel: Needs a `\return` comment.

tests/test_common.c
```diff
@@ -231,6 +231,21 @@ test_helper_create_state_file (const char *name,
...
+	if (cc_oci_state_file_create (config, timestamp)) {
+		fprintf (stderr, "ERROR: cc_oci_state_file_create "
+				"worked unexpectedly for vm %s (no config->process)\n", name);
+		return false;
+	}
+
+	if (snprintf(config->oci.process.cwd, sizeof(config->oci.process.cwd),
+				"%s", "/working_directory") < 0) {
+		fprintf (stderr, "ERROR: cc_oci_state_file_create "
+				"failed to copy process cwd for vm %s", name);
```
jodh-intel: Missing `\n`.

src/proxy.c
```diff
@@ -216,60 +224,89 @@ static gboolean
...
+	gboolean fd_read = false;
+	struct msghdr msg = { 0 };
+	struct cmsghdr *cmsg = NULL;
+	struct iovec io = { .iov_base = iov_buffer,
+	                    .iov_len = sizeof(iov_buffer) };
+	char ctl_buffer[LINE_MAX] = { 0 };
+	unsigned char *data;
+	/* length + reserved */
+	const size_t data_offset = sizeof(p_msg.length) + sizeof(p_msg.reserved);
+	char buf_len[4] = { 0 };
```
jodh-intel: Can you add a comment explaining that this is sized to fit a 32-bit int.

It would be clearer and safer to do something like:

```
char buf_len[__SIZEOF_INT__] = { 0 };
```

src/proxy.c
```diff
@@ -216,60 +224,89 @@ static gboolean
...
+	size_t bytes_to_copy = 0;
+
+	msg.msg_iov = &io;
+	msg.msg_iovlen = 1;
+	msg.msg_control = ctl_buffer;
+	msg.msg_controllen = sizeof(ctl_buffer);
+
+	/* read length */
+	bytes_read = recvmsg(proxy_data->socket_fd, &msg, 0);
+	if (bytes_read > 0) {
```
jodh-intel: Could you invert the logic here to check `if  (bytes_read <= 0)` and error if true. That way, there is less indent which make the code simpler to read.

The same applies to `if (proxy_data->oob_fd)` and `if (bytes_read < p_msg.length+data_offset)`, etc, in this function.

src/oci.c
```diff
@@ -619,132 +619,6 @@ set_env_home(struct cc_oci_config *config)
...
-		g_autofree gchar  *envpath = NULL;
-		g_autofree gchar  *env = NULL;
-
-		envpath = g_build_path ("/", config->oci.root.path,
-				CC_OCI_ENV_FILE, NULL);
-		if (! envpath) {
-			return false;
-		}
-
-		set_env_home(config);
```
jodh-intel: This function can be dropped now as it is unused (along with the `test_set_env_home()`).

src/process.c
jcvenegas: fixed

src/util.c
jcvenegas: fixed

src/util.c
```diff
@@ -559,6 +559,17 @@ cc_oci_enable_networking (void)
...
+ *
+ * \return \c guint32 in big endian order.
+ */
+guint32 cc_oci_get_big_endian_32(const char *buf) {
+	return (guint32)(buf[0] >> 24 | buf[1] >> 16 | buf[2] >> 8 | buf[3]);
```
gorozco1: we got a coverity issue with this on previous commit related to #362 

> 
> CID 1375208 (#3-1 of 3): Bad bit shift operation (BAD_SHIFT)
> 1. large_shift: In expression buf[2] >> 8, right shifting buf[2] by more than 7 bits always yields either all zeros or all ones. The shift amount is 8.

` return (uint32_t)(buf[0] >> 24 | buf[1] >> 16 | buf[2] >> 8 | buf[3]);
`

amshinde: @gorozco1  @jcvenegas This needs to as
```
uint32_t guint32 cc_oci_get_big_endian_32(uint8_t *buf){
        return (uint32_t)(buf[0] << 24 | buf[1] << 16 | buf[2] << 8 | buf[3]);
}
```
Note the use of unsigned pointer in the parameter as well, required for correct bit operations. I had made a similar change in shim as well, but got lost in the many rebases I did with 2 different branches since then. Going to be pushing that in my next PR for shim.

src/oci.c
```diff
@@ -1800,6 +1673,11 @@ cc_oci_config_update (struct cc_oci_config *config,
 		state->mounts = NULL;
 	}
 
+	if(state->process) {
+		config->oci.process = *state->process;
```
jodh-intel: This looks like a memory leak as you are copying the dynamically-allocated `state->process` to `config`, but then just `NULL`-ing `state->process`.

src/process.c
```diff
@@ -170,29 +186,42 @@ cc_oci_setup_child (struct cc_oci_config *config)
...
 
+/*! Perform setup on spawned shim process.
+ *
+ * \param proxy_fd Proxy socket connection 
+ * \param proxy_io_fd Proxy IO fd
+ *
+ * \return \c true on success, else \c false.
+ */
+static gboolean
+cc_oci_setup_shim (int proxy_fd, int proxy_io_fd)
```
jodh-intel: Could you add a check on the fd params to ensure they are `>= 0`.

gorozco1:
@jcvenegas any update on this?

jcvenegas:
@jodh-intel Travis check failed , seems that the issue comes from the target branch (redesign-for-cc-2.1) : 
The command "for p in $go_packages; do golint -set_exit_status $p; done" exited with 1.

dlespiau:
Problem has been fixed in master. Someone needs to rebase the redesign-for-cc-2.1 branch on top of master.

jcvenegas:
PR update: 
remove qemu interactive console from slave pts
cc_oci_setup_shim make dup from salve pts to stdio fds

src/process.c
```diff
@@ -198,19 +200,45 @@ cc_oci_setup_child (struct cc_oci_config *config)
...
+	if (! config || proxy_fd < 0 || proxy_io_fd < 0) {
+		return false;
+	}
 
 	/* become session leader */
 	setsid ();
 
+	// In the console case, the terminal needs to be dup'ed to stdio
+	if (config->oci.process.terminal) {
+		tty_fd = open(config->console, O_RDWR |  O_NOCTTY);
```
jcvenegas: the runtime allows to empty console, needs checking  before open this file

jodh-intel:
Re-synced with `master`.

gorozco1:
@jodh-intel @dlespiau can you take a look at latest changes by @jcvenegas

gorozco1:
👍

PR #373, opened 2016-11-04
data: Add cc-agent.target

jcvenegas:
This patch adds cc-agent.target and cc-target.service. The target
starts the clear containers guest agent (hyperstart). After the agent is finished
the VM will be powered off.

The agent output is sent to tty for debug information. Can be retrieved
by using a serial console in the guest.

### DIFF
--- a/data/cc-agent.target
+++ b/data/cc-agent.target
@@ -1,0 +1,7 @@
+[Unit]
+Description=Clear Containers Agent Target
+Requires=basic.target
+Requires=cc-agent.service
+Conflicts=rescue.service rescue.target
+After=basic.target rescue.service rescue.target
+AllowIsolate=yes
--- a/Makefile.am
+++ b/Makefile.am
@@ -104,6 +104,8 @@ defaultsdir = $(datadir)/defaults/$(PACKAGE_NAME)
 defaults_DATA = data/vm.json data/hypervisor.args data/kernel-cmdline
 
 cc_image_systemd_files = \
+	data/cc-agent.target \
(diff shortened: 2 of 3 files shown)


jodh-intel:
lgtm.

data/cc-agent.target
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Clear Containers Agent Target
+Requires=basic.target
+Requires=cc-agent.service
+Conflicts=rescue.service rescue.target
+After=basic.target rescue.service rescue.target
+AllowIsolate=yes
```
dlespiau: Why do we need AllowIsolate here?

jcvenegas: I dont think that it is really needed, it allows to use "systemctl isolate" to change a current target to this target, but it is recommended to used for systemd targets:
 "It probably is a good idea to leave this disabled except for target units that shall be used similar to runlevels in SysV init systems, just as a precaution to avoid unusable system states."
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#AllowIsolate=

dlespiau: ah right, read that sentence before but reading it again now, it make more sense :)

dlespiau:
lgtm

amshinde:
@jcvenegas Minor thing, could you change the commit message to say cc-agent.service instead of cc-target.service.

chavafg:
qa-passed

devimc:
:+1:

PR #375, opened 2016-11-04
Fixes for cc-shim

amshinde:


shim/utils.c
```diff
@@ -118,9 +118,10 @@ set_big_endian_32(uint8_t *buf, uint32_t val)
  *
  * \return Unsigned 32 bit network ordered integer
  */
-uint32_t 
-get_big_endian_32(char *buf) {
-	return (uint32_t)(buf[0] >> 24 | buf[1] >> 16 | buf[2] >> 8 | buf[3]);
+uint32_t
+get_big_endian_32(uint8_t *buf)
```
jodh-intel: `const`?

shim/utils.c
```diff
@@ -144,7 +145,7 @@ set_big_endian_64(uint8_t *buf, uint64_t val)
  * \return Unsigned 64 bit network ordered integer
  */
 uint64_t
-get_big_endian_64(char *buf) {
+get_big_endian_64(uint8_t *buf) {
```
jodh-intel: `const`?

shim/shim.c
```diff
@@ -368,7 +368,7 @@ read_IO_message(struct cc_shim *shim, uint64_t *seq, ssize_t *stream_len) {
 			}
 		}
 	}
-	*seq = get_big_endian_64(buf);
+	*seq = get_big_endian_64((uint8_t*)buf);
```
dlespiau: I think it'd make sense to declare buf as uint8_t * instead of having to cast it?

amshinde: I have used string functions at some other places. I would have to cast to char* there.

dlespiau: Hum, I don't see any string function on buf in read_IO_message()?

dlespiau: but then the return value is used elsewhere in a atoi(buf). I'd argue the cast is probably better there (interpreting part of the byte stream as something else), but it doesn't really matter, let's drop this.

dlespiau:
lgtm

grahamwhaley:
lgtm
presuming @jodh-intel is happy with rework

chavafg:
qa-passed

PR #399, opened 2016-11-14
Networking with  hyperstart

amshinde:


src/hypervisor.c
```diff
@@ -52,35 +52,9 @@ cc_oci_expand_net_cmdline(struct cc_oci_config *config) {
...
-		g_slist_nth_data(if_cfg->ipv4_addrs, 0);
-
-	if (ipv4_cfg == NULL){
-		goto out;
-	}
-
-	return ( g_strdup_printf("ip=%s::%s:%s:%s:%s:off::",
-		ipv4_cfg->ip_address,
+	return ( g_strdup_printf("ip=:::%s:::%s::off::",
 		config->net.gateway,
```
jodh-intel: This function needs a check to ensure `config != NULL` before dereferencing. Also, I'd add checks on `gateway` and `hostname` to avoid `(null)` being returned as part of the string on error.

src/hypervisor.c
```diff
@@ -132,6 +108,33 @@ cc_oci_expand_net_device_cmdline(struct cc_oci_config *config, guint index) {
 	return g_strdup("");
 }
 
+/*!
+ * Append qemu options for networking
+ *
+ * config \ref cc_oci_config.
```
jodh-intel: Both args need a `\param` in the function header.

src/hypervisor.c
```diff
@@ -132,6 +108,33 @@ cc_oci_expand_net_device_cmdline(struct cc_oci_config *config, guint index) {
...
+ * additional_args Array that will be appended
+ */
+static void
+cc_oci_append_network_args(struct cc_oci_config *config,
```
jodh-intel: Need to check both params to ensure they are not `NULL`.

src/hypervisor.c
```diff
@@ -479,11 +450,13 @@ cc_oci_vm_args_get (struct cc_oci_config *config,
...
 	}
 
-	 /*  append additional args array */
-        for (int i = 0; i < extra_args_len; i++) {
-                gchar* arg = g_ptr_array_index(hypervisor_extra_args, i);
-                new_args[line_count++] = g_strdup(arg);
-        }
+	/*  append additional args array */
+	for (int i = 0; i < extra_args_len; i++) {
+		gchar* arg = g_ptr_array_index(hypervisor_extra_args, i);
```
jodh-intel: It would be clearer if `arg` were `const` here.

src/networking.c
```diff
@@ -431,6 +431,20 @@ get_mac_address(const gchar *const ifname)
...
 
+/*
+ * Return the predicatable interface name based on pcie address
+ * Reference:
+ * https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
+ *
+ * \param index Index into the array that was used for the qemu
+ * pcie address option as well
+ */
+gchar *
```
jodh-intel: Missing `\return`.

src/oci.h
```diff
@@ -137,6 +137,11 @@
 /* Path to the stateless passwd file. */ 
 #define STATELESS_PASSWD_PATH "/usr/share/defaults/etc/passwd"
 
+/* Offset to add to the interface index for assigning the pci slot.
+ * First 3 slots are in use for pc-lite machine type
+ */
+#define PCI_OFFSET 4
```
jodh-intel: Is there any likelihood that qemu might change the number of slots it assigns itself? What are the first 3 slots used for specifically?

This looks like a potential maintenance issue, but I'm not sure how we can protect against future changes here? Hopefully, if qemu did change, we'd get errors logged in the hypervisor logs, but can we do better than that?

jodh-intel:
I'm seeing the dreaded "generic hyperstart error" testing this branch:

```
debug:writing message data to proxy socket: {"id":"hyper","data":{"hyperName":"startpod","data":{"hostname":"cb8f97ef7cae7078e50440a7d9b60e53634dc2352f0aa749973118a31da4ad58","containers":[],"shareDir":"rootfs","interfaces":[{"device":"enp0s4","newDeviceName":"eth0","ipAddresses":[{"ipAddress":"172.17.0.2","netMask":"255.255.0.0"}]}]}}}
debug:message read from proxy socket: {"success":false,"error":"hyperstart returned an error"}

```

amshinde:
@jodh-intel Looks like your image does not have patch for multiple ip addresses. Looking at our package in Clear, this seems to have fallen through. I'll add that to the image.

src/oci.h
amshinde: @jodh-intel The pci slots currently in use are:
  PCI: slot 0 function 0 => in use by pci-lite-device
  PCI: slot 1 function 0 => in use by PM_LITE
  PCI: slot 2 function 0 => in use by virtio-9p-pci
  PCI: slot 3 function 0 => in use by virtio-serial-pci

The first two 0 and 1 seem to be used by qemu for the pc-lite machine type, the last two are used by us for the 9p mount and for the virtual console respectively. I'll be updating the comments to reflect this.

I am inclined to set the offset to a higher value (maybe 8) to take into account any available free slots being assigned by qemu or used by us in the future. What do you think?

jodh-intel: Sounds reasonable to me. @anthonyzxu - do you have any comments on this strategy?

@amshinde - how about adding a `g_debug()` call where the value of `PCI_OFFSET` is logged (in `cc_oci_expand_net_device_cmdline()`?), so that if there is a future clash with qemu-lite, atleast there is a log entry to aid in debugging.

grahamwhaley: We should correlate if this is at all affected/related to the PCI BAR scanning optimisations in the guest kernel as well - if you look at the guest kernel patches you'll find:

0006-probe-patch
407-pci-guest....

where we optimise the BAR scanning for the pc-lite machine, I believe to 4 slots presently, as we 'know' there is nothing beyond there ... or wasn't.

Indeed, it would be good to get @anthonyzxu input here.
(some of those two should really be merged ;-)

amshinde:
@jodh-intel I have addressed your review comments.

jodh-intel:
lgtm.

dlespiau:
Haven't gone into details, but the general flow looks good to me. Better have it in the branch for all to test than in a PR.

lgtm

chavafg:
qa-passed

PR #409, opened 2016-11-15
Travis refactoring

jodh-intel:


.ci/ci-build.sh
```diff
@@ -29,16 +29,54 @@ tarball=`ls -1 cc-oci-runtime-*.tar.xz`
...
+
+# go checks
+export go_packages=$(go list ./... |\
+    grep -v cc-oci-runtime/vendor |\
+    sed -e 's#.*/cc-oci-runtime/#./#')
+
+go list -f '{{.Dir}}/*.go' "$go_packages" |\
+    xargs -I % bash -c "misspell -error %"
+
+go vet "$go_packages"
```
dlespiau: Quoting variables isn't always the right thing to do :)

```
$ go_packages=$(go list ./... | grep -v cc-oci-runtime/vendor | sed -e 's#.*/cc-oci-runtime/#./#')
$ echo $go_packages 
./proxy ./proxy/api ./tests/mock
$ go vet $go_packages
$ go vet "$go_packages"
can't load package: package ./proxy
./proxy/api
./tests/mock: open /home/damien/clear/sources/cc-oci-runtime/proxy
./proxy/api
./tests/mock: no such file or directory
```

(go vet will receive the whole list as a single argument instead of 3)

dlespiau:
(I'm sure you're on it already but it seems I didn't put a -x (/bin/bash -e -x) in -build.sh, it could be handy to trace the execution of the script)

jodh-intel:
yes, I assumed you'd left that out to avoid bloating the travis logs but I'll prolly add it back in temporarily atleast.

devimc:
lgtm

jodh-intel:
This branch now works with TravisCI and also with SemaphoreCI. However, there is an unresolved issue when run under SemaphoreCI that causes the build to freeze on the docker/integration tests, then mysteriously fail the build after some timeout.

I've reported this issue to Semaphore support who are investigating. At this stage, it seems like an issue with the SemaphoreCI environment itself (resource limit/quota maybe) rather than with this code (Semaphore made some changes recently and a few users are reporting similar-sounding issues).

Since we have a few changes that will land in `master` soon that will require this branch to be reworked yet again, I'd like to suggest we land this soon, and I'll then raise any follow-up PRs required to resolve the SemaphoreCI issue.

jodh-intel:
To be clear, once this lands, we will still continue to (only) use TravisCI, atleast until the SemaphoreCI issue is resolved.

.ci/ci-setup.sh
```diff
@@ -19,8 +19,7 @@
 
 set -e -x
 
-root=$(cd `dirname "$0"/..`; pwd -P)
```
dlespiau: Ah, it seems like the error was the end '`', it should have been:

```
root=$(cd `dirname "$0"`/..; pwd -P)
```

It may be useful to keep that to be able to invoke the script from any directory, not just the project root directory

.ci/ci-setup.sh
```diff
@@ -57,19 +57,52 @@ export CC=$(which "$CC")
...
 function compile {
 	name="$1"
 	tarball="$2"
 	directory="$3"
+	configure_opts="$4"
 
 	chronic tar -xvf ${tarball}
 	pushd ${directory}
-	chronic	./configure --disable-silent-rules
+	eval chronic ./configure --disable-silent-rules "$configure_opts"
```
dlespiau: I'm really not sure about all those evals, why not just use $configure_opts. Quoting strikes back again :)

dlespiau:
"evals" always trigger warnings that we may be doing something odd in my mind (in any scripting language really), so with that hopefully changed:

lgtm

.ci/ci-setup.sh
jodh-intel: Fixed.

jodh-intel:
I understand your view on `eval`, but it's pretty much required if you quote paths (which you really need to do to handle craziness like spaces in paths). We could remove the eval and a layer of quotes, but then we'd be less able to deal with such scenarios so I think the `eval` calls are the lesser evil atm ;(

chavafg:
qa-failed

chavafg:
@jodh-intel , I checked logs and only failure is `not ok 42 start and kill state`, which is the test that fails randomly, moving to passed

chavafg:
qa-passed

jodh-intel:
@chavafg - thanks. Yes, that failure looks like #327.

devimc:
lgtm

dlespiau:
lgtm

jodh-intel:
@chavafg - could you re-ack this PR as pullapprove seems to have got confused :(

chavafg:
qa-passed

PR #410, opened 2016-11-15
proxy: Add a "containerId" argument to Bye

dlespiau:
Archana suggested we add a containerId parameter to Bye so we don't have
to issue an attach payload to associate the connection beforehand. That
saves a round-trip with the proxy.

Because it's still early times and we don't have any user, let alone any
stable user, yet, I broke the API: the old behaviour using Bye with no
parameter when the client is already attached to a VM will not work
anymore.

### DIFF
--- a/proxy/api/api.go
+++ b/proxy/api/api.go
@@ -53,13 +53,17 @@ type Attach struct {
 }
 
 // The Bye payload does the opposite of what hello does, indicating to the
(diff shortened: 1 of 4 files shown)


amshinde:
lgtm

gorozco1:
👍

chavafg:
qa-failed

PR #411, opened 2016-11-16
cc-shim: Use simple assignment to get the exit code from buffer

amshinde:
Since the exit code is a single char byte, use simple assignment to get
the exit code instead of atoi() call.


dlespiau:
lgtm

gorozco1:
lgtm

chavafg:
qa-failed

chavafg:
known false negative on test: `not ok 42 start and kill state`

chavafg:
qa-passed

PR #423, opened 2016-11-21
Tests: Unbreak bats error handling.

jodh-intel:
Previously, if certain bats helper functions failed, they would log a
message and "exit 1". This was incorrect since "exit 1" caused the
script to exit, meaning the bats framework was unaware that an error
had occured. The result was that no error output would be produced for
failing tests.

Replacing the "exit 1" with a call to false(1) results in the correct
behaviour since bats notices the failing assertion triggered by the
false result and displays the error as expected, before failing the
test.

### DIFF
--- a/tests/functional/common.bash.in
+++ b/tests/functional/common.bash.in
@@ -189,7 +189,7 @@ function verify_runtime_dirs()
 
 	else
 		log_msg "Invalid state: '$state'"
-		exit 1
+		false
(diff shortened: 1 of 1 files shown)


grahamwhaley:
lgtm

dlespiau:
lgtm

chavafg:
qa-failed

PR #430, opened 2016-11-22
Functional tests for the most popular docker images from docker hub

GabyCT:


gorozco1:
@GabyCT can you add a tiny readme on how to run these tests?

tests/integration/docker_images/common.bash
```diff
@@ -0,0 +1,44 @@
...
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+DOCKER_EXE="docker"
+DOCKER_SERVICE="docker-cor"
+
+#Cleaning test environment
+function cleanDockerPs(){
```
jodh-intel: This file looks identical to `tests/integration/docker/common.bash`. I think it would make sense to:

- Move `tests/integration/docker/common.bash` to `tests/integration/lib/common.bash`.
- Make the `docker` tests and the `docker_images` tests both use `tests/integration/lib/common.bash`

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -0,0 +1,405 @@
...
+	load common
+	cleanDockerPs
+	runtimeDocker
+}
+
+@test "run hello world in java in an openjdk container" {
+	$DOCKER_EXE run --rm -i openjdk bash -c "echo 'public class HW{public static void main(String[]a){System.out.println(\"HelloWorld\");}}' > HW.java && javac HW.java && java HW"
+}
+
+@test "check kong configuration file is valid" {
```
jodh-intel: For consistency, it would be good if all the test names had the image name in a fix place. You've mostly done that but this test name doesn't end with `$image_name container`.

Aside: I'd personally prefer this format:

```
@test "$image_name container: $description_of_test_being_run"
```
but this isn't a blocker for merging this PR :-)

jodh-intel: ... Having said that, if you did adopt the format above, you could sort the tests alphabetically by container name.

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -0,0 +1,405 @@
...
+@test "check os-core-update bundle in a clearlinux container" {
+	$DOCKER_EXE run --rm -i clearlinux bash -c "ls /usr/share/clear/bundles | grep "os-core-update""
+}
+
+@test "run golang container" {
+	$DOCKER_EXE run --rm -i golang bash -c "echo -e 'package main\nimport \"fmt\"\nfunc main() { fmt.Println(\"hello world\")}' > hello-world.go && go run hello-world.go && go build hello-world.go"
+}
+
+@test "create files in an oraclelinux container" {
+	$DOCKER_EXE run --rm -i oraclelinux bash -c 'for NUM in `seq 1 1 10`; do touch $NUM-file.txt; done; ls -l /1-file.txt'
```
jodh-intel: You could make that just `seq 10`. But I'm not clear why you are creating 10 files then checking the first only? Could this test be simple:

- create file
- check it exists?

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -0,0 +1,405 @@
...
+@test "check the resolv.conf in a crux container" {
+	$DOCKER_EXE run --rm -i crux bash -c "cat /etc/resolv.conf | grep "nameserver""
+}
+
+@test "find the timestamp help in a gazebo container" {
+	$DOCKER_EXE run --rm -i gazebo gz log --help
+}
+
+@test "go env in a golang container" {
+	$DOCKER_EXE run --rm -i golang bash -c "golang go env | grep "GOARCH""
```
jodh-intel: The quoting on `GOARCH` looks wrong? You can drop the quotes around `GOARCH` itself anyway as it's a static string.

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -0,0 +1,405 @@
...
+@test "run python command in a hylang container" {
+	$DOCKER_EXE run --rm -i hylang bash -c "python -m timeit -s \"L=[]; M=range(1000)\" \"for m in M: L.append(m*2)\""
+}
+
+@test "display config information of a influxdb container" {
+	$DOCKER_EXE run --rm -i influxdb influxd config | grep "dir"
+}
+
+@test "search nano in an ubuntu upstart container" {
+	$DOCKER_EXE run --rm -i ubuntu-upstart bash -c "apt-cache search nano | grep "editor""
```
jodh-intel: Yay, upstart! :-)

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -0,0 +1,405 @@
...
+
+@test "run python command in a hylang container" {
+	$DOCKER_EXE run --rm -i hylang bash -c "python -m timeit -s \"L=[]; M=range(1000)\" \"for m in M: L.append(m*2)\""
+}
+
+@test "display config information of a influxdb container" {
+	$DOCKER_EXE run --rm -i influxdb influxd config | grep "dir"
+}
+
+@test "search nano in an ubuntu upstart container" {
```
jodh-intel: Yay, upstart! :-)

tests/integration/docker_images/popular-containers-images.bats
jodh-intel: This test (and some of the other tests which run `grep`) look like they might be fragile - what if `LANG` is not set to a `C` locale and the command being run is internationalised?

A quick solution would be to add something like this to the top of `popular-containers-images.bats`:

```
export LANG=C
export LC_ALL=C
```

jodh-intel:
Hi @GabyCT - this is great to see!

I like the way you've added specific application tests for "specialist images".

However, for the more general-purpose images, I wonder if we need a slightly different approach: some of the tests look like they might be fragile in the medium term (if the images change the contents of file, delete files, etc).

Rather than inventing different tests for each generic image, we could run a few, very simple tests on all of them. Taking two tests based on some you're already specifying we could run the following:

- touch a file and check it is created.
- check a "well-known" and stable system configuration file (something like `grep NAME /etc/os-release`). Of course, even that isn't necessarily reliable (think "stateless" :-)

The 2 tests above could be set in a variable and passed to `docker run -ti $image $commands_to_run` for each generic image.

GabyCT:
@jodh-intel thanks for all the feedback, I did the changes and I put the common.bash at /tests/integration/lib as you suggested. In my new request you will find the tests for the most popular images but also for the docker integration tests with the modifications that both use the common.bash from /tests/integration/lib. Thanks

jodh-intel:
Hi @GabyCT - thanks. I'm happy for us to merge this for now, but could you raise an issue to capture the generic tests changes for the future?

lgtm.

dlespiau:
lgtm

chavafg:
qa-failed

chavafg:
There was a rebase in this PR, @dlespiau , @jodh-intel can you approve again?

jodh-intel:
lgtm.

dlespiau:
lgtm

chavafg:
qa-passed

PR #441, opened 2016-11-23
config: Provide full path to the cc-shim for invoking it.

amshinde:
cc-shim is installed under /usr/libexec as it is not meant
to be started on its own.

### DIFF
--- a/src/oci.h
+++ b/src/oci.h
@@ -97,7 +97,7 @@
  *  - Reacts to signals.
  *  - Exits with return code of real workload process.
  */
-#define CC_OCI_SHIM 			"cc-shim"
+#define CC_OCI_SHIM                     LIBEXECDIR"/cc-shim"
 
 /** Full path to socket used to talk to \ref CC_OCI_PROXY. */
 #define CC_OCI_PROXY_SOCKET 		CC_OCI_RUNTIME_DIR_PREFIX \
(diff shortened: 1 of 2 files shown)


src/oci.h
```diff
@@ -97,7 +97,7 @@
  *  - Reacts to signals.
  *  - Exits with return code of real workload process.
  */
-#define CC_OCI_SHIM 			"cc-shim"
```
dlespiau: Actually that path is given by configure, it's $libexecdir (it can be configured in ./configure and the prefix affects it for instance). You should probably provide a define from the compiler flags in Makefile.am

```
cc_shim_CPPFLAGS = -DLIBEXECDIR=\"$(libexecdir)\"
```

and in the code you can use the C string concatenation

```
#define CC_OCI_SHIM        LIBEXECDIR"/cc-shim"
````

dlespiau:
lgtm

jcvenegas:
does it will work properly for functional tests ?

We can run functional test without  install. It will run cc-shim in that path that is either a wrong binary (updated) or not installed.  It is not a way to relay in the PATH env variable.

amshinde:
@jcvenegas how about run install before running functional tests?

jodh-intel:
@amshinde - that's fine for running in a CI (disposable) environment, but we must not do that on a normal system.

dlespiau:
- Isn't that the case for more than the shim? proxy is the same?
- We can't run the integration tests either without a make install, so we need to do it anyway.
- I'd still rather not have cc-shim/cc-proxy in the PATH

Now let's fish for potential solutions:

- we could have an environment variable that let us override the shim path in the runtime.
- ...

jodh-intel:
Agreed. An env var would be acceptable for the testing scenario as long as we have a `g_debug()` call to log the fact that it's being made use of.

jcvenegas:
@jodh-intel  so is it a approved by you, and we can create a issue to add a env variable to honor for testing?

jodh-intel:
@jcvenegas - yes. Maybe we should have an env var for proxy and a separate one for shim. But alternatively, I wonder if it would be safer though to add explicit command-line options for the shim and the proxy since a simply `ps` would make it very obvious what was running.

lgtm.

chavafg:
qa-failed

Issue #459, opened 2016-11-24
cc-proxy fails if /var/run/cc-oci-runtime is missing


sameo:
After a fresh reboot, cc-oci-runtime never started and I get the following error:

```
sudo -E ./cc-proxy -v 4 --alsologtostderr
init: couldn't create AF_UNIX socket: listen unix /var/run/cc-oci-runtime/proxy.sock: bind: no such file or directory
```

I guess cc-proxy should create /var/run/cc-oci-runtime/ if it's not present.

jodh-intel:
This could be fixed by adding a `RuntimeDirectory=` to `proxy/cc-proxy.service.in`.

dlespiau:
Not quite :) systemd creates the directory for us already in the socket activation case (IIRC). The problem is when started the proxy from the command line and this directory doesn't exists. I'll have a patch in a few minutes.

jodh-intel:
ah, right - I always run it manually :)

PR #461, opened 2016-11-25
proxy: Create runtime directory if needed when not started by systemd

dlespiau:
It's expected that cc-proxy is socket activated by systemd in
production, but, the proxy also has a path where it creates the socket
itself. This is useful for debugging as it allows people to launch the
proxy from the command line directly.

Unfortunately, we were not able to create the socket in that second
path when the /run/cc-oci-runtime/ directory wasn't already created
(systemd will do that for us in the socket activated case).

Fix this by creating the /run/cc-oci-runtime directory if needed.

Fixes: https://github.com/01org/cc-oci-runtime/issues/459

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 
@@ -261,6 +262,10 @@ func (proxy *proxy) init() error {
 			return fmt.Errorf("couldn't listen on socket: %v", err)
 		}
 	} else {
+		socketDir := filepath.Dir(socketPath)
+		if err = os.MkdirAll(socketDir, 0750); err != nil {
+			return fmt.Errorf("couldn't create socket directory: %v", err)
+		}
 		if err = os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
 			return fmt.Errorf("couldn't remove exiting socket: %v", err)
 		}


proxy/proxy.go
```diff
@@ -224,6 +225,10 @@ func (proxy *proxy) init() error {
 			return fmt.Errorf("couldn't listen on socket: %v", err)
 		}
 	} else {
+		socketDir := filepath.Dir(socketPath)
+		if err = os.MkdirAll(socketDir, 0755); err != nil {
```
jodh-intel: Could this be `0750`?

dlespiau: I think so, let me try.

dlespiau:
Updated PR:

- Create the directory with 0750.

jodh-intel:
lgtm.

amshinde:
lgtm

chavafg:
qa-failed

dlespiau:
Hum we have an interesting looking error:

```
# (in test file commit.bats, line 33) 
# `$DOCKER_EXE run -i --name container1 busybox /bin/sh -c "echo hello"' failed with status 125 
# docker: Error response from daemon: json: cannot unmarshal number into Go value of type runtime.message. 
```

dlespiau:
PR Updated:

  - Rebased on top of master to get the vendoring

sameo:
LGTM

gorozco1:
lgtm

gorozco1:
@chavafg this is not being tested?

dlespiau:
Ok, the failures in CI were due to virtcontainers not being vendored yet at the time this branch was tested. With that in mind, merging.

PR #462, opened 2016-11-25
proxy: Restrict access to the proxy socket a bit

dlespiau:
When not socket activated, the proxy creates the socket itself. That
socket doesn't need to be world rw, so remove those permissions.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -268,7 +268,7 @@ func (proxy *proxy) init() error {
 		if err != nil {
 			return fmt.Errorf("couldn't create AF_UNIX socket: %v", err)
 		}
-		if err = os.Chmod(socketPath, 0666|os.ModeSocket); err != nil {
+		if err = os.Chmod(socketPath, 0660|os.ModeSocket); err != nil {
 			return fmt.Errorf("couldn't set mode on socket: %v", err)
 		}
 	}


proxy/proxy.go
```diff
@@ -231,7 +231,7 @@ func (proxy *proxy) init() error {
 		if err != nil {
 			return fmt.Errorf("couldn't create AF_UNIX socket: %v", err)
 		}
-		if err = os.Chmod(socketPath, 0666|os.ModeSocket); err != nil {
+		if err = os.Chmod(socketPath, 0660|os.ModeSocket); err != nil {
 			return fmt.Errorf("couldn't set mode on socket: %v", err)
```
jodh-intel: The change itself looks reasonable, but this code looks racy.

dlespiau: This commit doesn't make it worse than before, so we shouldn't block it on that.  Given that it should only be a debugging path, not terribly critical I'd say.

I'm open to suggestions though!

jcvenegas:
lgtm

chavafg:
qa-failed

chavafg:
seems that this one needs rebase to pass the tests

dlespiau:
PR rebased.

gorozco1:
👍

PR #472, opened 2016-11-28
[DONOTMERGE] [RFC] [WIP] CRI-O integration

sameo:


src/oci.c
```diff
@@ -142,6 +143,24 @@ cc_oci_get_bundlepath_file (const gchar *bundle_path,
 	return g_build_path ("/", bundle_path, file, NULL);
 }
 
+/**
+ * Get the workload directory for a given container.
+ * For pod sandboxes, this is the sandbox workloads directory,
+ * while for regular containers this is the OCI root path.
+ *
+ * \param config Container configuration.
```
jodh-intel: Needs a `\return` comment statement.

src/pod.c
```diff
@@ -0,0 +1,250 @@
...
+#include <glib.h>
+#include <gio/gunixconnection.h>
+
+#include "pod.h"
+#include "process.h"
+#include "proxy.h"
+#include "state.h"
+
+int
+cc_pod_handle_annotations(struct cc_oci_config *config, struct oci_cfg_annotation *annotation)
```
jodh-intel: Could you add a header with Doxygen comments for this function?

src/pod.c
```diff
@@ -0,0 +1,250 @@
...
+{
+	/* We only handle CRI-O/ocid annotations for now */
+	if (strncmp(annotation->key, CC_POD_OCID_ANNOTATION_NAMESPACE,
```
jodh-intel: Need to check `config` and `annotation` before dereferencing.

jodh-intel: I'm not totally sure why we need `CC_POD_OCID_ANNOTATION_NAMESPACE_SIZE`? If that was dropped, you could use `g_strcmp0()`.

src/pod.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <stdbool.h>
+
+#include <glib.h>
+
+#include "util.h"
+#include "oci.h"
+
+#define CC_POD_OCID_ANNOTATION_NAMESPACE "ocid/"
+#define CC_POD_OCID_ANNOTATION_NAMESPACE_SIZE 5
```
jodh-intel: Do we really need this? It's only referenced once and we could just use `g_strcmp0()` since `CC_POD_OCID_ANNOTATION_NAMESPACE` is bounded anyhow?

src/proxy.c
```diff
@@ -1163,39 +1163,15 @@ cc_proxy_hyper_pod_create (struct cc_oci_config *config)
...
+				  const char *container_id,
+				  char *rootfs)
 {
 	JsonObject *newcontainer_payload= NULL;
 	JsonObject *process = NULL;
 	JsonArray *args= NULL;
 	JsonArray *envs= NULL;
-	gboolean ret = false;
-
-	if (! (config && config->proxy)) {
```
jodh-intel: We still need this check on the params.

src/state.c
```diff
@@ -403,6 +405,46 @@ handle_state_proxy_section(GNode* node, struct handler_data* data) {
...
+
+	if (! (node && node->data)) {
+		return;
+	}
+	if (! (node->children && node->children->data)) {
+		g_critical("%s missing value", (char*)node->data);
+		return;
+	}
+
+	g_assert (data->state);
```
jodh-intel: No check on `data` before dereferencing. Also, an we convert the assert to a test and call to `g_critical()`? (We're trying to phase out asserts wherever possible).

src/state.c
```diff
@@ -403,6 +405,46 @@ handle_state_proxy_section(GNode* node, struct handler_data* data) {
...
+
+	pod = data->state->pod;
+
+	g_assert (pod);
```
jodh-intel: Same comment as above.

jodh-intel:
This is great - just a few minor issues.

The only other comment is that there is no `src/pod_test.c` to test `cc_pod_new_container()`.

As a minimum, this could just test the failure scenarios (like being passed an invalid/`NULL` `config` object), but ideally would exercise the function when run normally.

An alternative initial approach might be a bats test (`tests/functional/`).

src/oci.c
sameo: Fixed.

chavafg:
qa-failed

sameo:
Changing branch...

PR #473, opened 2016-11-28
shim: Fix stdin handling

dlespiau:
We were doing a few things wrong:

  - The local pty handed out by docker hasn't been configured to do
    anything special, it's a "default" canonical terminal with the line
    discipline configured to line buffer input, display a local echo and
    send signals on ctrl+c, ...

  - We had ported a bit of code from runv to locally buffer the input on
    top of the above. Unfortunately that wasn't helping and was intended
    for something else in runv (telnet client for debugging)

So, the strategy to solve the input with cc-shim is as follow:

  - Configure the local pty in raw mode. The goal here is to just
    forward the raw content of stdin to the pty allocated for the
    process inside the VM. The process inside the VM will have the pty
    there configured as it wishes (canonical Vs raw) and we bypass the
    treatment of the line discipline on the host side.

  - We try to read BUFSIZE data on stdin. In raw mode (pty allocation),
    the read() can return with only 1 byte of data, ie. 1 key press.
    When using stdin to pipe data to the process, we should benefit from
    buffering. I've also removed an unnecessary copy between buf and
    wbuf.

### DIFF
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -279,59 +279,31 @@ handle_signals(struct cc_shim *shim) {
 void
 handle_stdin(struct cc_shim *shim)
 {
-	ssize_t      nread;
-	int          ret;
-	ssize_t      len;
-	static char  buf[BUFSIZ-STREAM_HEADER_SIZE] = {0};
-	static char  wbuf[BUFSIZ] = {0};
-	static bool  cr  = false;
-	static bool  emit = false;
-	static int   cnt = 0;
-	char         ch;
+	ssize_t        nread;
+	int            ret;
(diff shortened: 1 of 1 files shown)


dlespiau:
The failures are mismatches between virtcontainers and the proxy code (so it needs the vendoring PR we have in flight)

dlespiau:
Updated PR:

  - Rebase on top of the virtcontainers vendoring to have the tests pass

sameo:
LGTM

chavafg:
qa-failed

amshinde:
Tested this. Works well :)

lgtm.

PR #476, opened 2016-11-29
console: Fix missing prompt when bash is run interactively.

amshinde:
When we run bash in interactive mode under docker, the prompt is
not shown. We can see the prompt for other shells.

A quick test using this:
https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html

shows that bash is not running in interactive mode causing the
systemwide startup files not being read and settings not
being applied resulting in PS1 variable to be empty.

The shell is not run in interactive mode because stderr is not
pointing to a terminal even though stdout and stdin are attached to
a tty by hypertstart. The reason being, hyperstart does not attach
the terminal to tty when the stderr sequence number is not set to 0.

Hence, allocate just one stream for the tty case, and explicitly set the
error sequence number to 0 with the "newcontainer" command.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -767,8 +767,10 @@ cc_shim_launch (struct cc_oci_config *config,
 		args[6] = g_strdup_printf ("%d", proxy_io_fd);
 		args[7] = g_strdup ("-s");
 		args[8] = g_strdup_printf ("%d", proxy_io_base);
-		args[9] = g_strdup ("-e");
-		args[10] = g_strdup_printf ("%d", proxy_io_base + 1);
+		if ( ! config->oci.process.terminal) {
+			args[9] = g_strdup ("-e");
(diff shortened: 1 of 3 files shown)


sameo:
LGTM

dlespiau:
lgtm

chavafg:
qa-failed

dlespiau:
Hum we have an interesting looking error:

```
# (in test file commit.bats, line 33) 
# `$DOCKER_EXE run -i --name container1 busybox /bin/sh -c "echo hello"' failed with status 125 
# docker: Error response from daemon: json: cannot unmarshal number into Go value of type runtime.message. 
```

amshinde:
@dlespiau Did a manual test with docker. Works. Seeing this because of proxy not running for the docker tests.

dlespiau:
Ok, I'm happy with that, full sails ahead!

PR #478, opened 2016-11-29
create: Set the hostname to the one provided by docker in the network…

amshinde:
… namespace.

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -1082,7 +1082,7 @@ cc_proxy_hyper_pod_create (struct cc_oci_config *config)
 	JsonObject                   *ipaddr_obj = NULL;
 	JsonArray                    *ipaddr_arr = NULL;
 
-	if (! (config && config->proxy)) {
+	if (! (config && config->proxy && config->net.hostname)) {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


src/proxy.c
```diff
@@ -1082,15 +1082,15 @@ cc_proxy_hyper_pod_create (struct cc_oci_config *config)
 	JsonObject                   *ipaddr_obj = NULL;
 	JsonArray                    *ipaddr_arr = NULL;
 
-	if (! (config && config->proxy)) {
+	if (! (config && config->proxy && config->net.hostname)) {
```
sameo: @amshinde Are we guaranteed to get a networking stanza and a hostname ?

chavafg:
qa-failed

src/proxy.c
amshinde: @sameo Yes, the network setup runs before pod creation. 
https://github.com/01org/cc-oci-runtime/blob/576913626eeefd9d85cd028fe459a9d666f076f5/src/networking.c#L551

jcvenegas:
lgtm

sameo:
LGTM

PR #482, opened 2016-11-29
Functional test for swarm

GabyCT:
It will verify that each hostname of the replicas is always different.

### DIFF
--- a/tests/integration/docker/swarm.bats
+++ b/tests/integration/docker/swarm.bats
@@ -1,0 +1,80 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# Swarm testing
+
+SRC="${BATS_TEST_DIRNAME}/../../integration/lib/"
+# maximum number of replicas that will be launch
+number_of_replicas=15
+url=http://127.0.0.1:8080/hostname
+# number of attemps to obtain the hostname
+# of the replicas using curl
+number_of_attemps=5
+declare -a REPLICAS
+
+setup() {
+	source $SRC/common.bash
+	cleanDockerPs
+	runtimeDocker
+	$DOCKER_EXE swarm init
+}
+
+@test "check that the replicas' names are different" {
+	# currently using mcastelino/nginx but it will be
+	# modified when nginx image is working
+	$DOCKER_EXE service create --name testswarm --replicas $number_of_replicas --publish 8080:80 mcastelino/nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\"" 2> /dev/null
+	while [ `docker ps -a | tail -n +2 | egrep "Up [0-9]*"|  wc -l`  -lt  $number_of_replicas ]; do
+		sleep 1
+	done
+	# this will help to obtain the hostname of 
+	# the replicas from the curl
+        unset http_proxy
+        for j in `seq 0 $number_of_attemps`; do
+	        for i in `seq 0 $((number_of_replicas-1))`; do
+		        # this will help that bat does not exit incorrectly 
+			# when curl fails in one of the attemps
+			set +e
+		        REPLICAS[$i]="$(curl $url 2> /dev/null)"
+	                set -e
+                done
+		non_empty_elements="$(echo ${REPLICAS[@]} | egrep -o "[[:space:]]+" | wc -l)"
+                if [ "$non_empty_elements" == "$((number_of_replicas-1))" ]; then
+                	break
+                fi
+		# this will give enough time between attemps
+                sleep 5
+        done
+	for i in `seq 0 $((number_of_replicas-1))`; do
+                if [ -z "${REPLICAS[$i]}" ]; then
+                    false
+                fi
+		for j in `seq $((i+1)) $((number_of_replicas-1))`; do
+			if [ "${REPLICAS[$i]}" == "${REPLICAS[$j]}" ]; then
+				false
+			fi
+		done
+	done 
+}
+
+teardown () {
+	$DOCKER_EXE service remove testswarm
+	$DOCKER_EXE swarm leave --force
+}


tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,64 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# Swarm testing
+
+SRC="${BATS_TEST_DIRNAME}/../../integration/lib/"
+number_of_replicas=12
```
jodh-intel: Why 12? Could you add a comment.

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,64 @@
...
+@test "check that the replicas' names are different" {
+	$DOCKER_EXE service create --name testswarm --replicas $number_of_replicas --publish 8080:80 mcastelino/nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\"" 
+	while [ `docker ps -a | tail -n +2 | egrep "Up [0-9]*"|  wc -l`  -lt  $number_of_replicas ]; do
+		sleep 1
+	done
+	declare -a REPLICAS
+	for i in `seq 0 $((number_of_replicas-1))`; do
+		unset http_proxy
+		set +e
+		for j in `seq 0 5`; do
```
jodh-intel: Is this for a retry? If so, you could use `curl --retry` to make the logic clearer.

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,64 @@
...
+
+setup() {
+	source $SRC/common.bash
+	cleanDockerPs
+	runtimeDocker
+	$DOCKER_EXE swarm init
+}
+
+@test "check that the replicas' names are different" {
+	$DOCKER_EXE service create --name testswarm --replicas $number_of_replicas --publish 8080:80 mcastelino/nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\""
```
jodh-intel: Can we rely on `mcastelino/nginx` as being:

- always available?
- updated?

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,64 @@
...
+	declare -a REPLICAS
+	for i in `seq 0 $((number_of_replicas-1))`; do
+		unset http_proxy
+		set +e
+		for j in `seq 0 5`; do
+			REPLICAS[$i]="$(curl http://127.0.0.1:8080/hostname)"
+			if [ $? == 0 ]; then
+				break
+			fi
+			sleep 2
```
jodh-intel: Why is the `sleep` required (and why `2` seconds)?

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,64 @@
...
+
+@test "check that the replicas' names are different" {
+	$DOCKER_EXE service create --name testswarm --replicas $number_of_replicas --publish 8080:80 mcastelino/nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\"" 
+	while [ `docker ps -a | tail -n +2 | egrep "Up [0-9]*"|  wc -l`  -lt  $number_of_replicas ]; do
+		sleep 1
+	done
+	declare -a REPLICAS
+	for i in `seq 0 $((number_of_replicas-1))`; do
+		unset http_proxy
+		set +e
```
jodh-intel: Can you add a comment explaining why `set -e` is disabled for this block of code?

tests/integration/docker/swarm.bats
jodh-intel: Probably a question for @mcastelino :)

chavafg:
qa-failed

tests/integration/docker/swarm.bats
mcastelino: @jodh-intel eventually we should switch to upstream ngnix or I can create an auto built image on docker hub... I think @GabyCT ran into issues with the upstream image. I will look into the root cause for the upstream failure.

Alternately we can just generate the image on the fly. But it will not scale when we run this test on a true cluster. I will work on a local swarm test (which is all we are doing today anyway).

GabyCT:
Hi @jodh-intel 

Thanks for all the feedback. Here are some of the answers to your questions. I modified the number of replicas to 15 now as this is the maximum that my nuc can handle.  In order that the hostname of the replica is returned one of the things that you need to do is unset http_proxy because if you do not do it the curl does not return the hostname. I  moved the URL to the top to make it easier to maintain. I have the set +e and set -e because even if you are making a retry in doing the curl if the first time it fails, bats will put the test as failed immediately. Finally, I am still using Manohar image until we have another option as I have issues running the nginx image from Docker hub. Thanks for everything.

jodh-intel:
Hi @GabyCT - thanks for the info. I knew there would be reasons for all the things I added queries for :smile:

But it's best if the explanations live with the code - could you add comments to `swarm.bats` with the info you've added above. That would make reading the code much more understandable to those that don't know swarm and bats intimately.

Thanks!

GabyCT:
Hi @jodh-intel , I put the explanations with the code :) Thanks

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,75 @@
...
+	cleanDockerPs
+	runtimeDocker
+	$DOCKER_EXE swarm init
+}
+
+@test "check that the replicas' names are different" {
+	# currently using mcastelino/nginx but it will be
+	# modified when nginx image is working
+	$DOCKER_EXE service create --name testswarm --replicas $number_of_replicas --publish 8080:80 mcastelino/nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\"" 2> /dev/null
+	while [ `docker ps -a | tail -n +2 | egrep "Up [0-9]*"|  wc -l`  -lt  $number_of_replicas ]; do
```
jcvenegas: nice wait until all the containers are ready to start to query, I wonder if some of them die will wait this tests for ever?

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,75 @@
...
+		sleep 1
+	done
+	# this will help to obtain the hostname of 
+	# the replicas from the curl
+        unset http_proxy
+        for j in `seq 0 $number_of_attemps`; do
```
jcvenegas: why try number_of_attemps before compare the results?  is it a warm up?

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,75 @@
...
+	                set -e
+                done
+                sleep 5
+        done
+	for i in `seq 0 $((number_of_replicas-1))`; do
+                if [ -z "${REPLICAS[$i]}" ]; then
+                    exit 1
+                fi
+		for j in `seq $((i+1)) $((number_of_replicas-1))`; do
+			if [ "${REPLICAS[$i]}" == "${REPLICAS[$j]}" ]; then
```
jcvenegas: @mcastelino here we compare that we dont have a consecutive response from the same container, does swarm load balancer ensure that will return return with a different container (round robin) ?

mcastelino: @jcvenegas it does not ensure pure RR. It is based on rate of requests and time lag between then. So we need to poll enough times to ensure we see all of them.

tests/integration/docker/swarm.bats
GabyCT: @jcvenegas if they are not up the test will fail

GabyCT:
@jcvenegas  I did all the changes. Thanks

GabyCT:
@grahamwhaley thanks, I did the changes in the patch format style

tests/integration/docker/swarm.bats
```diff
@@ -0,0 +1,75 @@
...
+	done
+	# this will help to obtain the hostname of 
+	# the replicas from the curl
+        unset http_proxy
+        for j in `seq 0 $number_of_attemps`; do
+	        for i in `seq 0 $((number_of_replicas-1))`; do
+		        # this will help that bat does not exit incorrectly 
+			# when curl fails in one of the attemps
+			set +e
+		        REPLICAS[$i]="$(curl $url 2> /dev/null)"
```
GabyCT: @jcvenegas I tried it with run but it did not work out so that is the reason I am using set

gorozco1:
👍

devimc:
lgtm

PR #485, opened 2016-11-30
Restore terminal required for standalone case

amshinde:
We currently set the terminal in raw mode. When we run standalone and pass $(tty) to --console option to start and stop a container, we would need to restore the terminal back.
After restoring the terminal, it does not behave as a raw terminal.

### DIFF
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -342,12 +342,12 @@ read_IO_message(struct cc_shim *shim, uint64_t *seq, ssize_t *stream_len) {
 
 		ret = read(shim->proxy_io_fd, buf+bytes_read, (size_t)want);
 		if (ret == -1) {
-			shim_warning("Error reading from proxy I/O fd: %s\n", strerror(errno));
-			goto err;
+			free(buf);
+			err_exit("Error reading from proxy I/O fd: %s\n", strerror(errno));
 		} else if (ret == 0) {
 			/* EOF received on proxy I/O fd*/
-			shim_warning("EOF received on proxy I/O fd\n");
-			goto err;
+			free(buf);
+			err_exit("EOF received on proxy I/O fd\n");
 		}
 
 		bytes_read += ret;
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -31,6 +31,7 @@
 #include <stdbool.h>
 #include <limits.h>
 #include <fcntl.h>
+#include <termios.h>
 
 #include "utils.h"
 #include "log.h"
@@ -42,6 +43,8 @@
 int signal_pipe_fd[2] = { -1, -1 };
 
 static char *program_name;
+
+struct termios *saved_term_settings;
 
 /*!
  * Add file descriptor to the array of polled descriptors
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

shim/shim.c
```diff
@@ -665,6 +678,11 @@ main(int argc, char **argv)
 		struct termios term_settings;
 
 		tcgetattr(STDIN_FILENO, &term_settings);
+		saved_term_settings = malloc(sizeof(struct termios));
```
jodh-intel: I'd always use `calloc(3)` to avoid potential surprises.

shim/shim.c
```diff
@@ -432,6 +446,7 @@ handle_proxy_output(struct cc_shim *shim)
 		code = *(buf + STREAM_HEADER_SIZE); 	// hyperstart has sent the exit status
 		shim_debug("Exit status for container: %d\n", code);
 		free(buf);
+		restoreTerminal();
```
jodh-intel: It might be safer to also call this when receiving certain signals to ensure the user is returned to a sane terminal environment even if the shim crashes.

amshinde: Oh this is only case where we exit cleanly with the exit code. All others are error conditions and use err_exit.

shim/shim.c
amshinde: Sure.

shim/shim.c
dlespiau: I meant the camelCasing of restoreTerminal() instead of restore_terminal(), or terminal_restore() :)

jodh-intel: Sure. I'm talking about handling signals being delivered to the shim itself (including `SIGSEGV`) - we could make the shim catch these and call `restoreTerminal()`. It might be best to register `restoreTermina()` with `atexit()` too since currently it looks like if the shim fails to read the proxy control line, the shim will exit leaving the terminal in raw mode.

amshinde: @dlespiau: Sure :) Looks like I had been staring at go code long enough :)
@jodh-intel I'll look at adding it at atexit

jodh-intel: Thanks - yes, since there is already a signal handler, the `atexit`-registered function will be called for those signals already handled.

It's worth noting though that those signals that are currently not explicitly handled and which default to terminate will leave the terminal in raw mode. We should probably document those somewhere since someone is bound to try something like...

```
docker kill --signal SIGSEGV $container
```
... and wonder why their terminal is screwed.

`^Jstty sane^J` to the rescue! :smile:

dlespiau: Note: this is only an issue when the console is specified by hand in the standalone mode. Docker allocates a new pty in the normal case and it doesn't matter if we leave that one in raw mode.

```
docker kill --signal SIGSEGV $container
```

Shouldn't leave the terminal in raw mode then.

amshinde:
Pushed changes addressing review comments.

sboeuf:
@chavafg, could you re-run QA testing please ? @amshinde reworked her patches and it would be good to know about the status before we can get this PR merged.

chavafg:
Retest executed, again, functional tests failed (but will be fixed by 431). Also some docker tests got hanged, but seems that this PR is not rebase with @dlespiau patch that fixes the issue. Everything else looks ok.

dlespiau:
lgtm

sboeuf:
lgtm

PR #490, opened 2016-11-30
proxy: Add a way to serve pprof data over HTTP

dlespiau:
The -pprof flag will make the proxy server profiling data overt HTTP.
A nice benefit from having this is that we can always ask for the
backtraces of all goroutines:

  http://localhost:6060/debug/pprof/goroutine?debug=2

This is very handy to debug odd things happening to the proxy and check
we don't have a goroutine blocked somewhere.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"path/filepath"
 	"sync"
@@ -370,11 +372,41 @@ func initLogging() {
 	}
 }
 
+type profiler struct {
+	enabled bool
+	host    string
+	port    uint
+}
+
+func (p *profiler) setup() {
+	if !p.enabled {
+		return
+	}
+
+	addr := fmt.Sprintf("%s:%d", p.host, p.port)
+	url := "http://" + addr + "/debug/pprof"
+	glog.V(1).Info("pprof enabled on " + url)
+
+	go func() {
+		http.ListenAndServe(addr, nil)
+	}()
+}
+
 func main() {
+	var pprof profiler
+
 	initLogging()
+
+	flag.BoolVar(&pprof.enabled, "pprof", false,
+		"enable pprof ")
+	flag.StringVar(&pprof.host, "pprof-host", "localhost",
+		"host the pprof server will be bound to")
+	flag.UintVar(&pprof.port, "pprof-port", 6060,
+		"port the pprof server will be bound to")
 
 	flag.Parse()
 	defer glog.Flush()
 
+	pprof.setup()
 	proxyMain()
 }


chavafg:
qa-failed

dlespiau:
```
not ok 42 start and kill state
```

sameo:
LGTM

proxy/proxy.go
```diff
@@ -350,11 +352,26 @@ func initLogging() {
...
+
+	go func() {
+		http.ListenAndServe("localhost:6060", nil)
+	}()
+}
+
 func main() {
 	initLogging()
 
+	flag.BoolVar(&enablePprof, "pprof", false, "enable pprof (visit http://localhost:6060/debug/pprof)")
```
jodh-intel: Could we use a variable to store the port number rather than using another hard-coded string?

jodh-intel: Also, what about an option to specify the port number?

jcvenegas: and ip to listen as well, would help to not only listen in localhost if it is a remote server where I want to debug

proxy/proxy.go
```diff
@@ -350,11 +352,26 @@ func initLogging() {
...
 
+var enablePprof bool
+
+func setupPprof() {
+	if !enablePprof {
+		return
+	}
+
+	go func() {
+		http.ListenAndServe("localhost:6060", nil)
```
sboeuf: @jodh-intel , what are you suggesting instead ? bigger port number instead ?

sboeuf: sorry, didn't read below comments...

jodh-intel: I'm saying `6060` isn't in `/etc/services`, so this number seems to have been randomly chosen. So, what if another application running on the same machine also tries to use the same port number? If it's not a well-known port number (or even if it is infact), there should be a way to change it.

sboeuf: Yes agreed, and as you and @jcvenegas suggested, it would be great to have the ip and port as options.

sboeuf:
@dlespiau , have you got the time to rework your commit according to the comments, or are you planning to do it ?

dlespiau:
Reworking this isn't high priority at the moment, so we either merge it as it and file issues to track what's remaining to do it leaves this pending, so far, it seems we've chosen the second option.

sboeuf:
Let's leave this PR pending until it is reworked then.

sameo:
@jodh-intel @jcvenegas both the HTTP protocol and the 6060 port are defined by the Golang `pprof` package and can not be redefined afaik.
I agree we should warn the user about 2 things:
- This gives internal data through an insecure channel
- We might not be able to bind on 6060

jodh-intel:
@sameo - I wonder why Google haven't had that port registered to make it "more well-known" (well-known outside golang circles)?:

```
$ getent services 6060 || echo not well-known
not well-known
```

Until it is well-known, they should provide an option to change it...

sameo:
LGTM

jodh-intel:
@dlespiau - re https, yes I tend to agree.

lgtm.

PR #494, opened 2016-11-30
configure: Add configure option for running proxy tests

amshinde:
Certain distros do not support go build with -race. Provide
an option to disable running proxy tests for OBS builds.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -267,10 +267,13 @@ cc_proxy_extra_dist =			\
 	proxy/README.md			\
 	proxy/COPYING
 
+if PROXY_TESTS
+
 CHECK_DEPS += check-proxy
 
 check-proxy:
 	go test -v -race -timeout 2s $(srcdir)/proxy
+endif
 
 libexec_PROGRAMS = cc-shim
 
--- a/configure.ac
+++ b/configure.ac
@@ -89,6 +89,23 @@ AS_IF([test x"$enable_cppcheck" = "xyes"],
     [AC_SUBST([CPPCHECK], [0])])
 AM_CONDITIONAL([CPPCHECK], [test x$enable_cppcheck = x"yes"])
 
+# Runtime proxy tests
+
+AC_ARG_ENABLE(proxy-tests,
+              AS_HELP_STRING([--disable-proxy-tests],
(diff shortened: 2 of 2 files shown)


chavafg:
qa-failed

chavafg:
`make distcheck` failed 

```
/bin/sh: line 1: proxy/cc-proxy.service: No such file or directory
make[2]: *** [Makefile:4723: proxy/cc-proxy.service] Error 1
make[1]: *** [Makefile:1492: all] Error 2
make[2]: Leaving directory '/root/go/src/github.com/01org/cc-oci-runtime/cc-oci-runtime-0.1/_build/sub'
make: *** [Makefile:4469: distcheck] Error 1
```

jodh-intel:
It seems a little extreme to entirely disable the proxy tests if `-race` isn't supported. Surely they can be run without this option?

How about having `configure.ac` detect if `-race` is supported (by running `go build --help 2>&1|grep -- -race`) and simply adding `-race` to the `check-proxy` command in `Makefile.am` via a variable if that option is supported?

dlespiau:
I vote for James' suggestion!

amshinde:
@jodh-intel Sure, makes sense. I'd also like to keep the option to disable proxy tests as well as we support it for all  other tests.

dlespiau:
I'd actually be in favour of trying to reduce the number of configure options we support, they add more complexity and maintenance overhead.

sboeuf:
@dlespiau , same comment that PR#513, maybe we should tag this as do-not-merge, waiting for @amshinde to get back from vacation ?

dlespiau:
A small m4 macro as James suggested to ensure we can compile and execute a small go function and a dummy test with -race seems like the best option to me.

Until then, it's also possible to add a distro patch to remove the "-race" from the Makefile.

Note the problem happens with go 1.6 as distros packaged whatever they at the time.

gorozco1:
the obs packages were configured to not use -race at the moment.

grahamwhaley:
Hi @amshinde , all - this PR has been sat here for some time in WIP - it looks like we worked around the immediate problem in OBS. Do we want to keep this PR open/pending (I have no problem with that in the mid-term), or maybe we can move it all off to an Issue as it now feels more appropriate?

Issue #455, opened 2016-11-24
"kill" command doesn't work with shim/proxy


devimc:
we have different ways to fix this:
* delete command should stop to cc-shim
* cc-proxy should notify to cc-shim when the Vm ends
* cc-shim should monitor to the VM

dlespiau:
So, I don't think delete should stop cc-shim. delete should only be called when cc-shim has already exited

devimc:
I applied this patch #460 and then ```create -> start -> kill -> delete``` a container
cc-shim became crazy printing a lot of messages
```
cc-shim[14691]: read_IO_message:377:EOF received on proxy I/O fd
cc-shim[14691]: read_IO_message:377:EOF received on proxy I/O fd
.....
```

chavafg:
Not sure if this is related to this issue.

I got a cc-shim process that was not stopped after the container was shutdown.
I killed it with `killall cc-shim` and `cc-proxy` was also killed. is this a correct behavior? 

stderror from cc-proxy when it was killed: 
```
+ ./cc-proxy
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x40459a]

goroutine 635 [running]:
panic(0x526b80, 0xc42000c130)
        /usr/lib/golang/src/runtime/panic.go:500 +0x1a1
main.(*vm).writeMessage(0xc4203d0380, 0xc42003cd80, 0xc400000012, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/vm.go:150 +0x15a
main.(*vm).SendMessage(0xc4203d0380, 0xc4200742a0, 0xd, 0xc4200700c0, 0x5d, 0x60, 0x0, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/vm.go:350 +0x136
main.hyperHandler(0xc4200b8090, 0x82, 0x90, 0x513ae0, 0xc4203a9ce0, 0xc4203a9fa0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/proxy.go:198 +0x154
main.(*protocol).handleRequest(0xc420082020, 0xc42003cf20, 0xc4203baab0, 0xc4203a9fa0, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/protocol.go:95 +0xca
[...]

PR #495, opened 2016-11-30
cor: Fix kill command

devimc:
kill command shouldn't send 'killcontainer' command to hyperstart
because cc-shim doesn't receive that signal hence cc-shim never ends.
with this patch kill command sends a kill signal to cc-shim
and cc-shim sends killcontainer to hyperstart.

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -236,13 +236,9 @@ cc_oci_kill (struct cc_oci_config *config,
 		goto error;
 	}
 
-#ifndef UNIT_TESTING
-	if (! cc_proxy_hyper_kill_container(config, signum)) {
-		g_critical("failed to kill container");
-		goto error;
-	}
-#else
-	//FIXME: should we kill to cc-shim?
+	/* send signal to cc-shim because it
+	 * will send killcontainer to hyperstart
+	 */
 	if (kill (state->pid, signum) < 0) {
 		g_critical ("failed to stop container %s "
 				"running with pid %u: %s",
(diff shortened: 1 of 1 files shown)


jcvenegas:
lgtm

chavafg:
qa-failed

src/oci.c
```diff
@@ -236,13 +236,9 @@ cc_oci_kill (struct cc_oci_config *config,
...
-	if (! cc_proxy_hyper_kill_container(config, signum)) {
-		g_critical("failed to kill container");
-		goto error;
-	}
-#else
-	//FIXME: should we kill to cc-shim?
+	/* send signal to cc-shim because it
+	 * will send killcontainer to hyperstart
+	 */
 	if (kill (state->pid, signum) < 0) {
```
jodh-intel: This is fine for most circumstances, but not if `signum == SIGKILL` as the shim won't be able to handle that and will be killed immediately, leaving the container running.

We need a special case so that if `signum == SIGKILL`, the shim is killed and `cc_proxy_hyper_kill_container()` is also called to shutdown the container.

devimc: I think SIGKILL and SIGSTOP cannot be caught by cc-shim

gorozco1: @amshinde any comments?

jodh-intel:
Fixes #455.

sameo:
LGTM

sboeuf:
lgtm

Issue #489, opened 2016-11-30
2.1 - run.bats fails


dlespiau:
This may not be a problem but acutally a fix from 2.0, need investigation but I'm currently looking at another problem, so logging that one:

```
1..1 
not ok 1 Run shell echo 
# (in test file run.bats, line 34) 
# `[ "${output}" == $'Passed\r' ]' 
```

dlespiau:
#498 has been merged!

PR #498, opened 2016-12-01
tests: Make testing for the run.bats output more robust

dlespiau:
I had to include the "\r" in that string for things to work in the 2.0
era. It was a bit odd but I didn't question it too much then, the test
was doing what it was supposed to do: ensure the runtime doesn't output
anything on stdout/err.

2.1 fixes a few things with terminal handling, in particular we don't
have that spurious \r!

### DIFF
--- a/tests/integration/docker/run.bats
+++ b/tests/integration/docker/run.bats
@@ -31,5 +31,5 @@ setup() {
 @test "Run shell echo" {
 	run $DOCKER_EXE run busybox sh -c "echo Passed"
 	[ "${status}" -eq 0 ]
-	[ "${output}" == $'Passed\r' ]
+	[[ "${output}" == 'Passed'* ]]
 }


dlespiau:
Fixes: https://github.com/01org/cc-oci-runtime/issues/489

jodh-intel:
Probably best to specify "for CC 2.1" in the summary rather than "the new order" as the former is clearer.

lgtm.

dlespiau:
PR Updated: tweaked the commit message

jodh-intel:
Nice - thanks.

chavafg:
qa-failed

dlespiau:
But then somehow this fails on CI:

```
# (in test file /root/go/src/github.com/01org/cc-oci-runtime/cc-oci-runtime-0.1/tests/integration/docker/run.bats, line 34)
#   `[ "${output}" = 'Passed' ]' failed
```

I wonder what's going on :/ runc's tests are of the form [[ ${output} = *'Passed'* ]], there may be a reason for that!

jodh-intel:
You can debug by doing something like the following to get a dump on assertion failure:

```
run $DOCKER_EXE run busybox sh -c "echo Passed"
[ "${status}" -eq 0 ]

hex=$(echo "$output"|od -xc || :)
echo "output: '$output' ($hex)" >&2

[ "${output}" = 'Passed' ]
```

dlespiau:
What I don't get is why I see this error locally and in *some* CI logs, but not in the logs of this PR.

jodh-intel:
What's the difference in output you are seeing? Is it simply sometimes you see the `\r` and other times not?

dlespiau:
exactly!

dlespiau:
Locally, I only see no '\r' on the 2.1 branch, so my real question is why this isn't the case in CR (but I do see some of the 2.1 PRs have failed with the same problem there as well).

jodh-intel:
Sounds like the `\r` should be there but isn't always being delivered by the pty driver immediately.

I'm guessing the reason `runc`'s bats tests use `[[` is because that allows you to use bash's regex matching.

You could therefore do the following which should resolve the issue:

```
[[ "$output" =~ Passed\r* ]]
```
(quotes on the right hand side not used as that would stop the regex from matching).

sboeuf:
@dlespiau do you have an update on this one ?

dlespiau:
PR updated with [[ ]] and == 'Passed'*

@chavafg could you rerun the tests on this one? Thanks

sboeuf:
@chavafg , no update on this one from QA side ?

chavafg:
retested and some functional tests still failling, but that should be fixed when merging #431, everything else looks ok.

chavafg:
qa-passed

jodh-intel:
lgtm.

sboeuf:
lgtm

PR #501, opened 2016-12-01
cor: start command waits for cc-shim ends

devimc:
Since cc-oci-runtime must be able to run containers
standalone, start command is in charge of determinate
if shim is running or not also start command must
update state file once shim has finished.

This patch opens a file and locks it after fork and before
exec cc-shim when cc-shim ends this file will be unlocked
allowing to start command catch that event and update the state file,
this is another way to detect when a process ends.

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -33,6 +33,8 @@
 #include <stdio.h>
 #include <sys/types.h>
 #include <pwd.h>
+#include <fcntl.h>
+#include <sys/file.h>
 
 #include <glib.h>
 #include <glib/gstdio.h>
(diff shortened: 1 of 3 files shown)


chavafg:
qa-failed

src/oci.c
```diff
@@ -1070,7 +1050,26 @@ cc_oci_start (struct cc_oci_config *config,
 	              config->state.state_file_path, false);
 
 	if (wait) {
-		g_main_loop_run (data.loop);
+		/* connecting to hypervisor socket */
+		if (data.loop) {
+			g_main_loop_run (data.loop);
```
jcvenegas: I think we only need to use the main loop to wait for the hypervisor socket in "create", here is not  required anymore :) given that the  qemu process is already  running from "create" and is ready (even already talked with hyperstart on create). 
Probably should be enough just to  check a hypervisor socket exist to verify the vm is running

dlespiau:
Could you give more context about why "start" would need to wait for cc-shim to end? I thought having start exit once started was what we wanted to to?

sameo:
I don't think we need to marshal, store, unmarshall the flock file path here. We can just rebuild the path at runtime, since we know it's going to be under `/var/run/cc-oci-runtime/<id>`.
This would simplify the code at least.

jcvenegas:
+1

gorozco1:
👍

PR #502, opened 2016-12-01
cor: remove unused functions

devimc:
since hyperstart is in charge of the VM
the only right way to shutdown the VM is through
hyperstart's API hence qmp shutdown message is not
more needed.

This patch also fix the cppcheck error:
"The function 'cc_oci_vm_shutdown' is never used."

### DIFF
--- a/src/network.c
+++ b/src/network.c
@@ -453,37 +453,6 @@ out:
 }
 
 /*!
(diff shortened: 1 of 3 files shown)


sameo:
LGTM

amshinde:
lgtm

chavafg:
qa-failed

PR #503, opened 2016-12-01
Mmartinm redesign for cc 2.1

mmartinm:
Creating stress test folder and uploading test script 

containers_memory_stress.pl


chavafg:
qa-failed

dlespiau:
This PR has 149 commits instead of 1. You need to submit this PR against the 2.1 branch if you want the commit to be applied there, or base your work on the master branch if you to  commit this to master. I'd recommend you aim for master for this one.

Then, well, perl. I'm really not a fan of introducing a new language in the project and perl would definitely not be the first choice, it has lost the scripting war many moons ago.

It seems this script could just use shell scripting. That said shell scripting can be a bit arcane as well (not as much as perl mind you). Re-writing this script in go could be an interesting little project to see how easy it is to use go as a replacement for something close to shell scripting.

In short, I'd rather we not introduce yet another language in the project, I think the line has to be drawn somewhere before we end up with perl, python, ruby (why not!), php (the CI web site!) scripts creeping everywhere.

grahamwhaley:
Right, I agree with @dlespiau on many points.
To clarify, it looks like you should only have your single commit there, and all the other commits look like they have come from merges/tracking of the main trees in your repo - so, you need to rebase your repo against whichever tree you are submitting against.
Once you've done that, can you also rewrite your commit message to conform with our contributing patch format style: https://github.com/01org/cc-oci-runtime/blob/master/CONTRIBUTING.md#patch-format (I think you at least need to fix the subject line and add a signed off by from a quick look).

And I agree about usage of perl - not only as perl is one of my least favourite languages :-), but as Damien says, we should keep the number of languages used in this project to a minimum, and this would be a first for perl here.

Other than that, a quick glance over the script looks like it is well laid out etc. One query - do the workdir data directories need to be deleted during the cleanup?

gorozco1:
:-O  👎

mmartinm:
thanks for your comments, this script will be changed to another language (go) and I will do the pull request to the correct branch

chavafg:
qa-failed

dlespiau:
Ok, time for some cleanup, closing that one. Please do re-open with the new stress script in go and against master now.

PR #507, opened 2016-12-02
cor: Fix read out-of-band data

devimc:
Sometimes when cc-proxy sends only out-of-band
data cor tries to convert it to big endian
returning a negative value that is used to
append the iov buffer to msg_received
causing a crash (segment default)
    
To fix that crash this patch validates that
messages read from cc-proxy are not out-of-band data.

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -259,10 +259,6 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
 		return false;
 	}
 
-	memcpy(buf_len, iov_buffer, sizeof(p_msg.length));
-	p_msg.length = cc_oci_get_big_endian_32(buf_len);
-	g_debug("msg length: %d", p_msg.length);
-
 	if (proxy_data->oob_fd) {
 		/* check if oob data was received */
 		cmsg = CMSG_FIRSTHDR(&msg);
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

src/proxy.c
```diff
@@ -277,12 +273,26 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
-	bytes_to_copy = MIN((size_t)bytes_read-data_offset, (size_t)p_msg.length);
-	g_string_append_len(proxy_data->msg_received,
-		iov_buffer+data_offset, (ssize_t)bytes_to_copy);
-
-	if (bytes_read >= p_msg.length+data_offset) {
+	/* cc-proxy sends a fd as out-of-band data
+	 * followed by an 'F'.
+	 * *DO NOT* copy these characters to msg_received
+	 */
+	if (bytes_read > 1 || iov_buffer[0] != 'F') {
```
jcvenegas: so, when bytes_read is > 1  iov_buffer[0] will never be 'F'  (well will be too big ), right? 
lgtm

src/proxy.c
```diff
@@ -277,12 +273,26 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
+		g_string_append_len (proxy_data->msg_received,
+			iov_buffer+data_offset, (ssize_t)bytes_to_copy);
+
+		/* all message was read */
+		if (bytes_read >= p_msg.length+data_offset) {
+			goto out;
+		}
+	} else {
+		/* only out-of-band data was received */
 		goto out;
```
jodh-intel: Is this `goto` safe? It jumps down to a `g_debug()` call that displays the received message... but there isn't one in the OOB message scenario right? Do we need a new `out2` label maybe?

sboeuf:
lgtm

amshinde:
lgtm

PR #513, opened 2016-12-05
tty:Do not set current tty as controlling tty for shim in standalone mode

amshinde:
Added detailed explanation in the comments. This is to prevent the shell
from losing its controlling tty. I thought of overcoming this problem by
having the runtime gain control of the tty once the shim has ended.
But if the runtime is sent a SIGKILL in standalone, the shell would
never get its tty back.

In any case, I feel we should not be causing any ill effects on the
user's running shell. We should have the runtime create a slave pts in
standalone mode and pass that to the shim in 3.0(runc creates a tty as well
and prevents the current shell from being used)

A limitation introduced with this change is that the shim does not
get the SIGWINCH signals and these are not propogated to hyperstart.
I plan to add this limitation to the wiki once this is approved.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -227,8 +227,21 @@ cc_oci_setup_shim (struct cc_oci_config *config,
 	/* become session leader */
 	setsid ();
 
-	// In the console case, the terminal needs to be dup'ed to stdio
-	if (config->oci.process.terminal) {
+	/* In the console case, the terminal needs to be dup'ed to stdio
+	 * and the tty provided by docker is set as the controlling tty 
+	 * for the shim.
+	 *
+	 * For running standalone, we currently pass the current $(tty).
+	 * In that case, stdio is set up correctly. Do not set the tty as
+	 * the controlling tty for shim, since this makes the shell run with no
+	 * controlling tty resulting in the following error:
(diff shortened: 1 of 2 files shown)


chavafg:
qa-failed

sboeuf:
@amshinde did you check why the QA was failing ?

chavafg:
These were the functional tests that are currently failing in 2.1 branch ( @devimc is working on these failures). Docker tests were hanged, but this is already fixed with latest commit on the branch, so maybe a rebase will make docker tests pass.

dlespiau:
So, I'm not a big fan of losing the SIGWINCH handling for a corner case where we reuse the current shell. How important it really is that we can reuse the current shell? if it's only for the tests and in "standalone" mode, we could do what runc does (apparently), creating a new pty.

sboeuf:
@dlespiau , maybe you should tag this PR as do-not-merge until @amshinde is back, to continue this conversation ?

dlespiau:
@sboeuf yes, we should.

dlespiau:
As discussed, closing that PR with issue #571 to track this. We may want to do what runc used to, have a special meaning for --console /dev/pts/ptmx

Issue #54, opened 2016-07-27
Exit codes from non-interactive runs are not handled.


amshinde:
With runc:

```
$ sudo docker-upstream run --runtime runc  debian /usr/bin/perl -e 'exit 55'
$ echo $?
55
```

with cc-oci-runtime:

```
$ sudo docker-upstream run  debian  /usr/bin/perl -e 'exit 55'
$ echo $?
0
```

dlespiau:
So, after a first look at this, I think the best approach is to wait until the exec over qmp/qagent is done (issue #18) and refactor the run functionality using the same mechanism. Putting back to ready until then. Do shout if you don't agree.

dlespiau:
If we want that feature before #18 it should be easy enough (famous last words) to write a small temporary patch that solves the issue:

- Create a Virt I/O serial device bound to a unix socket for the guest to communicate the exit value back to the host. Open that socket and put in the set of fds to listen to.
- modify the generated .container exec to echo $? > /dev/virtio-ports/$name_of_the_device after the command executed (mechanism used by the 'run' command)
- make sure we read back the exit value when qemu exits and use it for the runtime own exit value.

chavafg:
I am adding some tests for this in PR #523 
But when running under a terminal, the return code is not handled correctly.

Without -t 
```
cc-tester@clear-cc-ci ~ $ sudo docker run -i ubuntu /usr/bin/perl -e 'exit 55'
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
cc-tester@clear-cc-ci ~ $ echo $?
55
```


With -t 
```         
cc-tester@clear-cc-ci ~ $ sudo docker run -ti ubuntu /usr/bin/perl -e 'exit 55'
cc-tester@clear-cc-ci ~ $ echo $?
0
```

chavafg:
Actually, the tests are passing, there is an issue with running functional stand-alone tests first (Issue #527) and then running this test.

devimc:
seems like this bus was fixed
@chavafg  @GabyCT  can I close it ?

GabyCT:
@devimc it has been fixed, you can close it

dlespiau:
@GabyCT Don't hesitate to close the issue yourself if you confirm/think it can be closed.

PR #523, opened 2016-12-08
Tests to verify exit codes and terminal setup

chavafg:
This PR contains two commits that add integration tests to verify exit codes and terminal setup:

integration tests: Add tests to verify that 
    1. a pseudo tty is properly setup when it is requested
    2. TERM environment variable is set when a tty is requested
    3. TERM environment variable is not set when a tty is not requested

integration tests: Add test to verify exit code from a container
Add tests to verify that a container returns the correct exit code of the executed process.

### DIFF
--- a/tests/integration/docker/exit-code.bats
+++ b/tests/integration/docker/exit-code.bats
@@ -1,0 +1,41 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


chavafg:
qa-failed

dlespiau:
Considering your comment in #54, it means those tests are all passing? I'd suggest we skip the ones that are known to fail, we shouldn't add failing tests before they are fixed.

sboeuf:
@chavafg at least you should tag this PR as Work In Progress until you have those tests fixed. Or you remove them as @dlespiau suggested.

chavafg:
done :)

chavafg:
After research, found that these tests are working, 
```
1..2 
ok 1 Exit Code from container process when running non-interactive 
ok 2 Exit Code from container process when running interactive 
	
1..3 
ok 1 TERM env variable is set when allocating a tty 
ok 2 TERM env variable is not set when not allocating a tty 
ok 3 Check that pseudo tty is setup properly when allocating a tty
```
but there is an issue which breaks the terminal #527 

can we merge this tests?

tests/integration/docker/exit-code.bats
```diff
@@ -0,0 +1,41 @@
...
+exit_status=55
+
+setup() {
+	source $SRC/test-common.bash
+	clean_docker_ps
+	runtime_docker
+}
+
+@test "Exit Code from container process when running non-interactive" {
+	run $DOCKER_EXE run ubuntu /usr/bin/perl -e "exit $exit_status"
```
devimc: I think that you could use the perl container

sameo:
LGTM

devimc:
lgtm

PR #526, opened 2016-12-12
error handling

jcvenegas:
This PR adds error handling to some functions:

create: Before open console check if is not NULL
log: Add error handling in log setup for hypervisor
process: close duplicated shim fds explicitly
process: handle errors errors in cc_oci_setup_child

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -598,7 +598,7 @@ out:
  * \param shim_socket_fd Writable file descriptor caller should use to
  *   send the proxy IO fd to child before \ref CC_OCI_SHIM is launched.
  *
- * \return \c shim-pid  on success, else \c -1.
+ * \return \c true on success, else \c false.
  */
 private gboolean
 cc_shim_launch (struct cc_oci_config *config,
@@ -607,6 +607,7 @@ cc_shim_launch (struct cc_oci_config *config,
 		int *shim_socket_fd,
 		gboolean initial_workload)
 {
+	gboolean  ret = false;
 	GPid      pid = -1;
 	int       child_err_pipe[2] = {-1, -1};
 	int       shim_args_pipe[2] = {-1, -1};
@@ -818,6 +819,7 @@ child_failed:
 	*shim_args_fd = shim_args_pipe[1];
 	*shim_socket_fd = shim_socket[1];
 
+	ret = true;
 
 out:
 	close (child_err_pipe[1]);
@@ -825,7 +827,7 @@ out:
 	close (shim_socket[0]);
 	close (shim_flock_fd);
 
-	return pid;
+	return ret;
 }
 
 /*!
--- a/src/process.c
+++ b/src/process.c
@@ -724,31 +724,20 @@ cc_shim_launch (struct cc_oci_config *config,
 		 * Since these need to be assigned to the terminal fd, make sure the
 		 * proxy fds are assigned >= 3
 		 */
-		while(proxy_socket_fd < 3) {
-			int fd = dup(proxy_socket_fd);
-			if (fd < 0) {
-				g_critical("dup failed: %s", strerror(errno));
-				goto child_failed;
-			}
-			proxy_socket_fd = fd;
+
+		if (! dup_over_stdio(&proxy_socket_fd)) {
+			g_critical("failed to dup proxy_socket_fd");
+			goto child_failed;
 		}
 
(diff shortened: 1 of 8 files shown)


chavafg:
qa-failed

tests/util_test.c
```diff
@@ -517,6 +518,30 @@ START_TEST(test_cc_oci_enable_networking) {
...
+	/* fd pointer is NULL */
+	ck_assert(! dup_over_stdio(NULL));
+
+	/* fd number is -1 */
+	fd = -1;
+	ck_assert(! dup_over_stdio(&fd));
+
+
+	/* dup not used but open STDIN fd */
+	fd = STDIN_FILENO ;
```
jodh-intel: This test is going to close whatever `STDIN_FILENO` is attached to. It should probably do something like the following to minimise surprises in the future if the test gets expanded [*]:

```
/* save */
int saved_stdin = dup(STDIN_FILENO).

<< run tests >>

/* restore */
dup2(STDIN_FILENO, saved_stdin);
```

(although of course if an assertion fires that won't help ;)

---

[*] - It should be ok to close stdin in a single `START_TEST/END_TEST` block as they are run in sub-processes.

jcvenegas: Good catch! , fixed

jcvenegas:
@jodh-intel thanks for the feedback, now is  fixed, does it look better ? @sameo pullaprove complains about a missing "Signed-off-by" but   I don see anything  missing

jodh-intel:
lgtm.

btw - that pullapprove failure is caused by merge commits that don't have sign-offs (0ae117c, 8eaaf07, 2e70364).

jodh-intel:
(we might need to disable pullapprove to allow this PR through as a result?)

grahamwhaley:
@jodh-intel how did we get merge commits in here - should we not be avoiding merge commits that end up in PRs using rebases?

jcvenegas:
@jodh-intel yes, seems that will need to disable pullaprove, just wonder why other PR does not complain, I'll try to re-rebase master

jodh-intel:
@grahamwhaley - Good question. It looks we might want to ask @dlespiau and @sboeuf. It seems to be easily done, so we need to find a way to disable them going forward to keep the pullapprove gods happy.

src/process.c
```diff
@@ -149,12 +150,19 @@ cc_oci_setup_child (struct cc_oci_config *config)
 
 	/* Do not close fds when VM runs in detached mode*/
 	if (! config->detached_mode) {
-		cc_oci_close_fds (NULL);
+		if (! cc_oci_close_fds (NULL)) {
+			goto out;
```
devimc: ```return false``` would be more clear

src/process.c
```diff
@@ -149,12 +150,19 @@ cc_oci_setup_child (struct cc_oci_config *config)
...
+		}
 	}
 
-	cc_oci_setup_hypervisor_logs(config);
+	if (! cc_oci_setup_hypervisor_logs(config)) {
+		goto out;
```
devimc: same here

src/process.c
```diff
@@ -149,12 +150,19 @@ cc_oci_setup_child (struct cc_oci_config *config)
...
+	}
 
-	return true;
+	ret = true;
+
+out:
+	return ret;
```
devimc: and here just ```return true```

src/util.c
```diff
@@ -606,6 +606,86 @@ cc_oci_handle_signals (void)
...
+		goto close_tmp_fds;
+	} else  {
+		/* success */
+		g_debug("fd moved from %d to  %d", *fdp, dup_fd);
+		*fdp = dup_fd;
+		dup_fd = -1;
+		ret = true;
+	}
+
+close_tmp_fds:
```
devimc: seems like this label is not necessary

src/util.c
```diff
@@ -606,6 +606,86 @@ cc_oci_handle_signals (void)
...
+	for (int i = 0 ; i < CC_OCI_ARRAY_SIZE(tmp_fds) ; i++) {
+		/* Dont close the original fd on error */
+		if( ret == false && tmp_fds[i] == *fdp)
```
devimc: ``` ! ret ``` looks better

sameo:
LGTM

Issue #508, opened 2016-12-02
2.1 - docker: Error response from daemon: shim error: open pid: no such file or directory.


dlespiau:
When running `make docker-tests`, I see some tests failing with this error:

```
2.1 - docker: Error response from daemon: shim error: open pid: no such file or directory.
```

It happens "from time to time" but is fairly reproducible: once every few runs of `make docker-tests`. It smells like our writing of the pid file could be racy, somehow? or we simply fail at writing it.

jcvenegas:
could be related with PR #507 ?

sameo:
I can reproduce that one as well. Some logs:

## Proxy
```
I1205 15:29:56.207216   24760 proxy.go:63] [client #17] client connected
I1205 15:29:56.383403   24760 proxy.go:73] [client #17] hello(containerId=6985974f8db65851ecb4f0dd7e6382325f87794f5503dc4add43433e74eff854,ctlSerial=/var/run/cc-oci-runtime/6985974f8db65851ecb4f0dd7e6382325f87794f5503dc4add43433e74eff854/ga-ctl.sock,ioSerial=/var/run/cc-oci-runtime/6985974f8db65851ecb4f0dd7e6382325f87794f5503dc4add43433e74eff854/ga-tty.sock,console=/var/run/cc-oci-runtime/6985974f8db65851ecb4f0dd7e6382325f87794f5503dc4add43433e74eff854/console.sock)
I1205 15:29:56.546283   24760 vm.go:110] [vm 6985974f hyperstart] open hyper channel /dev/vport1p1
I1205 15:29:56.546499   24760 vm.go:110] [vm 6985974f hyperstart] send ready message
I1205 15:29:56.546561   24760 vm.go:110] [vm 6985974f hyperstart] hyper send type 8, len 0
I1205 15:29:56.546744   24760 vm.go:110] [vm 6985974f hyperstart] open hyper channel /dev/vport1p2
I1205 15:29:56.546869   24760 vm.go:110] [vm 6985974f hyperstart] hyper_init_event hyper channel event 0x41e1b0, ops 0x41e030, fd 3
I1205 15:29:56.547028   24760 vm.go:110] [vm 6985974f hyperstart] hyper_add_event add event fd 3, 0x41e030
I1205 15:29:56.547185   24760 vm.go:110] [vm 6985974f hyperstart] hyper_init_event hyper ttyfd event 0x41e178, ops 0x41e010, fd 4
[...]

Issue #522, opened 2016-12-08
stale cc-shim - "start" fails leaving cc-shim behind


aduenasd:
@The test consist on creating 10 containers in a loop and then destroying those containers, this was done 1000 times.
After tests finishes several instances of cc-shim are still alive.
I'm running something like this:

```
$ docker run -v $workdir/$_:/tdir -tid clearlinux bash -c 'while true; do echo "$date" > /tdir/out.log; sleep 1; done'
```
and then killing the container like this:
```
$ docker rm -f $(docker ps -aq)
```

After the test finishes:
.......
Running 10 containers...
        Running 998 iteration
Running 10 containers...
        Running 999 iteration
Running 10 containers...
        Running 1000 iteration
Running 10 containers...
```
root@clr-skylake ~/MeMLeakTest # pidof qemu-system-x86_64
root@clr-skylake ~/MeMLeakTest # docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
root@clr-skylake ~/MeMLeakTest # cc-oci-runtime list
ID                                                               PID   STATUS   BUNDLE                                    CREATED
caadd10e42ffb72ea0836f50469e606952e95999b437a835130c9b2830e99340 16733 stopped  /run/docker/libcontainerd/caadd10e42ffb72ea0836f50469e606952e95999b437a835130c9b2830e99340 2016-12-07T15:18:10.651089Z
3694a20009b11e5099fd49ceb6902271d567600761ab8a51bf2bbea66ae10405 10329 stopped  /run/docker/libcontainerd/3694a20009b11e5099fd49ceb6902271d567600761ab8a51bf2bbea66ae10405 2016-12-07T09:45:23.910624Z
root@clr-skylake ~/MeMLeakTest # pidof cc-shim
32293 30371 25307 24966 23762 18046 15217 11690 9324 5755 5137 4000 2571 1527
root@clr-skylake ~/MeMLeakTest # pidof docker-containerd-shim
32284 30362 25298 24957 23753 18036 15208 11681 9315 5746 5128 3991 2561 1518
root@clr-skylake ~/MeMLeakTest #
```

I get occasionally receive this type of messages: 
```
[...]

PR #537, opened 2016-12-16
cor: proxy: Fix read from proxy issues

sboeuf:
This patch provide a better protocol handling than the way it was done.
It expects to read buffer length to then determine if more reading is
necessary. At last, it retrieves the file descriptor passed as an out
of band data.

However, there is still an unconsistent behavior related to the OOB data
received. Sometimes when recvmsg() is called to fast, it cannot receive
OOB data containing the fd, ending in a MSG_CTRUNC error. This issue is
only related to OOB data because this error means that nothing is sent
as OOB data. In the case of a regular buffer passed on the band, even if
it is truncated, you can still retrieve the other part by issuing another
recvmsg() call.
That's why we need to make sure to wait a bit before we can try to read
something on the socket, avoiding OOB corruption.

This PR fixes issue #508.

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+#include <stdlib.h>
 #include <string.h>
 #include <errno.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

sboeuf:
@chavafg could you please re-trigger a test, it should pass now that I fixed check issues.

src/proxy.c
```diff
@@ -238,113 +239,131 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
-			data = CMSG_DATA(cmsg);
-			if (data) {
-				*proxy_data->oob_fd = *((int*) data);
-				g_message("oob fd read from proxy socket %d",
-					*proxy_data->oob_fd);
-				fd_read = true;
-			}
+		bytes_read = recvmsg(proxy_data->socket_fd, &msg, 0);
+		if (bytes_read < 1) {
+			usleep(1e3);
```
devimc: is this sleep needed only to get oob data? if this is the case you should use it only when oob is expected
```if (bytes_read < 1 && proxy_data->oob_fd) { ```

File: src/proxy.c [hunk 2]
```diff
@@ -238,113 +239,131 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
+	for (timeout = 0; timeout < 100; timeout++) {
+		if (tot_b_rd >= tot_b_expected) {
 			break;
 		}
-		if (! proxy_data->oob_fd) {
-			/* this message has not out-of-band data */
+
+		bytes_read = recvmsg(proxy_data->socket_fd, &msg, 0);
+		if (bytes_read < 1) {
+			usleep(1e3);
```
devimc: waiting for oob data?

src/proxy.c
```diff
@@ -238,113 +239,131 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
 
-		/* check if oob data was received */
+		tmp_ptr = realloc(frame, tot_b_rd + (size_t)bytes_read);
+		if (!tmp_ptr) {
+			g_critical("failed to realloc memory");
+			goto out;
+		}
+		frame = tmp_ptr;
+		memcpy(frame + tot_b_rd, iov_buffer, (size_t)bytes_read);
+		tot_b_rd += (size_t)bytes_read;
```
devimc: could you use a ```GString``` ?

src/proxy.c
```diff
@@ -238,113 +239,131 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
-	if (bytes_read == -1){
-		g_critical("failed to read proxy message %s", strerror(errno));
-		return false;
+	/* 
+	 * This sleep is very important to make sure we don't
+	 * get MSG_CTRUNC message from msghdr. There is no way
+	 * to recover from a MSG_CTRUNC the control message lost.
+	 */
+	if (proxy_data->oob_fd) {
+		usleep(1e3);
```
jodh-intel: Sleeps trigger very loud warning bells to me. A few questions:

- Is there any way to avoid using a fragile sleep? Surely there is some primitive we can wait for rather than actually sleeping?
- How was this value chosen?
- Is that value guaranteed to work under all conditions on all systems?

chavafg:
CI still getting these errors sometimes: 
```
	
1..6 
not ok 1 modifying a container with exec 
# (in test file exec.bats, line 35) 
# `$DOCKER_EXE exec -d containertest bash -c "echo 'hello world' > file"' failed 
# 8b4812b6d72a266379bd76a872a9a15e433957883254a87579c1381323a15766 
# 8b4812b6d72a ubuntu "bash -c 'sleep 50'" 1 seconds ago Up Less than a second containertest 
# Error response from daemon: rpc error: code = 2 desc = containerd: container not started 
not ok 2 exec a container with privileges 
# (in test file exec.bats, line 41) 
# `$DOCKER_EXE exec -i --privileged containertest bash -c "mount -t tmpfs none /mnt"' failed with status 126 
# 8b4812b6d72a 
# 8b4812b6d72a 
# c2264a231210c02e1419d951a2164014b65f4356020f4813b864bdd755b7482d 
# rpc error: code = 2 desc = containerd: container not started  
```

src/proxy.c
sboeuf: Nope this sleep is to avoid CPU overload by looping too fast, but I can remove it.

src/proxy.c
sboeuf: No, I am not waiting for OOB data specifically, just a sleep to avoid CPU overload.

src/proxy.c
sboeuf: Yep

src/proxy.c
sboeuf: Well, if you read my commit message and the comment on top of that usleep(), it is because of MSG_CTRUNC is received. If you have a good understanding about why I receive this MSG_CTRUNC sometimes without the usleep(), let me know, because I really don't get what is the reason...
I 100% agree it is fragile, and I would expect a better solution if I was able to understand the real root cause.
About the value, I tried with 1s, 100ms, 1ms, and it was still working. I didn't have time to try 100us, 10us and 1us yet. But the point is that without understanding the root cause, I don't know if it will work on all systems.
I know on mine it is solving this issue and issue #522 too.

sboeuf:
@jodh-intel , @devimc As a global comment, I suggest we will get rid of OOB data as it seems to be a very fragile thing and we could do very simple by sending the fd at the end of the frame instead, through In Band.
I will work on a solution removing this OOB support because it would be much easier and robust.

sameo:
@sboeuf @jodh-intel @devimc Please have a look at PR #538. It is simpler and does fix 508 for me.

sboeuf:
PR #538 fixes what this PR was fixing, but in a better way :). Let's close this one.

Issue #535, opened 2016-12-14
"docker exec $container ps" doesn't work without -it


dlespiau:
I start a container with:

```
docker run busybox sh -c "while :; do echo .; sleep 5; done"
```

Then, this works just fine:

```
docker exec -it $container ps
```

But the following doesn't:

```
$ docker exec loving_hugle ps
rpc error: code = 2 desc = containerd: container not started
```

sboeuf:
@dlespiau , @jcvenegas have you been able to reproduce it ? It is working fine on my laptop... No need to add -it, I get the result of the ps without any error.

jcvenegas:
@dlespiau  @sboeuf , same here I have tried this case but seems that works well.

But tryting to see if was a radom bug I did
```
while sudo docker exec $(sudo docker ps -ql) ps;do echo ok; done
```
and after a while and exec workload  hang as the container

sboeuf:
@jcvenegas if it hangs after a while, this means this could be something fixed by @sameo PR #538. But you confirm that it does work for you.
I mean if you look at @dlespiau issue description, it is clear that he could not even make this work once.

jcvenegas:
@sboeuf yes it is working, I was not able to reproduce @dlespiau  issue

sboeuf:
Let's close this one then :)

jcvenegas:
sure, lets re-open it if someone else see this behavior.

jcvenegas:
reopen the issue was seen in #538

devimc:
seems like this issue was fixed in #538

PR #538, opened 2016-12-19
proxy: Simplify fd passing and the message read loop

### LINKED ISSUE #508: 2.1 - docker: Error response from daemon: shim error: open pid: no such file or directory.

sameo:
When passing the OOB fd data, we pass the dummy data ('F') and
that simplifies the message reading loop:

- When receiving a message, we no longer need to read the OOB data
  at any point in time, but only at first.

- We only need to read actual data in the read loop, and there we
  can handle EAGAIN more robustly: We retry as long as we don't
  get the whole message length.

Fixes #508

### DIFF
--- a/proxy/api/client.go
+++ b/proxy/api/client.go
@@ -17,6 +17,7 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net"
 	"os"
 )
@@ -62,6 +63,39 @@ func (client *Client) sendPayload(id string, payload interface{}) (*Response, er
 	}
 
 	return &resp, nil
+}
+
+// sendPayloadGetFd will send a command payload and get a response back
+// but also an out of band file descriptor.
+func (client *Client) sendPayloadGetFd(id string, payload interface{}) (*Response, *os.File, error) {
+	var err error
+
+	req := Request{}
+	req.ID = id
+	if payload != nil {
+		if req.Data, err = json.Marshal(payload); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	if err := WriteMessage(client.conn, &req); err != nil {
+		return nil, nil, err
+	}
+
+	// I/O fd
+	newFd, err := ReadFd(client.conn)
+	if err != nil {
+		return nil, nil, fmt.Errorf("sendPayloadGetFd: couldn't read fd for request %s", id)
+	}
+
+	ioFile := os.NewFile(uintptr(newFd), "")
+
+	resp := Response{}
+	if err := ReadMessage(client.conn, &resp); err != nil {
+		return nil, nil, err
+	}
+
+	return &resp, ioFile, nil
 }
 
 func errorFromResponse(resp *Response) error {
@@ -114,7 +148,7 @@ func (client *Client) AllocateIo(nStreams int) (ioBase uint64, ioFile *os.File, 
 		NStreams: nStreams,
 	}
 
-	resp, err := client.sendPayload("allocateIO", &allocate)
+	resp, ioFile, err := client.sendPayloadGetFd("allocateIO", &allocate)
 	if err != nil {
 		return
 	}
@@ -130,14 +164,6 @@ func (client *Client) AllocateIo(nStreams int) (ioBase uint64, ioFile *os.File, 
 	}
 
 	ioBase = (uint64)(val.(float64))
-
-	// I/O fd
-	newFd, err := ReadFd(client.conn)
-	if err != nil {
-		return 0, nil, errors.New("allocateio: couldn't read fd")
-	}
-
-	ioFile = os.NewFile(uintptr(newFd), "")
 
 	return
 }
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -47,6 +47,7 @@ struct watcher_proxy_data
 	 * otherwise it WILL wait
 	 */
 	int         *oob_fd;
+	gboolean    oob_fd_found;
 };
 
 /** Format of a proxy message */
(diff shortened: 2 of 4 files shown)


src/proxy.c
```diff
@@ -257,14 +256,15 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
 
 	/* read length */
 	bytes_read = recvmsg(proxy_data->socket_fd, &msg, 0);
+	total_bytes_read += bytes_read;
```
jodh-intel: This will make `total_bytes_read` negative if `recvmsg(2)` fails.

chavafg:
qa-failed

src/proxy.c
sameo: Yep, and I need to fix a proxy unit test as well.

sameo:
@chavafg Fixed the unit tests failures, could you please re-run the tests ?

GabyCT:
@sameo I am re-running again the tests :)

sameo:
thanks @GabyCT !

proxy/api/client.go
```diff
@@ -64,6 +64,37 @@ func (client *Client) sendPayload(id string, payload interface{}) (*Response, er
...
+	}
+
+	if err := WriteMessage(client.conn, &req); err != nil {
+		return nil, nil, err
+	}
+
+	// I/O fd
+	newFd, err := ReadFd(client.conn)
+	if err != nil {
+		return nil, nil, errors.New("sendPayloadGetFd: couldn't read fd")
```
jodh-intel: I wonder if it's worth adding the `id` to this error message?

src/proxy.c
```diff
@@ -277,14 +279,14 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
 					*proxy_data->oob_fd);
-				fd_read = true;
 			}
 		}
 	}
 
 	/* only out-of-band data was received */
 	if (bytes_read < sizeof (buf_len) || iov_buffer[0] == 'F' ) {
-		goto out1;
+		proxy_data->oob_fd_found = true;
```
devimc: I think that you should move this to line 280

src/proxy.c
```diff
@@ -264,11 +264,13 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
 	if (bytes_read == 0) {
-		g_critical("failed to read lenght");
+		g_critical("failed to read length");
 		return false;
 	}
 
-	if (proxy_data->oob_fd) {
+	total_bytes_read += bytes_read;
+
+	if (proxy_data->oob_fd && !proxy_data->oob_fd_found) {
```
devimc: ```!proxy_data->oob_fd_found``` is not needed because at this point it always will be true

src/proxy.c
```diff
@@ -47,6 +47,7 @@ struct watcher_proxy_data
 	 * otherwise it WILL wait
 	 */
 	int         *oob_fd;
+	gboolean    oob_fd_found;
```
devimc: you could remove this field and create a local variable in cc_proxy_msg_read since this variable is only used by this function

src/proxy.c
sameo: It will be false in the first round.

src/proxy.c
sameo: Yes, but I need to carry it through 2 `cc_proxy_msg_read()` calls. Once we find the OOB data, we return from this routine and get called again when the actual data comes in. On the second run we want to skip the OOB check.

src/proxy.c
devimc: ```oob_fd_found``` will change to true when **only** out-of-band data is read
I think that this is not equivalent , ```oob_fd_found``` have to change when the fd is copied

src/proxy.c
jodh-intel: Sounds like we might benefit from a `g_debug()` call or two?

sameo:
@GabyCT How did the tests go ?

src/proxy.c
sameo: We do 2 reads for the 2 separate messages that the proxy send us on its blocking socket. So at first we read the OOB ancillary data and the fd, and then get the 2nd message for reading the actual message. I changed the code slightly to make that clearer, with even code comments :)

src/proxy.h
```diff
@@ -28,6 +28,14 @@
 /* allocate 2 streams, stdio and stderr */
 #define IO_STREAMS_NUMBER 2
 
+/*
+ * 4 bytes for the message length.
+ * 4 bytes for the message flags.
+ */
+#define HEADER_MESSAGE_LENGTH 4
+#define HEADER_MESSAGE_FLAGS  4
+#define MESSAGE_HEADER_LENGTH 8
```
jodh-intel: `#define MESSAGE_HEADER_LENGTH (HEADER_MESSAGE_LENGTH+HEADER_MESSAGE_FLAGS)` ?

proxy/api/client.go
```diff
@@ -64,6 +65,37 @@ func (client *Client) sendPayload(id string, payload interface{}) (*Response, er
 	return &resp, nil
 }
 
+func (client *Client) sendPayloadGetFd(id string, payload interface{}) (*Response, *os.File, error) {
```
jodh-intel: I think it would be worth a comment on this function since although the name is fairly descriptive, there are 3 operations going on here (send payload, read fd, read message), but the name suggests 2.

src/proxy.c
```diff
@@ -264,88 +280,81 @@ cc_proxy_msg_read (GIOChannel *source, GIOCondition condition,
...
+			g_critical("could not read the control message");
+			goto out;
 		}
-	}
 
-	/* only out-of-band data was received */
-	if (bytes_read < sizeof (buf_len) || iov_buffer[0] == 'F' ) {
-		goto out1;
+		data = CMSG_DATA(cmsg);
+		if (! (data && iov_buffer[0] == 'F')) {
```
jodh-intel: This "magic `F`" is important. But it's also mentioned explicitly in various places (4 times). From a maintenance perspective, I'd be inclined to create a `#define` for it and change all hard-coded references to use the define (including all the references in comments). This would bring `proxy.c` sort of in line with `proxy.go` which defines `fileFlag` (although `proxy.go` could do a bit better in this respect too as there are various mentions of "`F`" scattered about in comments).

jodh-intel:
lgtm

sameo:
@GabyCT Could we please run our QA tests on this PR ?

sameo:
@devimc Please give it another review round.

GabyCT:
@sameo I checked the test results and I saw that the failures that we have in docker integration tests and make distcheck are related with the issue https://github.com/01org/cc-oci-runtime/issues/535 as all the errors happen using exec (Error response from daemon: rpc error: code = 2 desc = containerd: container not started docker integration)

sboeuf:
lgtm

PR #541, opened 2016-12-20
tests: Do not play with stdin in unit tests

### LINKED ISSUE #527: Docker tests (that involve tty) fail after running functional tests on redesign-for-cc-2.1
We have some tests that are failing randomly on the redesign-for-cc-2.1 branch and we do not have the root cause yet. 

```
not ok 4 stdout forwarded using exec
# (in test file exec.bats, line 55)
#   `$DOCKER_EXE exec -ti containertest ls /etc/resolv.conf 2>/dev/null | grep "/etc/resolv.conf"' failed
# 0d41931b116e135b0a32cfc361b68e25fdfa1a0cf8d8153f77c492581116a338
```

also from #523:
```
not ok 3 Exit Code from container process when running under a terminal
# (in test file exit-code.bats, line 52)
#   `[ "${status}" -eq "${exit_status}" ]' failed
# 7e89e7c6b16d
# 
# 0
Dubious, test returned 1 (wstat 256, 0x100)

not ok 3 Check that pseudo tty is setup properly when allocating a tty
# (in test file terminal.bats, line 27)
#   `echo "${output}" | grep "$tty_dev"' failed
# d1c92879937d

```

devimc:
This patch removes all references to STDIN_FILENO
in process_tests since if it is modified all
future tests will be affected causing bugs like #527

fixes #527

### DIFF
--- a/tests/process_test.c
+++ b/tests/process_test.c
@@ -198,8 +198,8 @@ START_TEST(test_cc_oci_setup_shim) {
 	int tmpf2_fd = -1;
 
 	ck_assert (! cc_oci_setup_shim (NULL, -1, -1));
-	ck_assert (! cc_oci_setup_shim (NULL, STDIN_FILENO, -1));
-	ck_assert (! cc_oci_setup_shim (NULL, -1, STDIN_FILENO));
+	ck_assert (! cc_oci_setup_shim (NULL, tmpf1_fd, -1));
+	ck_assert (! cc_oci_setup_shim (NULL, -1, tmpf1_fd));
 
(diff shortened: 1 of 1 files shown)


sameo:
LGTM

jcvenegas:
if using stdin was causing the problem, should we need to modify this tests as well https://github.com/01org/cc-oci-runtime/blob/master/tests/util_test.c#L525 ? 
note:  that that test use (closes stdin fd) but restore it after it finish

gorozco1:
lgtm

chavafg:
qa-failed

PR #545, opened 2016-12-22
Fix docker exec and test memory leak

sameo:
The CRI-O merge broke docker exec and introduced a test memory leak.
We only saw these issues when using cc-oci-runtime directly and not cc-oci-runtime.sh.

The console argument we receive is not NULL but empty (to signal us that we simply got `--console`) is set. So we need to check for the argument being non NULL and not empty to determine that we're using a terminal.

With the CRI-O merge we introduced a pointless config update that was breaking the interactive docker exec use case.

The `pod_test.c` tests were allocating a config pointer without releasing it. This is fixed as well.

### DIFF
--- a/src/commands/exec.c
+++ b/src/commands/exec.c
@@ -165,8 +165,13 @@ handler_exec (const struct subcommand *sub,
 
 	process->user.uid = start_data.user.uid;
 	process->user.gid = start_data.user.gid;
-	process->terminal = (start_data.console != NULL ? true : false);
 	process->env = env;
+	if ((start_data.console != NULL) && (start_data.console[0])) {
+		process->terminal = true;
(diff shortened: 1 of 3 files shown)


sameo:
LGTM

jcvenegas:
lgtm , tested here

chavafg:
qa-failed

grahamwhaley:
lgtm   Will wait for the qa cycle to re-spin...

devimc:
👍

gorozco1:
lgtm

sameo:
@devimc You're right, thanks for spotting that one.

Although this memory leak was not related to the goal of this PR (fixing docker exec), it is now fixed.

PR #546, opened 2016-12-22
Tests: Validation for the stdout and stderr

GabyCT:
This tests are to check issue #171
They are using the run command

### DIFF
--- a/tests/integration/docker/run.bats
+++ b/tests/integration/docker/run.bats
@@ -33,3 +33,12 @@ setup() {
 	[ "${status}" -eq 0 ]
 	[[ "${output}" == 'Passed'* ]]
 }
+
+@test "stdout using run" {
+	$DOCKER_EXE run busybox sh -c "ls /etc/resolv.conf 2>/dev/null | grep "/etc/resolv.conf""
+}
+
+@test "stderr using run" {
+	skip "this is not working (https://github.com/01org/cc-oci-runtime/issues/171)"
+	$DOCKER_EXE run busybox sh -c "if ls /etc/foo > /dev/null; then false; else true; fi"
+}


tests/integration/docker/run.bats
```diff
@@ -33,3 +33,12 @@ setup() {
...
 	[[ "${output}" == 'Passed'* ]]
 }
+
+@test "stdout using run" {
+	$DOCKER_EXE run busybox sh -c "ls /etc/resolv.conf 2>/dev/null | grep "/etc/resolv.conf""
+}
+
+@test "stderr using run" {
+	skip "this is not working (https://github.com/01org/cc-oci-runtime/issues/171)"
+	$DOCKER_EXE run busybox sh -c "if ls /etc/foo > /dev/null; then false; else true; fi"
```
gorozco1: What are you expecting here.. :-/ trying to understand... 

When running with --runtime runc I got the output of stderr or stdout in the result... but not true or false (ie 0 or 1)

f=`docker run --runtime runc -it  busybox sh -c "if ls /etc/foo >/dev/null; then false; else true; fi"`

chavafg:
qa-failed

tests/integration/docker/run.bats
GabyCT: @gorozco1  if you ran the tests for stderr using cor the exit code is always 0 even if the file does not exists.

GabyCT: @gorozco1 and I am using the same technique that was requested for the exec tests (stdout and stderr) but now using the run command

GabyCT: @gorozco1 what I am trying to validate here it is that /etc/foo does not exists or otherwise it should failed the test

gorozco1: @GabyCT is this for both runc and cor ? or only cor?

GabyCT: @gorozco1 only cor as currently our docker integration tests are using only cor

gorozco1: @GabyCT thanks for clarify

GabyCT:
qa-passed

devimc:
lgtm

PR #554, opened 2016-12-30
shim: atexit returns 0 if successful

wcwxyz:
Signed-off-by: WANG Chao <wcwxyz@gmail.com>

### DIFF
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -691,7 +691,7 @@ main(int argc, char **argv)
 	}
 
 	ret = atexit(restore_terminal);
-	if ( !ret) {
+	if (ret) {
 		shim_debug("Could not register function for atexit");
 	}
 


grahamwhaley:
Hi @wcwxyz - I'm presuming you found this from code reading, or did you see it in a real use case?
A little more explanation in the merge and/or commit message might be nice :-), but ...
lgtm

thanks!

grahamwhaley:
Hi @wcwxyz - we also noticed (thanks @jodh-intel) that our pullapprove auto-checker is rejecting the commit due to what it thinks is a 'missing signed off by' - but, we can see the signed-off by line is indeed there :-)  What we suspect is that pullapprove has not spotted the SOB tag as there is not commit body text or blank line before it (even though pullapprove do not mention those being necessary here: http://docs.pullapprove.com/requirements/signed_off_by/ )
I suggest we see if that gets cleared when you add some body text to the commit - and if not, then we will investigate further and maybe enquire with pullapprove.

sameo:
@wcwxyz Thanks for the PR. A few lines of commit message would be helpful indeed.

wcwxyz:
@grahamwhaley Actually I'm working on a similar project like cc-oci-runtime (cor). It requires something like cc-shim, so I settled on cc-shim. I spotted the atexit error message on the screen after cc-shim started.

wcwxyz:
Guys, I've updated the commit. It seems like pullapprove SOB check passed.

jodh-intel:
@wcwxyz - thanks!

lgtm.

gorozco1:
lgtm

gorozco1:
@chavafg no tests results yet ?

chavafg:
there were some failures in our functional tests, but those are fixed in PR 551, which is now merged. I think we can merge this PR

chavafg:
qa-passed

dlespiau:
Merging then, thanks @wcwxyz !

wcwxyz:
Thanks for review :)

PR #555, opened 2017-01-04
shim: Handle stdin in non-tty case.

### LINKED ISSUE #506: docker run/exec does not handle input from pipe
Clear containers 2.1 does not handle input  running a workload to read data from a pipe.
the next command  works  running with runc
```
echo hello  | sudo  docker run --runtime runc -i debian  grep hello
```
but not for clear  containers
```
echo hello  | sudo  docker run  -i debian  grep hello
```
The same behavior happens in docker exec command.

amshinde:
This makes it possible to handle all input in non-tty case.
Fixes #506.

### DIFF
--- a/shim/shim.c
+++ b/shim/shim.c
@@ -39,6 +39,11 @@
 
 /* globals */
 
+struct pollfd poll_fds[MAX_POLL_FDS] = {{-1}};
+
+// stdin is always added at index 3 in the poll_fds array
+#define STDIN_INDEX 3
+
 /* Pipe used for capturing signal occurence */
 int signal_pipe_fd[2] = { -1, -1 };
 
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

shim/shim.c
```diff
@@ -39,9 +39,14 @@
 
 /* globals */
 
+// stdin is always added at index 3 in the poll array
+#define STDIN_INDEX 3
```
jodh-intel: Might be slightly safer to make this `(MAX_POLL_FDS-1)`?

I do wonder if we should change `add_pollfd()` so that it is told which index to assign the provided fd to make this code easier to maintain.

shim/shim.c
```diff
@@ -39,9 +39,14 @@
 
 /* globals */
 
+// stdin is always added at index 3 in the poll array
```
jodh-intel: `s/poll array/poll_fds array/`. Also, why not put that array immediately below the `#define` for clarity?

amshinde: @jodh-intel yeah, I'll change that.

shim/shim.c
amshinde: @jodh-intel Yes I was thinking the same too. I'll add that change as well.

GabyCT:
this PR will be re-test

sameo:
LGTM

dlespiau:
lgtm

GabyCT:
qa-passed

GabyCT:
The PR was re-tested and no failures

PR #556, opened 2017-01-05
tests: Add docker test for handling stdin from pipe.

amshinde:
Test for issue #506.

### DIFF
--- a/tests/integration/docker/run.bats
+++ b/tests/integration/docker/run.bats
@@ -42,3 +42,11 @@ setup() {
 	skip "this is not working (https://github.com/01org/cc-oci-runtime/issues/171)"
 	$DOCKER_EXE run busybox sh -c "if ls /etc/foo > /dev/null; then false; else true; fi"
 }
+
+@test "stdin from pipe" {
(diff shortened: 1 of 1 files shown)


jcvenegas:
lgtm

chavafg:
qa-passed

dlespiau:
lgtm

tests/integration/docker/run.bats
```diff
@@ -42,3 +42,11 @@ setup() {
 	skip "this is not working (https://github.com/01org/cc-oci-runtime/issues/171)"
 	$DOCKER_EXE run busybox sh -c "if ls /etc/foo > /dev/null; then false; else true; fi"
 }
+
+@test "stdin from pipe" {
+	run bash -c "echo hello | $DOCKER_EXE run -i busybox"
+	echo status: "${status}"
```
dlespiau: do we need to keep what seems to be debug echos? does it matter at all?

amshinde: The output from the echos is seen only when the test fail. Would be useful in failure cases.

dlespiau: okay, merging then.

PR #557, opened 2017-01-05
CRI exec fixes

sameo:
Support for CRI-O exec was broken:

- The config pod pointer is empty when calling exec on cc-oci-runtime. Therefore, we can not know which VM to attach to when being called from CRI-O
- CRI-O calls exec without a process.json file and thus we're giving hyperstart a process with an empty CWD to exec and that fails.

This PR fixes both issues.

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -1606,7 +1606,9 @@ cc_proxy_hyper_exec_command (struct cc_oci_config *config)
 		json_array_add_object_element (envs, env_var);
 	}
 
-	json_object_set_string_member (process_node, "workdir", process->cwd);
+	if (process->cwd[0]) {
+		json_object_set_string_member (process_node, "workdir", process->cwd);
+	}
 	json_object_set_array_member (process_node, "args", args);
 	json_object_set_array_member (process_node, "envs", envs);
 	json_object_set_object_member (payload,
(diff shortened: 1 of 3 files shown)


chavafg:
qa-failed

chavafg:
qa-passed

src/proxy.c
```diff
@@ -1606,7 +1606,9 @@ cc_proxy_hyper_exec_command (struct cc_oci_config *config)
 		json_array_add_object_element (envs, env_var);
 	}
 
-	json_object_set_string_member (process_node, "workdir", process->cwd);
+	if (process->cwd[0]) {
```
dlespiau: This should segfault if cwd is null. That said the rest of the code is also doing this check without checking if cwd if null beforehand, so probably missing something.

dlespiau:
lgtm

src/proxy.c
sameo: @dlespiau `process` is pointing at `config->oci.process` and cwd is defined as follow:

```
struct oci_cfg_process {
	gchar              **args;

	/** Full path to working directory to run workload command in. */
	gchar                cwd[PATH_MAX];
[...]
```

So it can't be NULL if `process` is not NULL. So if a check is needed here, it probably should be on `process` itself, which is a separate issue.

dlespiau: ah fair enough, should have looked there!

sboeuf:
LGTM

PR #560, opened 2017-01-09
proxy fd leak & fd leak detector

dlespiau:
@devimc noticed a fd leak when killing the proxy savagely, it was due to the proxy never receiving the Bye payload.

Instead, the proxy now cleans up the per-VM resources when it detects the qemu process is gone.

I've added a generic "fd leak detector" that can help catch fd leaks and instrumented the proxy unit tests to check for those. A neat little thing that can be reused in other go projects:

  https://github.com/dlespiau/cc-oci-runtime/blob/20170109-fd-leak-detector/proxy/fdleak_test.go

This leaves the case where the qemu process never goes away. It shouldn't happen, but we may want to improve our robustness with a mechanism killing qemu "after some time". It's not quite clear at this point if we should do that, it'd hide bugs and we'd have to decide where to put that mechanism (delete of the runtime? proxy?)

### DIFF
--- a/proxy/fdleak_test.go
+++ b/proxy/fdleak_test.go
@@ -1,0 +1,493 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"reflect"
+	"sort"
+	"strconv"
+	"strings"
+	"syscall"
+	"testing"
+)
+
+func fcntl(fd int, cmd int, arg int) (val int, err error) {
+	r0, _, errno := syscall.Syscall(
+		syscall.SYS_FCNTL, uintptr(fd), uintptr(cmd), uintptr(arg))
+	val = int(r0)
+	if errno != 0 {
+		err = errno
+	}
+	return
+}
+
+// FdLeakDetector detects fd leaks by letting its user take snapshots of the
+// list of opened fds at key points in the profiled program and compare two
+// snapshots.
+type FdLeakDetector struct {
+}
+
+type FdInfo struct {
+	Fd          int
+	Flags       int
+	CloseOnExec bool
+	Text        string
+}
+
+func (info *FdInfo) dump(w io.Writer) {
+	flags := make([]string, 5)
+
+	if info.CloseOnExec {
+		flags = append(flags, "cloexec")
+	}
+
+	// file status
+	if info.Flags&syscall.O_APPEND != 0 {
+		flags = append(flags, "append")
+	}
+	if info.Flags&syscall.O_NONBLOCK != 0 {
+		flags = append(flags, "nonblock")
+	}
+
+	// acc mode
+	if info.Flags&syscall.O_RDONLY != 0 {
+		flags = append(flags, "read-only")
+	}
+	if info.Flags&syscall.O_RDWR != 0 {
+		flags = append(flags, "read-write")
+	}
+	if info.Flags&syscall.O_WRONLY != 0 {
+		flags = append(flags, "write-only")
+	}
+	if info.Flags&syscall.O_DSYNC != 0 {
+		flags = append(flags, "dsync")
+	}
+	if info.Flags&syscall.O_RSYNC != 0 {
+		flags = append(flags, "rsync")
+	}
+	if info.Flags&syscall.O_SYNC != 0 {
+		flags = append(flags, "sync")
+	}
+
+	fmt.Fprintf(w, "  %d: %s (%s)\n", info.Fd, info.Text,
+		strings.Join(flags, ""))
+}
+
+func (info *FdInfo) equal(other *FdInfo) bool {
+	return reflect.DeepEqual(info, other)
+}
+
+type FdSnapshot struct {
+	Fds []FdInfo
+}
+
+// ByFd implements sort.Interface for []FdInfo based on the Fd field.
+type ByFd []FdInfo
+
+func (a ByFd) Len() int           { return len(a) }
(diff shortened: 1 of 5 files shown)


chavafg:
qa-failed

chavafg:
qa-passed

proxy/fdleak_test.go
```diff
@@ -0,0 +1,499 @@
...
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"syscall"
+	"testing"
+)
+
+func fcntl(fd int, cmd int, arg int) (val int, err error) {
+	r0, _, e1 := syscall.Syscall(
```
jodh-intel: Couldn't you just use `err` here (which would also allow you to remove the error test below)?

proxy/fdleak_test.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+	if info.Flags&syscall.O_RSYNC != 0 {
+		flags += "rsync "
+	}
+	if info.Flags&syscall.O_SYNC != 0 {
+		flags += "sync "
+	}
+
+	fmt.Fprintf(w, "  %d: %s\n", info.Fd, info.Text)
+	fmt.Fprintf(w, "   : %s\n", flags)
```
jodh-intel: niggle - you'll end up with a trailing space here :)

It's not going to block this PR, but you could maybe make `flags` an array and then `Join` the values to avoid the easy-to-forget-to-add trailing spaces in the string values above).

proxy/fdleak_test.go
```diff
@@ -0,0 +1,499 @@
...
+		return false
+	}
+	if info.CloseOnExec != other.CloseOnExec {
+		return false
+	}
+	if info.Text != other.Text {
+		return false
+	}
+
+	return true
```
jodh-intel: Wouldn't `reflect.DeepEqual()` be safer to guard against any future change to `Fdinfo`? There might be a perf hit, but this is test code.

proxy/proxy.go
```diff
@@ -357,6 +372,11 @@ func proxyMain() {
 		os.Exit(1)
 	}
 	proxy.serve()
+
+	// Wait for all the goroutines started by helloHandler to finish. Not
+	// stricly necessary, as the process is about to exit, but it's eye
+	// please to match wg.Add() by the corresponding wg.Wait()
```
jodh-intel: "but it's eye-pleasing to match wg.Add() with the corresponding wg.Wait()."?

proxy/vm.go
```diff
@@ -301,3 +306,13 @@ func (vm *vm) Close() {
 	// Wait for VM global goroutines
 	vm.wg.Wait()
 }
+
+// OnVmLost returns a channel that can be waiting on to be signalled the end
+// of the qemu process.
+func (vm *vm) OnVMLost() <-chan interface{} {
```
jodh-intel: "can be waited on to signal the end of the qemu process."?

proxy/fdleak_test.go
dlespiau: not quite, this err is a bit special, it's an instance of Errno and it needs special treatment: https://golang.org/pkg/syscall/#Errno

jodh-intel: ugh - in which case maybe rename `e1` to `errno` to make that clearer?

proxy/proxy.go
dlespiau: The purpose of wg.Add() is to be able to wait on a number of goroutines to wg.Wait() until all of them are finished. The test code does a useful wg.Wait() to ensure we end up in a clean state, but the proxy being an ever lasting daemon, the wg.Wait() here is never executed in the current state of things.

dlespiau:
PR updated:

- Used a slice of strings and strings.Join to build the flags string
- Used reflect.DeepEqual in FdInfo.Equal
- Renamed e1 to errno in fnctl
- Reworded a godoc comment that went through too many rewrites

proxy/fdleak_test.go
```diff
@@ -0,0 +1,489 @@
...
+		t.Fatal()
+	}
+}
+
+func TestFdDetectorCompare(t *testing.T) {
+	tests := []struct {
+		old, new *FdSnapshot
+		equal    bool
+	}{
+		// Same fds
```
jodh-intel: I don't this needs to block the PR, but this static slice is huge. I'd be tempted to define individual tests such that you could then define `tests` as:

```
type snapshotTest struct {
    old, new *FdSnapshot
    equal bool
}

tests := []snapshotTest = {
    test_same_fds,
    test_same_fds_different_flags,
    test_same_fds_different_cloexe,
      :
}
```

That would make the file easier to maintain and would also mean you could optionally drop the comments as the tests would document themselves.

proxy/fdleak_test.go
```diff
@@ -0,0 +1,489 @@
...
+
+	new, err := detector.Snapshot()
+	if err != nil {
+		t.Error(err)
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	equal := detector.Compare(buffer, old, new)
+	if buffer.Len() != 0 {
+		t.Fatal()
```
jodh-intel: Shouldn't `buffer` be displayed to show the differences here?

proxy/fdleak_test.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+	if !equal {
+		t.Fatal()
```
jodh-intel: And here?

proxy/fdleak_test.go
```diff
@@ -0,0 +1,489 @@
...
+
+	new, err := detector.Snapshot()
+	if err != nil {
+		t.Error(err)
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	equal := detector.Compare(buffer, old, new)
+	if equal {
+		t.Fatal()
```
jodh-intel: Print `buffer`?

proxy/proxy.go
jodh-intel: Ok. My comment was simply to suggest a rewording as it currently doesn't parse fully in my mind :)

grahamwhaley: s/it's eye please/it reads better/  maybe?

proxy/fdleak_test.go
grahamwhaley: I agree - having run across these missing in another project that has caused @jodh-intel and myself reasonable pain tracking down and adding in more verbose failure logs, I vote for putting them in now :-) This one, and obviously the others...

dlespiau:
PR updated:

- Print the output of Compare() on error in the tests
- Reword confusing comment
- Fixed a hunk that ended up in the wrong commit during a rebase

jodh-intel:
lgtm.

sameo:
LGTM

PR #562, opened 2017-01-09
2.1.0-rc.3 release

sameo:
- Added docker tests for handling pipe input
- Added pprof data export through HTTP
- Fixed CRI-O exec support
- Fixed shim atexit return value
- Fixed docker pipe inout handling (Issue #506)
- Fixed version functional test

Shortlog:

c22f33f proxy: Do not set the process CWD if it's empty
dbc634d oci: Build an accurate pod pointer before exec'ing
42acfd9 tests: Add explicit checking for stdout and stderr redirection tests.
a5a4f2d tests: Add docker test for handling stdin from pipe.
ba6e92d proxy: Add a way to serve pprof data over HTTP
80ceb96 shim: Add file descriptors at specifix index.
5d7ac88 shim: Handle stdin in non-tty case.
73e7d95 shim: atexit returns 0 if successful
baca26f test: Fix version functional test
6be54d4 tests: fix cc_oci_setup_shim test

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,0 +1,7 @@
+2.1.0-rc.3:
+    Added docker tests for handling pipe input
+    Added pprof data export through HTTP
+    Fixed CRI-O exec support
+    Fixed shim atexit return value
+    Fixed docker pipe inout handling (Issue #506)
+    Fixed version functional test
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@
 #
 
 AC_PREREQ([2.69])
-AC_INIT([cc-oci-runtime], [v2.1.0-rc.2])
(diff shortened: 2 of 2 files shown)


dlespiau:
lgtm

dlespiau:
It may be nice to maintain a NEWS file or equivalent with the summary you've given in the PR.

sboeuf:
@dlespiau that's a good idea to summarize each release news, but I would maybe name this file RELEASES instead of NEWS.

sboeuf:
LGTM

sameo:
@dlespiau Good point, NEWS file added
@sboeuf `NEWS` is a GNU standard, so we may want to stick to that.

sboeuf:
@sameo Ah I didn't know that, let's go for NEWS then !

PR #568, opened 2017-01-10
tests: show functional tests output in tap format

chavafg:
This commit makes functional tests to show its output in tap format
(ok and not ok) instead of using ✗ and ✓
This change will help to search for failures.

### DIFF
--- a/data/run-bats.sh.in
+++ b/data/run-bats.sh.in
@@ -25,5 +25,5 @@
 cmd="@BATS_PATH@"
 #Networking is not configured using non-root user
 #run bats functional tests with network namespace unshared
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

dlespiau:
The commit doesn't seem to match the commit message?

sboeuf:
@dlespiau What is not matching ? The commit enables -t option to output functional tests format to tap format, is that right ?

dlespiau:
oh my bad, I read the "-t" as an option to unshare.

lgtm

sboeuf:
LGTM

PR #573, opened 2017-01-12
Better feedback when running go static checks

dlespiau:
We don't really have information on which test is running and which has failed. On top of that, the gofmt check was consuming stdout and so we were losing the list of guilty files failing the format check.

Running the checks now displays something like:

```
Doing go static checks on packages: ./proxy ./proxy/api
Running misspell...
Running go vet...
Running gofmt...
/home/damien/go/src/github.com/01org/cc-oci-runtime/proxy/fdleak_test.go
Makefile:5230: recipe for target 'check-go' failed
```

### DIFF
--- a/.ci/ci-go-static-checks.sh
+++ b/.ci/ci-go-static-checks.sh
@@ -69,12 +69,22 @@ install_package github.com/golang/lint/golint
 
 echo Doing go static checks on packages: $go_packages
 
+echo "Running misspell..."
 go list -f '{{.Dir}}/*.go' $go_packages |\
     xargs -I % bash -c "misspell -error %"
+
+echo "Running go vet..."
 go vet $go_packages
+
+echo "Running gofmt..."
 go list -f '{{.Dir}}' $go_packages |\
     xargs gofmt -s -l | wc -l |\
     xargs -I % bash -c "test % -eq 0"
+
+echo "Running cyclo..."
 go list -f '{{.Dir}}' $go_packages | xargs gocyclo -over 15
 
+echo "Running golint..."
 for p in $go_packages; do golint -set_exit_status $p; done
+
+echo "All Good!"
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

chavafg:
there was a false positive on a swarm test, after retesting everything was OK.

chavafg:
qa-passed

.ci/ci-go-static-checks.sh
```diff
@@ -69,12 +69,22 @@ install_package github.com/golang/lint/golint
...
     xargs -I % bash -c "misspell -error %"
+
+echo "Running go vet..."
 go vet $go_packages
+
+echo "Running gofmt..."
 go list -f '{{.Dir}}' $go_packages |\
-    xargs gofmt -s -l | wc -l |\
-    xargs -I % bash -c "test % -eq 0"
+    xargs gofmt -s -l | tee /dev/tty | \
```
devimc: Wouldn't be better to use tty command?

grahamwhaley:
taking into account the feedback from @devimc ,
lgtm

.ci/ci-go-static-checks.sh
dlespiau: why would it be better to use the tty command? :) [controlling terminal](http://tldp.org/HOWTO/Text-Terminal-HOWTO-7.html#ss7.3) of the process group (/dev/tty) Vs the path of the tty if stdin is indeed a terminal (tty)

It's also not very handy to do a onliner with tty. tty works by doing a readlink() on /proc/self/fd/0, which is of course a pipe when following |, so we can't use it on that line:

```
$ ls /etc/passwd |  echo `tty` 
not a tty
```

and so:

```
$ ls /etc/passwd | tee `tty` > /dev/null
$ ls /etc/passwd | tee /dev/tty > /dev/null
/etc/passwd
$
```

devimc: 👍

devimc:
lgtm

PR #577, opened 2017-01-13
namespaces: enable mount namespace

### LINKED ISSUE #325: Mounted files in /etc are zeroed out after docker cp
Started a container with cc-runtime. Verified that /etc/resolve.conf is not empty within the container and has the contents as provided in /run/docker/libcontainerd/$container_id/resolv.conf. Then copied a file from the host to the clear container

```
$ sudo docker cp ./testfile 34f30fc0b76c:/root/testfile
```

After this, the /etc/resolv.conf file inside the container is empty. Same behaviour seen for /etc/hostname and /etc/hosts.
All these files are present in the mount section of config.json and are bind mounted by the runtime.

This behaviour is not seen with runc.

devimc:
A new mount namespace MUST be created before mounting anything
since dockerd remounts container's filesystem when certain commands
are executed by docker causing issues like #325

fixes #325

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -58,6 +58,7 @@
 #include "command.h"
 #include "proxy.h"
 #include "pod.h"
+#include "namespace.h"
 
 extern struct start_data start_data;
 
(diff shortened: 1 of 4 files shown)


chavafg:
qa-passed

grahamwhaley:
lgtm

src/oci.c
```diff
@@ -561,6 +562,11 @@ cc_oci_create (struct cc_oci_config *config)
 		return false;
 	}
 
+	if (! cc_oci_ns_setup (config)) {
```
jodh-intel: `cc_oci_ns_setup()` is currently called in `cc_oci_vm_launch()` (which is called soon after `cc_oci_handle_mounts()` below). Can we therefore drop the call in `cc_oci_vm_launch()` ?

devimc: sure , np :)

grahamwhaley:
@jodh-intel OK with this now?

jodh-intel:
lgtm

sameo:
LGTM

PR #580, opened 2017-01-17
proxy: Reorder response and fd

dlespiau:
In:

  commit 50d42dc434f283ccb6e5daa17ea98d376b86fc4c
  Author: Samuel Ortiz <sameo@linux.intel.com>
  Date:   Mon Dec 19 15:00:25 2016 +0100

      proxy: Simplify fd passing and the message read loop

Samuel changed the order to send the fd before the response and simplify
the reading loop.

In this commit, I put it back the other way: response, then fd. It
allows for a more natural checking of the response status before trying
to receive the fd. If the allocateIO fails, we don't even try to read a
fd from the proxy.

That change is motivated by another one: we now have a
cc_proxy_receive_fd() function that is orthogonal to reading the
response back from the proxy. That means we can use it several times in
a row to retrieve more than one fd, a useful feature we'd like to use
very soon.

### DIFF
--- a/proxy/protocol.go
+++ b/proxy/protocol.go
@@ -128,7 +128,14 @@ func (proto *protocol) Serve(conn net.Conn, userData interface{}) error {
 		// Execute the corresponding handler
 		resp := proto.handleRequest(ctx, &req, &hr)
 
-		// First send an fd if the handler associated a file with the response
+		// Send the response back to the client.
+		if err = api.WriteMessage(conn, resp); err != nil {
+			// Something made us unable to write the response back
+			// to the client (could be a disconnection, ...).
+			return err
+		}
+
+		// And send a fd if the handler associated a file with the response
 		if hr.file != nil {
 			if err = api.WriteFd(conn.(*net.UnixConn), int(hr.file.Fd())); err != nil {
 				return err
@@ -136,11 +143,5 @@ func (proto *protocol) Serve(conn net.Conn, userData interface{}) error {
 			hr.file.Close()
 		}
 
-		// Then send the response back to the client.
-		if err = api.WriteMessage(conn, resp); err != nil {
-			// Something made us unable to write the response back
-			// to the client (could be a disconnection, ...).
-			return err
-		}
 	}
 }
--- a/proxy/protocol.go
+++ b/proxy/protocol.go
@@ -138,6 +138,7 @@ func (proto *protocol) Serve(conn net.Conn, userData interface{}) error {
 		// And send a fd if the handler associated a file with the response
 		if hr.file != nil {
 			if err = api.WriteFd(conn.(*net.UnixConn), int(hr.file.Fd())); err != nil {
+				hr.file.Close()
 				return err
 			}
 			hr.file.Close()
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -42,13 +42,10 @@ struct watcher_proxy_data
 	GString     *msg_received;
 	int          socket_fd;
 	/**
-	 * indicates that we expect an out-of-band file descriptor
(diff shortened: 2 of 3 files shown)


proxy/protocol.go
```diff
@@ -128,19 +128,20 @@ func (proto *protocol) Serve(conn net.Conn, userData interface{}) error {
...
+			// to the client (could be a disconnection, ...).
+			return err
+		}
+
+		// And send a fd if the handler associated a file with the response
 		if hr.file != nil {
 			if err = api.WriteFd(conn.(*net.UnixConn), int(hr.file.Fd())); err != nil {
 				return err
 			}
 			hr.file.Close()
```
grahamwhaley: Not directly related to your patch, but as you are in the area...
It looks like if we err on the WriteFd(), then we return before the hd.file.Close() - which looks like a potential file handle leak?
I note there is no file.Close() in any of the other error paths either - it feels a little odd and inconsistent. Any thoughts (though this should not block this PR, as that code was untouched I think)

chavafg:
qa-failed

src/proxy.c
```diff
@@ -225,124 +222,125 @@ cc_proxy_disconnect (struct cc_proxy *proxy)
...
-	/* The entire message was read */
-	if (bytes_read >= payload_length + MESSAGE_HEADER_LENGTH) {
+	/*
+	 * Start by reading the message header and the payload length.
+	 */
+	bytes_read = read(proxy_data->socket_fd, header, sizeof(header));
+	if (bytes_read <= 0) {
+		g_critical("couldn't read header from proxy: %s",
+			   strerror(errno));
 		goto out;
```
grahamwhaley: I've a feeling `goto out` now goes to some code (the msg_received->len check code) that is redundant, as at this point `len` will always be 0. That appears to be the only jump to out:, so maybe it can be moved further down at the bottom of the function? (unless len can be incremented somewhere else I have missed?)

grahamwhaley:
phew, that is a dense patch :-)  Looking over the code nothing critical jumped out, just a couple of minor things. As I'm not overly familiar with that code, will appreciate and wait for review from others.

devimc:
```sh -e -c "while :; do docker run busybox ls; echo ====; done"``` passed

lgtm

proxy/protocol.go
dlespiau: Good catch! will add a commit on top to fix that.

src/proxy.c
dlespiau: right, that's still correct :) but indeed we can move the test outside of the error path. Will do.

dlespiau:
@grahamwhaley It is indeed a big patch and I'm afraid there's no way to split it in smaller commits if we want to keep each commit working.

PR updated:
- Added patch to fix a fd leak in an error path (spotted by Graham)
- Moved the msg_received->len check in cc_proxy_msg_read() outside of the error path in
- Added a check on payload_length so we don't blindly trust data coming from the wire

grahamwhaley:
Thanks. The 'magic' '1024' length cap in the middle of the file could have maybe been a const somewhere or maybe could be defined by the protocol header or something, but it's not a blocker, so:
lgtm

grahamwhaley:
I see we still have a qa-failed on this, and that was before some updates - so @chavafg can we re-run qa on this one please?

dlespiau:
If the goal is to remove allocateIO altogether, we don't really need this PR (we can, hopefully, write the replacement for allocateIO soon-ish).

dlespiau:
Also, split out the fd leak fix in a separate PR  (#610) in case we want to close that one.

sameo:
LGTM

Issue #553, opened 2016-12-28
ENV variables not persevered across build layers.


pdxjohnny:
When building a Dockerfile which uses the ENV command it will not persevere those environment variables across build layers.

For example if you were to build this Dockerfile with runc as your default runtime it would update fine. But if cor is default it hangs because the proxy variable was not preserved.
```
FROM ubuntu
ENV http_proxy http://proxy.example.com:1080/
RUN apt -y update
```

docker info (changed default to runc before running info)
```
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 7
Server Version: 1.12.1
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 8
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: null host bridge overlay
Swarm: inactive
Runtimes: cor runc
Default Runtime: runc
Security Options: apparmor seccomp
Kernel Version: 4.4.0-31-generic
Operating System: Ubuntu 16.04.1 LTS
OSType: linux
Architecture: x86_64
CPUs: 16
Total Memory: 31.42 GiB
Name: ubuntu
ID: 33NB:6JG2:ECXL:G3FC:4KU7:TEBV:Y6B6:ZB23:YYZJ:RY7N:FOEH:VJYD
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Http Proxy: http://proxy.example.com:1080/
No Proxy: .intel.com,10.0.0.0/8,192.168.0.0/16,localhost,127.0.0.0/8,134.134.0.0/16
Registry: https://index.docker.io/v1/
Insecure Registries:
 127.0.0.0/8
```

jcvenegas:
I tested this using  2.1 (59db2cc63a6)
```
FROM ubuntu
ENV http_proxy LOCAL_PROXY
ENV https_proxy LOCAL_PROXY
ENV VAR var_data
RUN apt-get -y update
RUN apt-get -y install vim
RUN echo $VAR $(uname -a) $http_proxy $https_proxy
```
Build the new image:
```
$ sudo docker build -t cc-envvars .                                                                                                                                 
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM ubuntu
latest: Pulling from library/ubuntu
b3e1c725a85f: Pull complete
[...]

PR #590, opened 2017-01-19
test: Add "docker build" integration test

jcvenegas:
Add initial bats test to verify "docker build"
compatibility.

Verifies support for #553 

### DIFF
--- a/tests/integration/docker/build.bats
+++ b/tests/integration/docker/build.bats
@@ -1,0 +1,52 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+SRC="${BATS_TEST_DIRNAME}/../../lib/"
+IMG_NAME="ccbuildtests"
+
+
+setup() {
+	source $SRC/test-common.bash
+	runtime_docker
+}
+
+teardown () {
+	echo "teardown:"
+	$DOCKER_EXE rmi $IMG_NAME
+}
+
+@test "docker build env vars" {
+	var_value="test_env_vars"
+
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

tests/integration/docker/build.bats
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
```
grahamwhaley: (C) 2017 :-)

tests/integration/docker/build.bats
```diff
@@ -0,0 +1,50 @@
...
+@test "docker build env vars" {
+	var_value="test_env_vars"
+
+	run $DOCKER_EXE build -t "${IMG_NAME}" - <<EOF
+	FROM busybox
+	ENV VAR "${var_value}"
+	RUN sh -c 'echo $VAR'
+EOF
+	echo output: "${output}"
+	echo "${output}" | grep "${var_value}"
```
grahamwhaley: Silly question. Did we 'negative test' this? I just have a niggle that if it failed, maybe it would return something like 'could not set test_env_vars' in the output, and that would still match the grep? Maybe we can set the var to something more comprehensive (guessing: test_env_vars=test_value or similar) and/or make the grep more complex to ensure we are picking up the env setting and not an error message? Just a thought.

grahamwhaley:
lgtm
But I did post a note and a query (if we could make the test more robust maybe).
And it looks like we need to understand the qa-failed as well.

tests/integration/docker/build.bats
grahamwhaley: Ah, I've just seen #553 - so maybe my comment is not applicable - please respond as appropriate :-)

chavafg:
I got again this error: 

```
 ✗ HostName is passed to the container
   (from function `clean_docker_ps' in file /home/cc-tester/go/src/github.com/01org/cc-oci-runtime/cc-oci-runtime-2.1.0-rc.3/tests/integration/docker/../../lib//test-common.bash, line 28,
    from function `setup' in test file /home/cc-tester/go/src/github.com/01org/cc-oci-runtime/cc-oci-runtime-2.1.0-rc.3/tests/integration/docker/network.bats, line 28)
     `clean_docker_ps' failed with status 123
   Error response from daemon: Cannot kill container 3b169d9555fa: Container 3b169d9555fa1b06c79261cceed49268473cd7379603180a92cabc380ffbe805 is not running
```

We need to investigate why this appears, anything else looks good.

tests/integration/docker/build.bats
jcvenegas: updated :)

tests/integration/docker/build.bats
jcvenegas: @grahamwhaley  I tried to make it more robust by checking  if the env var is set using "env" command, like this : 
```
env | grep "VAR=${var_value}" 
```
and check docker build exit code in case it fail

jcvenegas:
@chavafg seems that that error comes from other test (network.bats), I dont see how this test affect network.bats, should we create an issue for that ?

chavafg:
@jcvenegas yes, I created #593, moving to qa-passed

chavafg:
qa-passed

devimc:
lgtm

Issue #463, opened 2016-11-25
cc-proxy died when trying to kill cc-shim with killall


chavafg:
I got a cc-shim process that was not stopped after the container was shutdown.
I killed it with `killall cc-shim` but `cc-proxy` process was also killed. 

stderror from cc-proxy when it was killed:

```
+ ./cc-proxy
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x40459a]

goroutine 635 [running]:
panic(0x526b80, 0xc42000c130)
        /usr/lib/golang/src/runtime/panic.go:500 +0x1a1
main.(*vm).writeMessage(0xc4203d0380, 0xc42003cd80, 0xc400000012, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/vm.go:150 +0x15a
main.(*vm).SendMessage(0xc4203d0380, 0xc4200742a0, 0xd, 0xc4200700c0, 0x5d, 0x60, 0x0, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/vm.go:350 +0x136
main.hyperHandler(0xc4200b8090, 0x82, 0x90, 0x513ae0, 0xc4203a9ce0, 0xc4203a9fa0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/proxy.go:198 +0x154
main.(*protocol).handleRequest(0xc420082020, 0xc42003cf20, 0xc4203baab0, 0xc4203a9fa0, 0x0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/protocol.go:95 +0xca
main.(*protocol).Serve(0xc420082020, 0x5e0940, 0xc42002b6f8, 0x513ae0, 0xc4203a9ce0)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/protocol.go:130 +0x1ad
main.(*proxy).serveNewClient(0xc420084100, 0xc420082020, 0x5e0940, 0xc42002b6f8)
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/proxy.go:250 +0xb5
created by main.(*proxy).serve
        /root/go/src/github.com/01org/cc-oci-runtime/proxy/proxy.go:270 +0x3eb
```

dlespiau:
[...]

PR #594, opened 2017-01-19
vendor: Revendor virtcontainers to 0.2.3 to fix a panic

dlespiau:
Let's vendor virtcontainers 0.2.3, it's a good version if we want to
minimize the changes while grabbing a fix for a panic that happens if
the shim disappears underneath us.

It's debatable if we should jump straight away to 0.3.0. That version
would also need us to re-vendor the runv package. It will have to
happen, eventually, but we can control when.

An benefit of going slowly on the updates is that we can revert a later
vendoring to 0.3.0 if we find out it introduces regressions.

Fixes: https://github.com/01org/cc-oci-runtime/issues/463

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -21,7 +21,7 @@
 	},
 	"github.com/containers/virtcontainers": {
 		"url": "https://github.com/containers/virtcontainers.git",
-		"version": "v0.2.1",
+		"version": "v0.2.3",
 		"license": "Apache 2.0"
 	},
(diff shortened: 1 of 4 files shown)


jcvenegas:
lgtm

chavafg:
qa-failed

devimc:
lgtm

grahamwhaley:
Looks like the qa fail is the swarm one that I believe we are tracking down, so am merging this...

PR #595, opened 2017-01-19
Futureproofing api.client.Hello and api.Client.Attach

dlespiau:
Those two proxy payloads will have some additional (optional) parameters, sometimes with the corresponding return values. In fact, we already have an optional parameter: the qemu console path that the runtime can provide to the proxy to display hyperstart messages.

So, we need a way to provide those optional parameters. This PR introduces the following pattern:

```
        ret, err := client.Hello(containerID, ctlSocketPath, ioSocketPath,
                &api.HelloOptions{
                        Console: "/path/to/qemu/console",
        })
```

I could have introduced a HelloWithOptions that take that 3rd argument, but thought that might has well change it now while we can (unrelease API only used by unreleased tip of virtcontainers).

This will be used by followup PRs:
  - The introduction of a version field in the Hello/Attach response
  - The introduction of a new parameter for Hello/Attach to return a virtual hyperstart ctl fd

### DIFF
--- a/proxy/api/client.go
+++ b/proxy/api/client.go
@@ -112,20 +112,38 @@ func errorFromResponse(resp *Response) error {
 	return nil
 }
 
+// HelloOptions holds extra arguments one can pass to the Hello function. See
+// the Hello payload for more details.
+type HelloOptions struct {
+	Console string
+}
+
+// HelloReturn contains the return values from Hello. See the Hello and
+// HelloResult payloads.
+type HelloReturn struct {
+}
+
 // Hello wraps the Hello payload (see payload description for more details)
-func (client *Client) Hello(containerID, ctlSerial, ioSerial string) error {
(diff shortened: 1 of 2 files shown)


chavafg:
qa-failed

grahamwhaley:
@dlespiau - was that update in response to the qa-failed, which generated a:
`docker: Error response from daemon: shim error: open pid: no such file or directory.`
If yes, then @chavafg , we should re-run the qa.
If no, then we need to investigate the failure ?

dlespiau:
This PR changes come library code that is isn't used any where in this repo (but that is used in virtcontainers). We can safely ignore the qa tests here as long as the travis ones are green (they are the ones with the go checks)

grahamwhaley:
OK, qa failure is being looked at (not specific to this PR), so:
lgtm

jodh-intel:
lgtm

Issue #593, opened 2017-01-19
Investigate why we got :  Error response from daemon: Cannot kill container 


chavafg:
Recently on the CI, we have got this error very frequently (almost on every execution the error appears).

```
not ok 1 HostName is passed to the container 
# (from function `clean_docker_ps' in file ../../lib//test-common.bash, line 28, 
# from function `setup' in test file network.bats, line 28) 
# `clean_docker_ps' failed with status 123 
# Error response from daemon: Cannot kill container 93606e5bd45d: Container 93606e5bd45d05718640e7747238f546534bd7b88363e05138230b240d71932c is not running 
```

grahamwhaley:
My immediate thought is that we can expand the code in https://github.com/01org/cc-oci-runtime/blob/master/tests/lib/test-common.bash#L28 to check if the `kill` and `rm` calls work. In particular, as we are seeing the `kill` fail in some cases as the container is `not running`, maybe we can print out a `status` dump if the command fails for instance?

chavafg:
It can be reproduced by running: 
```
for i in {1..100}; do prove -v logs.bats network.bats || break ; done
```
On cc-oci-runtime.log: 
```
2017-01-19T14:56:51.393474Z:7939:cc-oci-runtime:debug:message read from proxy socket: {"success":true}
2017-01-19T14:56:51.393500Z:7939:cc-oci-runtime:debug:msg received: {"success":true}
2017-01-19T14:56:51.393510Z:7939:cc-oci-runtime:debug:unmounting /var/lib/docker/overlay/f89ad6475cdfe06be2e8717704775d4cc642362134854d7e3edcbcdf0b5ba12a/merged/etc/resolv.conf
2017-01-19T14:56:51.393525Z:7939:cc-oci-runtime:critical:failed to umount /var/lib/docker/overlay/f89ad6475cdfe06be2e8717704775d4cc642362134854d7e3edcbcdf0b5ba12a/merged/etc/resolv.conf
2017-01-19T14:56:51.393541Z:7939:cc-oci-runtime:critical:failed to stop container d769801b922a2f244b3195603d44b72c18bd288011ac8839e64b6a149db3dc21
```

grahamwhaley:
I see one of the logs mentions failure to umount resolve.conf - could this be in any way related to #570 maybe?

chavafg:
Also maybe related to #577 . I did some retest before #577 was merged and I didnt get the failure.

devimc:
[...]

PR #597, opened 2017-01-19
tests: Add docker test to check environment.

amshinde:

### DIFF
--- a/tests/integration/docker/env.bats
+++ b/tests/integration/docker/env.bats
@@ -1,0 +1,43 @@
+#!/usr/bin/env bats
+# *-*- Mode: sh; sh-basic-offset: 8; indent-tabs-mode: nil -*-*
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


devimc:
lgtm

tests/integration/docker/env.bats
```diff
@@ -0,0 +1,42 @@
...
+
+setup() {
+	source $SRC/test-common.bash
+	clean_docker_ps
+	runtime_docker
+}
+
+@test "Verify LANG is not set in env" {
+	skip "this is not working (see https://github.com/01org/cc-oci-runtime/issues/324)"
+	if $DOCKER_EXE run -i ubuntu env | grep LANG; then false; else true; fi
```
dlespiau: ah, you don't seem to know about grep -v :)

```
$ echo ENV | grep ENV >/dev/null; echo $?
0
$ echo ENV | grep -v ENV >/dev/null; echo $?
1
```

dlespiau:
lgtm

tests/integration/docker/env.bats
amshinde: Just a copy from a previous test there :)

chavafg:
qa-failed

grahamwhaley:
In the qa, the 'exec' test *immediately after* this one failed - which I thus suspect may mean this test had some dangling effect that rolled onto the next one - so, @amshinde , can we look into that?

chavafg:
retested, got 1 failure wrt #593, everything else looks ok

chavafg:
qa-passed

PR #598, opened 2017-01-19
runtime: verify if the system supports virtualization

### LINKED ISSUE #596: Feature request: Provide error messages about missing KVM support on command line if cor runtime crashes back to prompt
Recently I did a bios upgrade and it turned off my virtualization without my knowledge.  When I tried to run a clear container (--runtime=cor) it just crashed back to the prompt and the error logs using journalctl -u docker.service did not help me realize this.  It would be helpful if I got some kind of indication that kvm was not enabled or working.   In general if any common error messages especially those related to the VM are passed back to the command prompt it would have helped me in this case.

devimc:
This patch verify if the system supports virtualization
reporting an issue when it has not support

fixes #596

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -31,6 +31,8 @@
 #include "oci-config.h"
 #include "priv.h"
 
+#define KVM_PATH "/dev/kvm"
+
 /* globals */
 static char *program_name;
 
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

jcvenegas:
probably /dev/kvm would be enough to check if kvm is enabled, also when qemu launch a VM it send to stderr that is not able to start it if KVM is not supported, I wonder if would be more useful always log  qemu stderr in the runtime log so we can not only detect the missing kvm support but other critical error.

dlespiau:
I was about to say the same thing as @jcvenegas, we can't forget we have to try minimise start time. A single syscall to check the existence of /dev/kvm seems enough to me.

devimc:
commit updated :-)

grahamwhaley:
lgtm

dlespiau:
lgtm

jcvenegas:
lgtm

sameo:
LGTM

PR #599, opened 2017-01-19
packaging: Add option to set GOPATH (packaging, tests, automated builds)

gorozco1:
We need to set GOPATH variable if it not previously defined.
This aims make easier to build cc-oci-runtime in different distros.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,7 @@ AUTOMAKE_OPTIONS = parallel-tests
 
 CHECK_DEPS =
 FUNCTIONAL_TESTS_DEPS =
+PROXY_DEPS =
 
 # Has a value if building in a git tree
 GIT_COMMIT := $(shell git rev-parse HEAD 2>/dev/null)
@@ -250,8 +251,21 @@ systemdservicedir   = $(systemdsystemunitdir)
 systemdservice_DATA = $(systemdservice_files)
 endif
 
+if AUTOGOPATH
+export GOPATH := $(shell mktemp -d --suffix=-cor-gopath)
+PKG_LINK_BASE = $(GOPATH)/src/github.com/01org
+PKG_LINK = $(PKG_LINK_BASE)/cc-oci-runtime
+
+$(PKG_LINK):
+		mkdir -p $(PKG_LINK_BASE)
+		ln -sfn $(CURDIR) $(PKG_LINK)
+		cp -r $(CURDIR)/vendor/* $(GOPATH)/src
+
+PROXY_DEPS += $(PKG_LINK)
+endif
+
 proxy_ldflags = "-X main.DefaultSocketPath=$(localstatedir)/run/cc-oci-runtime/proxy.sock"
-cc-proxy: $(cc_proxy_sources)
+cc-proxy: $(cc_proxy_sources) | $(PROXY_DEPS)
 	$(AM_V_GO)go build -o $@ -ldflags=$(proxy_ldflags) $(srcdir)/proxy
 
 cc_proxy_sources =			\
@@ -276,7 +290,7 @@ cc_proxy_extra_dist =			\
 
 CHECK_DEPS += check-proxy
 
-check-proxy:
+check-proxy: $(cc_proxy_sources) | $(PROXY_DEPS)
 	$(AM_V_GEN)proxy_test_common_args="-v -timeout 2s $(srcdir)/proxy" ; \
 	go test -race $$proxy_test_common_args || go test $$proxy_test_common_args
 
@@ -756,6 +770,9 @@ check: $(CHECK_DEPS)
 
 clean-local:
 	rm -f commit_id
+if AUTOGOPATH
+	rm -rf /tmp/*cor-gopath
+endif
 if FUNCTIONAL_TESTS
 if AUTO_BUNDLE_CREATION
 	$(AM_V_GEN)echo "Deleting generated bundle '$(BUNDLE_TEST_PATH)'"
--- a/configure.ac
+++ b/configure.ac
@@ -302,6 +302,17 @@ AC_ARG_WITH([cc-image],
     ]
 )
 
+AC_ARG_ENABLE([autogopath],
+	AS_HELP_STRING([--enable-autogopath], [Set GOPATH variable for you, used when packaging @<:@default=false@:>@]),
+	[case "${enableval}" in
+		yes) autogopath=true ;;
+		 no) autogopath=false ;;
+		  *) AC_MSG_ERROR([bad value ${enableval} for --enable-autogopath]) ;;
+	esac],
+	[autogopath=false]
+)
+AM_CONDITIONAL([AUTOGOPATH], [test x$autogopath = xtrue])
+
 AC_OUTPUT
 
 dnl === Summary ===============================================================


Makefile.am
```diff
@@ -275,7 +287,14 @@ cc_proxy_extra_dist =			\
 CHECK_DEPS += check-proxy
 
 check-proxy:
-	go test -v -race -timeout 2s $(srcdir)/proxy
+	[ -n "$$GOPATH" ] && \
+		( \
+			go test -v -race -timeout 2s $(srcdir)/proxy ;\
```
jcvenegas: if [ -z $GOPATH] then \\
set GOPATH \\
fi \\
got test  ...

chavafg:
qa-failed

Makefile.am
```diff
@@ -250,7 +250,19 @@ endif
...
+			$(AM_V_GO)go build -o $@ -ldflags=$(proxy_ldflags) $(srcdir)/proxy ;\
+		) || \
+		( \
+			export GOPATH=$$(readlink -f $(srcdir))/.gopath ; \
+			rm -rf "$$GOPATH"; \
+			mkdir "$$GOPATH"; \
+			mkdir -p "$$GOPATH"/src/github.com/01org/ ; \
+			ln -sf $$(readlink -f $(srcdir)) "$$GOPATH"/src/github.com/01org/cc-oci-runtime ; \
+			cp -r $(srcdir)/vendor/* "$$GOPATH/src" ; \
+			go build -o $@ -ldflags=$(proxy_ldflags) $(srcdir)/proxy ;\
```
gorozco1: If I use $(AM_V_GO) it breaks...

gorozco1: I tried to do that but the GOPATH variable only lives in the if sentence...

Makefile.am
gorozco1: same here, I tried to set the GOPATH variable only if is not defined...

Makefile.am
jodh-intel: This looks a bit complex to my eyes. I suggest we turn this into a `.sh.in` script which gets called by `Makefile.am`. That way:

- You can add comments to explain what is happening, without having them appear in the make output.
- You won't need to use the "double-dollar" syntax which is difficult to read.
- You won't need to use the semi-colons.

Overall, I think a separate script (containing `set -e` at the top) should be easier to maintain and would make the logic clearer.

Ideally, the script would live in a `scripts/` directory. We haven't got one of those yet but this PR could always introduce one (we cant then move the scripts currently in `data/` to `scripts/`).

Please can you make sure that "`make dist` continues to work (it should include the new script).

Makefile.am
amshinde: @gorozco1 Have you tried exporting the variable ?

gorozco1: @amshinde I will re-factor this PR based on comments.

Makefile.am
gorozco1: @jodh-intel thanks for the feedback updating this

Makefile.am
jodh-intel: Before landing this PR we might want to wait until #613 has landed as it looks like that one will conflict with your `check-proxy` change.

jodh-intel: Hi @gorozco1 - #613 has now landed, so you can rebase on `master`.

dlespiau:
For this one, I do like the simple solution runc seem to have found, linking the root of the repo in the GOPATH AFAICT

amshinde:
@gorozco1  Can you take a look at @dlespiau 's suggestion? They do the linking as a makefile target, adding the target as a dependency instead of making explicit shell checks here. Would simplify things.

gorozco1:
@dlespiau @amshinde yes.. for sure, thanks for the suggestion, @jcvenegas also proposed another  solution: https://github.com/jcvenegas/cc-oci-runtime/commit/da6eac0c2df1d556c53831512cb0b22ade2b3772 I'll take a look at the runc and update this PR

chavafg:
qa-passed

configure.ac
```diff
@@ -302,6 +302,17 @@ AC_ARG_WITH([cc-image],
     ]
 )
 
+AC_ARG_ENABLE([autogopath],
+	AS_HELP_STRING([--enable-autogopath], [enable autogopath @<:@default=false@:>@]),
```
jodh-intel: Could you add a bit more detail here? I think the description should mention `$GOPATH` (and ideally why you'd want to use this option if you already have a `$GOPATH` set).

gorozco1: @jodh-intel This option will help to automate builds on OBS, OBS does not have GOPATH defined and it has to be defined, and prepare the environment to build the package.

jodh-intel: Sure - I know what it's for, but I think it would be useful to specify those words inside the `AS_HELP_STRING()` call as folk outside the project who want to build the code might wonder why we have this option ;)

gorozco1:
@jodh-intel updated

jcvenegas:
@sameo @devimc can you take a look ?

Makefile.am
```diff
@@ -250,8 +251,21 @@ systemdservicedir   = $(systemdsystemunitdir)
 systemdservice_DATA = $(systemdservice_files)
 endif
 
+if AUTOGOPATH
+export GOPATH := $(CURDIR)/.gopath
+PKG_LINK := $(CURDIR)/.gopath/src/github.com/01org/cc-oci-runtime
+
+$(PKG_LINK):
+		rm -rf $(CURDIR)/.gopath
+		mkdir -p  $(CURDIR)/.gopath/src/github.com/01org
```
jodh-intel: This path is part of `PKG_LINK` and `PKG_`LINK` isn't using `GOPATH`. I'd suggest doing something like:

```
PKG_LINK_BASE := $(GOPATH)/src/github.com/01org
PKG_LINK := $(PKG_LINK_BASE)/cc-oci-runtime

   :
   rm -rf $(GOPATH)
   mkdir -p $(PKG_LINK_BASE)
```

Makefile.am
```diff
@@ -250,8 +251,21 @@ systemdservicedir   = $(systemdsystemunitdir)
...
+		ln -sfn $(CURDIR) $(PKG_LINK)
+		cp -r $(CURDIR)/vendor/* $(CURDIR)/.gopath/src
```
jodh-intel: `cp -r $(CURDIR)/vendor/* $(GOPATH)/src`

Makefile.am
```diff
@@ -276,7 +290,7 @@ cc_proxy_extra_dist =			\
 
 CHECK_DEPS += check-proxy
 
-check-proxy:
+check-proxy: $(cc_proxy_sources) | $(PROXY_DEPS)
```
jodh-intel: ooh - our first order-only prereq ;-)

Makefile.am
gorozco1: thanks for the suggestion @jodh-intel

Makefile.am
gorozco1: ;-)

jodh-intel:
lgtm

Makefile.am
```diff
@@ -250,8 +251,22 @@ systemdservicedir   = $(systemdsystemunitdir)
 systemdservice_DATA = $(systemdservice_files)
 endif
 
+if AUTOGOPATH
+export GOPATH := $(CURDIR)/.gopath
```
jodh-intel: Oh - just spotted that this is writing to cwd - this isn't going to be reliable - we should really use something like:

```
export GOPATH := $(mktemp -d --suffix=-cor-gopath)
```

configure.ac
```diff
@@ -302,6 +302,17 @@ AC_ARG_WITH([cc-image],
     ]
 )
 
+AC_ARG_ENABLE([autogopath],
+	AS_HELP_STRING([--enable-autogopath], [Set GOPATH variable for you, used when packaging @<:@default=false@:>@]),
```
jodh-intel: I'm wondering if we want to have this default to `true` (and rename and change behaviour so it can be disabled with `--disable-autogopath`). @sameo, @dlespiau - what do you think?

gorozco1:
@jodh-intel @sameo any update on this?

jodh-intel:
I still wonder if we need to always enable this by default (outstanding question for some of the reviewers), but other than that...

lgtm.

gorozco1:
@jodh-intel @sameo updated commit, default=true

Makefile.am
```diff
@@ -756,6 +770,7 @@ check: $(CHECK_DEPS)
 
 clean-local:
 	rm -f commit_id
+	rm -rf /tmp/*cor-gopath
```
gorozco1: good catch

configure.ac
jodh-intel: @sboeuf - any thoughts on this?

sboeuf: @jodh-intel yes I agree it would make more sense to have autogopath enabled by default.

sameo:
LGTM

PR #601, opened 2017-01-19
Support plain "go build" for the proxy

dlespiau:
We currently rely on a link-time option to specifiy the path of the proxy domain socket. We can provide a reasonable fallback in case the proxy is just compiled with "go build".

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -277,11 +277,6 @@ var ArgSocketPath = flag.String("socket-path", "", "specify path to socket file"
 func (proxy *proxy) init() error {
 	var l net.Listener
 	var err error
-	var socketPath = DefaultSocketPath
-
-	if len(*ArgSocketPath) != 0 {
-		socketPath = *ArgSocketPath
-	}
 
 	// flags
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

proxy/proxy.go
```diff
@@ -299,6 +294,18 @@ func (proxy *proxy) init() error {
...
+		// populate DefaultSocketPath, so fallback to a reasonable
+		// path.
+		if DefaultSocketPath == "" {
+			DefaultSocketPath = "/var/run/cc-oci-runtime/proxy.sock"
+		}
+
+		socketPath := DefaultSocketPath
+		if len(*ArgSocketPath) != 0 {
+			socketPath = *ArgSocketPath
+		}
```
jcvenegas: IMHO feel this more like if else  :P

jcvenegas:
lgtm

proxy/proxy.go
grahamwhaley: @jcvenegas indeed :-)

grahamwhaley:
qa failure looks unrelated, spurious, so:
lgtm

Issue #519, opened 2016-12-07
systemd cc-proxy.socket cannot be started in Fedora 24 when selinux=enforced


chavafg:
I am trying to start `cc-proxy` via systemctl, but I got some failures:

```
[cc-tester@fedora-cc-ci cc-oci-runtime]$ sudo systemctl start cc-proxy.socket
Job for cc-proxy.socket failed. See "systemctl status cc-proxy.socket" and "journalctl -xe" for details.


[cc-tester@fedora-cc-ci cc-oci-runtime]$ systemctl status cc-proxy.socket
● cc-proxy.socket - Clear Containers Proxy Socket
   Loaded: loaded (/usr/lib/systemd/system/cc-proxy.socket; disabled; vendor preset: disabled)
   Active: failed (Result: resources)
     Docs: https://github.com/01org/cc-oci-runtime/proxy
   Listen: /var/run/cc-oci-runtime/proxy.sock (Stream)

Dec 07 11:28:58 fedora-cc-ci.intel.com systemd[1]: Failed to listen on Clear Containers Proxy Socket.
Dec 07 11:50:43 fedora-cc-ci.intel.com systemd[1]: cc-proxy.socket: Failed to listen on sockets: Permission denied
Dec 07 11:50:43 fedora-cc-ci.intel.com systemd[1]: Failed to listen on Clear Containers Proxy Socket.
Dec 07 11:52:07 fedora-cc-ci.intel.com systemd[1]: cc-proxy.socket: Failed to listen on sockets: Permission denied
Dec 07 11:52:07 fedora-cc-ci.intel.com systemd[1]: Failed to listen on Clear Containers Proxy Socket.
Dec 07 11:54:51 fedora-cc-ci.intel.com systemd[1]: cc-proxy.socket: Failed to listen on sockets: Permission denied
Dec 07 11:54:51 fedora-cc-ci.intel.com systemd[1]: Failed to listen on Clear Containers Proxy Socket.
Dec 07 11:55:06 fedora-cc-ci.intel.com systemd[1]: Closed Clear Containers Proxy Socket.
Dec 07 11:55:11 fedora-cc-ci.intel.com systemd[1]: cc-proxy.socket: Failed to listen on sockets: Permission denied
Dec 07 11:55:11 fedora-cc-ci.intel.com systemd[1]: Failed to listen on Clear Containers Proxy Socket.

```

I needed to set selinux to permissive to be able to initiate cc-proxy.socket:
```
[...]

PR #612, opened 2017-01-23
selinux: Initial commit for clear-containers selinux modules

gorozco1:
selinux: Initial commit for cc-proxy selinux module
Initial commit to allow cc-proxy to run with selinux enforced.
This branch fixes issue #519 "systemd cc-proxy.socket cannot
be started in Fedora 24 when selinux=enforced"


chavafg:
qa-failed

dlespiau:
Could you put the file in the proxy/ directory and the amend the existing README there?

Also, more than the links on the bug, I'd rather have the explanation on how to use the policy.

sameo:
Adding @ereshetova for review.
Elena, this is supposed to fix #519, and we'd appreciate if you could review the te file.

sameo:
@gorozco1 Please also specify which issue this is fixing, in your commit message.

ereshetova:
There is no reason not to combine permissions:

+allow init_t unconfined_service_t:unix_stream_socket listen;
+
+#!!!! This avc is allowed in the current policy
+allow init_t unconfined_service_t:unix_stream_socket { bind create setopt };

Should become:

+allow init_t unconfined_service_t:unix_stream_socket { bind create setopt listen };

I think base selinux doesn't have macros, so just calling out exact access types are ok. 

One think I would double check in your case is that the process running in the unconfined_service_t domain that needs new allow to be added is actually supposed to run in that domain. Just to double check that no mistakes were done.

gorozco1:
@ereshetova updated module thanks for your feedback.

gorozco1:
@dlespiau updated module, path and readmes

sameo:
@gorozco1 2 comments:
1. Can we move the selinux directory under `proxy/` ? We're going to move the proxy code into its own repo, so better separate the selinux specific bits now.
2. I'm fine with carrying the spec and deb files, but I'd like it to be in a dedicated directory where we gather all meta file for each distro.

gorozco1:
@sameo done, moved to proxy and removed the spec file.

proxy/README.md
```diff
@@ -113,6 +113,42 @@ handled by systemd and can be viewed with:
 journalctl -u cc-proxy -f
 ```
 
+## SELinux
+
+`cc-proxy` can't run out of the box on fedora systems running SELinux enforced. 
+This is documented in issue #519
```
dlespiau: Can you remove this sentence? the necessary information should be part of this README, the "how we got there" does't really matter to the end user.

sameo:
LGTM

jcvenegas:
lgtm

PR #613, opened 2017-01-23
make: Add test-program to check -race is supported with go

amshinde:

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -274,8 +274,23 @@ cc_proxy_extra_dist =			\
 
 CHECK_DEPS += check-proxy
 
+define CONTENT
+package main
+import "fmt"
+func main() {
+	fmt.Printf("Executing test program")
+}
+endef
+export CONTENT
+
 check-proxy:
-	go test -v -race -timeout 2s $(srcdir)/proxy
+	$(AM_V_GEN)go_test_extra_args="" ; \
+	echo "$$CONTENT">./racetest.go ; \
+	if go build -race ./racetest.go; then \
+		go_test_extra_args+="-race" ; \
+	fi ; \
+	rm ./racetest* ; \
+	go test -v "$$go_test_extra_args" -timeout 2s $(srcdir)/proxy
 
 check-go:
 	@$(top_srcdir)/.ci/ci-go-static-checks.sh


chavafg:
qa-failed

sameo:
@amshinde Could you please describe in your commit log why would `-race` not be supported ?

amshinde:
@sameo Added some description in the commit message. This is related to #494.
But since I force pushed on the closed PR, github does not allow me to open the PR.

sameo:
LGTM

jcvenegas:
lgtm
@amshinde just a question reading your commit message I wonder if we could check for the go compiler version instead of try to compiler a test program with race flag

dlespiau:
A question I had was: is this enough to check go test can actually compile and link the test binary (which is the reason why we try to run go test -race instead of just checking the version, it wasn't working on some Ubuntu version IIRC).

One can ask go test to output the test binary with the `-o` option. Doing it without any test file doesn't produce any binary for me:

```
$ ll
total 4
-rw-rw-r--. 1 damien damien 81 Jan 25 16:15 foo.go
$ cat foo.go 
package main
import "fmt"
func main() {
	fmt.Printf("Executing test program")
}
$ go test -race -o test-binary .
?   	_/home/damien/articles/foo	[no test files]
$ ll
total 4
-rw-rw-r--. 1 damien damien 81 Jan 25 16:15 foo.go
```

Addind a dummy _test.go file does help:

```
$ cat foo_test.go 
package main
import "testing"
func TestFoo(t *testing.T) {
}
$ go test -race -o test-binary .
ok  	_/home/damien/articles/foo	1.008s
$ ll
total 3564
-rw-rw-r--. 1 damien damien      81 Jan 25 16:15 foo.go
-rw-rw-r--. 1 damien damien      63 Jan 25 16:19 foo_test.go
-rwxrwxr-x. 1 damien damien 3640298 Jan 25 16:19 test-binary
```

amshinde:
Yes, ubuntu 16.04 in OBS did show -race in the build flags, but the actual build with -race was broken as  it needed race-detection package. This package was causing go itself to break!

@dlespiau I am checking if go build succeeds with the -race flag instead of go test. Do you see any issues with that ?

dlespiau:
Oh, I hadn't noticed it was go build. Well, we use -race with go test, but I guess that may not matter, not sure. Let's say nevermind my comment :)

sameo:
LGTM

sameo:
@chavafg Can we please re-test? I think the failures were due to another bug that should be fixed now.

jodh-intel:
Hi @chavafg - could you schedule a re-test?

Makefile.am
```diff
@@ -274,8 +274,23 @@ cc_proxy_extra_dist =			\
...
+	fmt.Printf("Executing test program")
+}
+endef
+export CONTENT
+
 check-proxy:
-	go test -v -race -timeout 2s $(srcdir)/proxy
+	echo "$$CONTENT">./racetest.go
+	if ! go build -race ./racetest.go; then \
+		go test -v -race -timeout 2s $(srcdir)/proxy; \
```
jodh-intel: It would be better to create a variable for the extra argument, rather than having two almost identical `go test` calls:

```
go_test_extra_args=
  :
# set if "go build" supports -race
go_test_extra_args+= -race
  :
go test -v $(go_test_extra_args) -timeout 2s $(srcdir)/proxy
```

That would make the code clearer, cleaner and easier to maintain.

chavafg:
qa-passed

Makefile.am
gorozco1: @amshinde could you address @jodh-intel suggestion?

amshinde: @gorozco1 Pushed changed addressing @jodh-intel 's suggestions.

jodh-intel:
lgtm.

PR #615, opened 2017-01-23
 mount: fix umount function

### LINKED ISSUE #325: Mounted files in /etc are zeroed out after docker cp

### LINKED ISSUE #593: Investigate why we got :  Error response from daemon: Cannot kill container

devimc:
support for mount namespace was add in PR #577 to fix #325
all resources mounted by the runtime are part of a new
mount namespace hence when delete is called the runtime is
unable to unmount them. This patch fixes that issue joining
to the namespace and unmounting all resources

fixes #593

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -27,6 +27,7 @@
 
 #include "mount.h"
 #include "common.h"
+#include "namespace.h"
 
 /** Mounts that will be ignored.
  *
@@ -330,9 +331,36 @@ gboolean
 cc_oci_handle_unmounts (const struct cc_oci_config *config)
 {
 	GSList  *l;
+	struct oci_cfg_namespace *ns;
+	gboolean mountns = false;
 
 	if (! config) {
 		return false;
+	}
+
+	/**
+	 * At this point qemu is not running if there is
+	 * a mount namespace with a path we have join
+	 * it and umount all mounted resources
+	 */
+	for (l = config->oci.oci_linux.namespaces;
+			l && l->data;
+			l = g_slist_next (l)) {
+		ns = (struct oci_cfg_namespace *)l->data;
+
+		if (ns->type == OCI_NS_MOUNT && ns->path) {
+			mountns = cc_oci_ns_join (ns);
+			break;
+		}
+	}
+
+	/**
+	 * if there is NOT a specific mount namespace return true
+	 * since namespace created by unshare in \ref cc_oci_ns_setup
+	 * is destroyed when qemu ends
+	 */
+	if (! mountns) {
+		return true;
 	}
 
 	/* umount files and directories */
--- a/src/namespace.c
+++ b/src/namespace.c
@@ -36,17 +36,25 @@
 #include "util.h"
 
 /** Map of \ref oci_namespace values to human-readable strings. */
-static struct cc_oci_map oci_ns_map[] = 
-{
-	{ OCI_NS_CGROUP  , "cgroup"  },
-	{ OCI_NS_IPC     , "ipc"     },
-	{ OCI_NS_MOUNT   , "mount"   },
-	{ OCI_NS_NET     , "network" },
-	{ OCI_NS_PID     , "pid"     },
-	{ OCI_NS_USER    , "user"    },
-	{ OCI_NS_UTS     , "uts"     },
+static struct cc_oci_ns_map {
+	/* namespace value */
+	const enum oci_namespace ns;
 
-	{ OCI_NS_INVALID , NULL      }
+	/* string representation for namespace */
+	const gchar* name;
+
+	/* set to true if namespace is supported by runtime */
+	const gboolean supported;
+} oci_ns_map[] = {
+	{ OCI_NS_CGROUP  , "cgroup"  , false },
+	{ OCI_NS_IPC     , "ipc"     , false },
+	{ OCI_NS_MOUNT   , "mount"   , true  },
+	{ OCI_NS_NET     , "network" , true  },
+	{ OCI_NS_PID     , "pid"     , false },
+	{ OCI_NS_USER    , "user"    , false },
+	{ OCI_NS_UTS     , "uts"     , false },
+
+	{ OCI_NS_INVALID , NULL      , false}
 };
 
 /*!
@@ -66,6 +74,28 @@ cc_oci_ns_free (struct oci_cfg_namespace *ns)
 }
 
 /**
+ * check if namespace is supported
+ *
+ * \param ns \ref oci_namespace.
+ *
+ * \return true is namespace is supported else false.
+ */
+static gboolean
+cc_oci_ns_supported (enum oci_namespace ns)
+{
+	struct cc_oci_ns_map  *p;
+
+	for (p = oci_ns_map; p && p->name; p++) {
+		if (p->ns != ns) {
+			continue;
+		}
+		return p->supported;
+	}
+
+	return false;
+}
+
+/**
  * Convert a \ref oci_namespace into a human-readable string.
  *
  * \param ns \ref oci_namespace.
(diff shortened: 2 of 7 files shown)


src/namespace.c
```diff
@@ -152,7 +152,7 @@ cc_oci_str_to_ns (const char *str)
  */
 gboolean cc_oci_ns_join(struct oci_cfg_namespace *ns)
 {
-	int fd = -1;
+	int fd;
```
jcvenegas: any special reason to not initialize this fd, it is reassigned in the test lines, but if the code changes in the future would be better to initialize  it

src/state.c
```diff
@@ -284,6 +287,37 @@ handle_state_mounts_section(GNode* node, struct handler_data* data) {
...
+	}
+	if (! (node->children && node->children->data)) {
+		g_critical("%s missing value", (char*)node->data);
+		return;
+	}
+
+	if (! g_strcmp0(node->data, "type")) {
+		n = g_new0 (struct oci_cfg_namespace, 1);
+		n->type = cc_oci_str_to_ns((char*)node->children->data);
+		data->state->namespaces = g_slist_append(data->state->namespaces, n);
```
jcvenegas: data is used but not does not check if is not null

src/state.c
```diff
@@ -284,6 +287,37 @@ handle_state_mounts_section(GNode* node, struct handler_data* data) {
...
 
 /*!
+ * handler for namespaces section
+ *
+ * \param node \c GNode.
+ * \param data \ref handler_data.
+ */
+static void
+handle_state_namespaces_section(GNode* node, struct handler_data* data) {
+	struct oci_cfg_namespace* n;
```
jcvenegas: probably you would like to initialize with NULL

src/mount.c
```diff
@@ -331,13 +331,15 @@ gboolean
...
 {
 	GSList  *l;
-	struct oci_cfg_namespace *ns;
-	gboolean mountns = false;
 
 	if (! config) {
 		return false;
 	}
 
+#ifndef UNIT_TESTING
```
jcvenegas: why it is needed just with UNIT_TESTING?

chavafg:
qa-failed

src/namespace.c
amshinde: In any case, the above seems like a random commit.

src/namespace.c
```diff
@@ -115,6 +144,56 @@ cc_oci_str_to_ns (const char *str)
...
+			type ? type : "", ns->path);
+		goto err;
+	}
+
+	close (fd);
+
+	/* reset to ensure any subsequent failure is
+	 * handled correctly.
+	 */
+	fd = -1;
```
amshinde: Do you need to set the fd here, since you are immediately returning after this statement.

src/mount.c
```diff
@@ -330,11 +331,37 @@ gboolean
...
+		if (ns->type == OCI_NS_MOUNT && ns->path) {
+			mountns = cc_oci_ns_join (ns);
+			break;
+		}
+	}
+
+	/**
+	 * if there is NOT a specific mount namespace returns true
+	 * since namespace crated by unshare in \ref cc_oci_ns_setup
+	 * is destoryed when qemu ends
```
amshinde: typo s/destoryed/destroyed/

src/mount.c
amshinde: Same question. Also, this could go in a separate commit itself.

src/mount.c
```diff
@@ -330,11 +331,37 @@ gboolean
...
+
+		if (ns->type == OCI_NS_MOUNT && ns->path) {
+			mountns = cc_oci_ns_join (ns);
+			break;
+		}
+	}
+
+	/**
+	 * if there is NOT a specific mount namespace returns true
+	 * since namespace crated by unshare in \ref cc_oci_ns_setup
```
sameo: s/crated/created/

src/mount.c
```diff
@@ -330,11 +331,37 @@ gboolean
...
+		ns = (struct oci_cfg_namespace *)l->data;
+
+		if (ns->type == OCI_NS_MOUNT && ns->path) {
+			mountns = cc_oci_ns_join (ns);
+			break;
+		}
+	}
+
+	/**
+	 * if there is NOT a specific mount namespace returns true
```
sameo: s/returns/return/

src/state.c
devimc: according to documentation https://developer.gnome.org/glib/2.26/glib-Singly-Linked-Lists.html#g-slist-append

```
g_slist_append ()
Returns : the new start of the GSList
```

src/state.c
devimc: is not needed

src/mount.c
devimc: I have to remove this line and fix mount test

src/mount.c
```diff
@@ -335,6 +336,35 @@ cc_oci_handle_unmounts (const struct cc_oci_config *config)
 		return false;
 	}
 
+#ifndef UNIT_TESTING
+	struct oci_cfg_namespace *ns;
+	gboolean mountns = false;
+
+	/**At this point qemu is not running if there is
```
devimc: nice catch

src/mount.c
devimc: nice catch

src/namespace.c
devimc: nice catch

src/mount.c
```diff
@@ -335,6 +336,35 @@ cc_oci_handle_unmounts (const struct cc_oci_config *config)
...
+	 * a mount namespace with a path we have join to
```
devimc: nice catch

src/namespace.c
```diff
@@ -115,6 +144,56 @@ cc_oci_str_to_ns (const char *str)
...
+	int saved;
+	const gchar *type;
+
+	type = cc_oci_ns_to_str (ns->type);
+
+	fd = open (ns->path, O_RDONLY);
+
+	if (fd < 0) {
+		saved = errno;
+		g_critical ("failed to open %s", ns->path);
```
devimc: agree

src/namespace.c
```diff
@@ -115,6 +144,56 @@ cc_oci_str_to_ns (const char *str)
 }
 
 /**
+ * Join to a specific namespace
```
devimc: enough,  going back to English classes  x_x

src/state.c
jcvenegas: Sorry, I mean the "struct handler_data* data" parameter is not checked but you access to its members

devimc: ohhh ok, fixing...

src/namespace.c
devimc: 1. cc_oci_ns_join is a new function
2. cppcheck reports errors if fd is initialized here
3. fd is initialized below

jcvenegas:
lgtm

sameo:
LGTM

PR #621, opened 2017-01-25
docs: Remove the section about starting the hypervisor

dlespiau:
This whole section is out of date. I'm not a big fan of documenting this
level of details in an external document, it's bound to get of date
creating more work than necessary. We should have an adequate amount of
comments embedded in the code instead.

### DIFF
--- a/DESIGN.rst
+++ b/DESIGN.rst
@@ -207,31 +207,6 @@ All command-line commands should have a corresponding functional test.
 For example, the ``version`` command has a BATS_ functional test at
 `tests/functional/version.bats`_.
 
-Starting the hypervisor
------------------------
-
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

amshinde:
lgtm

grahamwhaley:
In message - s/get of date/get out of date/

I can see this has/will rot and is a finer level of detail that we probably don't need here - asking @jodh-intel , as I suspect he was the original author. But, given no strong complaints:

lgtm

jodh-intel:
That chunk referred to the 2.0 architecture and it is now out-of-date. I think I added it to help explain the design as the code appeared overly complex and there wasn't an obvious place to place that explanatory chunk.

I agree we should remove it.

lgtm.

sameo:
LGTM

PR #629, opened 2017-01-27
wip: Add support for running container with additional gids

amshinde:


chavafg:
qa-failed

src/spec_handlers/process.c
```diff
@@ -37,6 +37,9 @@ handle_user_section (GNode *root, struct cc_oci_config *config)
 	if (g_strcmp0(root->data, "gid") == 0) {
 		config->oci.process.user.gid = (gid_t)atoi (root->children->data);
 	}
+	if (g_strcmp0(root->data, "additionalGids") == 0) {
+		config->oci.process.user.additionalGids = node_to_strv (root);
```
devimc: probably you have to update ```cc_oci_config_free```

amshinde: Yeah..thanks for pointing that out. Fixing it.

tests/integration/docker/user.bats
```diff
@@ -45,3 +45,22 @@ setup() {
...
+	echo "${output}" | grep "nogroup"
+}
+
+@test "Run with non-existing numeric gid" {
+	skip "This is not working"
+	# When passed a non-existent numeric id, hyperstart errors out saying that the
+	# user does not exist. This works with runc.
+	# Docker mentions "When passing a numeric ID, the user does not have
+	# to exist in the container."
+	# This needs to be fixed in hyperstart codebase.
```
devimc: did you open the issue in hyperstart?

amshinde: Was about to open one, but saw just now that there is an issue created for this already : https://github.com/hyperhq/hyperstart/issues/250.
It pretty much covers what we are seeing here. Will add the issue here as well.

src/proxy.c
```diff
@@ -1383,10 +1384,7 @@ cc_proxy_run_hyper_new_container (struct cc_oci_config *config,
 	json_object_set_string_member (process, "workdir",
 			config->oci.process.cwd);
 
-	/* Hyperstart is not able to handle uid 0. This is a bug.
-	 * Simply dont sent the uid for now as the container
-	 * runs as root by default until the bug is resolved.
-	 */
+	// Hyperstart is not able to handle uid 0. This is a bug.
```
jodh-intel: For consistency -- and to keep the diff minimal -- please can you use the existing C-style comment.

src/proxy.c
```diff
@@ -1383,10 +1384,7 @@ cc_proxy_run_hyper_new_container (struct cc_oci_config *config,
...
 	if ((int)config->oci.process.user.uid != 0) {
```
jodh-intel: Note strictly part of this PR, but it would be *much* clearer to my mind if the logic was:

```
if ((int)config->oci.process.user.uid -= 0) {
    json_object_set_string_member (process, "user", "root");
} else {
   ...
}

```

src/oci.h
```diff
@@ -184,7 +184,7 @@ struct oci_cfg_root {
 struct oci_cfg_user {
 	uid_t    uid; /*!< User ID to run workload as. */
 	gid_t    gid; /*!< Group ID to run workload as. */
-	gid_t  **additionalGids; /*!< extra Group IDs to set workload as a member of */
+	gchar  **additionalGids; /*!< extra Group IDs to set workload as a member of */
```
jodh-intel: It would be worth adding a comment explaining why additional groups are stored as strings whereas the primary group is a numeric.

tests/integration/docker/user.bats
```diff
@@ -45,3 +45,22 @@ setup() {
 	[ "${status}" -eq 0 ]
 	[[ "${output}" == 'root'* ]]
 }
+
+@test "Run with additional groups" {
+	run $DOCKER_EXE run --rm --group-add audio --group-add nogroup busybox id
```
jodh-intel: What if these groups already exist?

tests/integration/docker/user.bats
```diff
@@ -45,3 +45,22 @@ setup() {
...
+	[ "${status}" -eq 0 ]
+	echo "${output}" | grep "audio"
```
jodh-intel: This test is quite loose. I'd prefer that we matched on an expected success message (`group foo added successfully`).

Also, I'd anchor the match:

```
echo "${output}" | egrep "blah blah.*\<audio\>.*blah"
```

tests/integration/docker/user.bats
amshinde: These groups are meant to exist. "--group-add" options is for running the container with these additional groups added to the user's groups.

src/oci.h
amshinde: I added in the commit message. Will add here too for clarity.

src/proxy.c
amshinde: Just converted this to a single line comment, hence chose the other style.

jodh-intel: Sure, I guessed as much, but the comment already existed. The point is, this is no longer the minimal change. 99% of the codebase uses C-style comments even for single-line comments, so new code should strive to honour the existing conventions... even if it is "only a comment". There are only 2 other C++ style comments in that file, and they relate to `FIXME`'s which by definition require that code (comment) to be rewritten anyway.

Aside: Ideally, this comment would also reference the hyperstart bug to help the reader, which might bump it into a multi-liner anyway.

src/proxy.c
```diff
@@ -1290,12 +1290,13 @@ cc_proxy_run_hyper_new_container (struct cc_oci_config *config,
 				  const char *container_id,
 				  const char *rootfs, const char *image)
 {
-	JsonObject *newcontainer_payload= NULL;
-	JsonObject *process = NULL;
-	JsonArray *args= NULL;
-	JsonArray *envs= NULL;
-	gchar     *uid_str = NULL;
-	gchar     *gid_str = NULL;
+	JsonObject *newcontainer_payload = NULL;
```
jodh-intel: You seem to have introduced some whitespace changes here too. It would be helpful if you could make such changes on a separate commit as again, this artificially make the diff bigger, and thus harder to read (bad!)

tests/integration/docker/user.bats
jodh-intel: Gotcha. In which case, it would be good to have a test that asserts the expected failure if a non-existent group is specified using `--group-add`.

amshinde: If the user provides a group name [not a gid], that is non-existent, docker itself handles that and throws this error "linux spec user: Unable to find group". If its a non-existent gid, then that is expected to be added by the runtime. Thats the reason I did not add test for the former scenario(non-existent group) but for the latter(non-existent gid).

amshinde:
@jodh-intel Pushed changes addressing all your review comments.

tests/integration/docker/user.bats
jodh-intel: ... in which case, a comment to that effect would have been useful, since the Docker behaviour appears inconsistent.

jodh-intel:
Minor doc nits remain, but insufficient to block landing this change.

lgtm.

gorozco1:
@amshinde any update on this ?

amshinde:
@gorozco1 I just added comments that jodh-intel had mentioned. However, this PR should land after this one is merged  : https://github.com/clearcontainers/hyperstart/pull/3.

sameo:
@amshinde @gorozco1 clearcontainers/hyperstart#3 is merged, we can merge that one as well.
LGTM.

gorozco1:
@sameo just merged

Issue #622, opened 2017-01-25
cc-proxy file descriptors leak


aduenasd:
Use of ‘lsof’ to track cc-proxy file descriptors.
The following are the fd after cc-proxy initialize and idle status:
```
root@clr-skylake ~ # lsof -p $(pidof cc-proxy)                                                                                                                                                                                                                 [0/1842]
COMMAND   PID USER   FD      TYPE             DEVICE SIZE/OFF     NODE NAME
cc-proxy 4216 root  cwd       DIR                8,3     4096        2 /
cc-proxy 4216 root  rtd       DIR                8,3     4096        2 /
cc-proxy 4216 root  txt       REG                8,3  8252484  6422860 /usr/libexec/cc-proxy
cc-proxy 4216 root  mem       REG                8,3    47616 11540156 /usr/lib64/libnss_files-2.24.so
cc-proxy 4216 root  mem       REG                8,3  1849048 11539827 /usr/lib64/libc-2.24.so
cc-proxy 4216 root  mem       REG                8,3   106072 11539927 /usr/lib64/libpthread-2.24.so
cc-proxy 4216 root  mem       REG                8,3    42824 10885123 /usr/lib64/libnss_altfiles.so.2
cc-proxy 4216 root  mem       REG                8,3  1263256 11539788 /usr/lib64/ld-2.24.so
cc-proxy 4216 root    0r      CHR                1,3      0t0        5 /dev/null
cc-proxy 4216 root    1u     unix 0xffff8d75f6078c00      0t0   478616 type=STREAM
cc-proxy 4216 root    2u     unix 0xffff8d75f6078c00      0t0   478616 type=STREAM
cc-proxy 4216 root    3u     unix 0xffff8d76869d1c00      0t0   477617 /var/run/cc-oci-runtime/proxy.sock type=STREAM
cc-proxy 4216 root    4u  a_inode               0,12        0    12078 [eventpoll]
```
Then let’s start a simple workload:
```
root@clr-skylake # docker run -ti clearlinux bash
```
And wait for the console to prompt.

Now ‘lsof’ shows new fd opened:
```
root@clr-skylake ~ # lsof -p $(pidof cc-proxy)
COMMAND   PID USER   FD      TYPE             DEVICE SIZE/OFF     NODE NAME
cc-proxy 4216 root  cwd       DIR                8,3     4096        2 /
cc-proxy 4216 root  rtd       DIR                8,3     4096        2 /
[...]

PR #628, opened 2017-01-27
tests: fix docker functional tests

devimc:
Before this patch some containers were not deleted in the test
where them were created causing random failures into the next tests
With this patch all container will be deleted where them were created

The main issue was this line:

```
"$DOCKER_EXE" ps -q | xargs -r "$DOCKER_EXE" kill
```

for example:
One container called 'tequila' is running.
When ```"$DOCKER_EXE" ps -q``` is executed the word 'tequila' is returned
but immediately 'tequila' dies ;( .
When ```xarg -r "$DOCKER_EXE" kill``` is executed docker tries
to kill to 'tequila' but it is already dead causing issue #593

"""Error response from daemon: Cannot kill container 'tequila'"""

RIP 'tequila'

fixes #593

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #631, opened 2017-01-30
proxy: Fix the VM lost detection site

dlespiau:
I placed the signal that we've lost the VM when the read() on the I/O
socket was failing. That was missing the rest of hyper to client I/O
copy loop: there are other places where having lost the VM will result
in this goroutine returning.

So, instead, place the VM lost signal at the end of the goroutine, we'll
signal we've lost the VM in all circumstances that way.

Fixes: https://github.com/01org/cc-oci-runtime/issues/622

### DIFF
--- a/proxy/vm.go
+++ b/proxy/vm.go
@@ -135,8 +135,6 @@ func (vm *vm) ioHyperToClients() {
 	for {
 		msg, err := vm.hyperHandler.ReadIoMessage()
 		if err != nil {
-			// VM process is gone
-			vm.signalVMLost()
 			break
 		}
 
@@ -158,6 +156,8 @@ func (vm *vm) ioHyperToClients() {
 		}
 	}
 
+	// Having an error on read/write is interpreted as having lost the VM.
+	vm.signalVMLost()
 	vm.wg.Done()
 }
 


chavafg:
qa-failed

chavafg:
got failures related to #593 which should be fixed by #628 
moving to qa-passed

chavafg:
qa-passed

jodh-intel:
The fix looks fine, but what about tests for this? We don't appear to have a `vm_test.go` but fwics it should be possible to test atleast some of the vm receivers?

dlespiau:
the coverage of vm.go comes from proxy_test.go and is at 61%. It's not ideal but that's better than nothing.

Trying to trigger the problem with unit tests is a bit delicate and certainly time consuming in this instance. Not sure I can find the time to do it any time soon.

jodh-intel:
How about we capture the need for a new test here in a fresh issue to avoid blocking landing this?

dlespiau:
https://github.com/clearcontainers/proxy/issues/7

jodh-intel:
Thanks @dlespiau.

lgtm.

gorozco1:
👍

Issue #638, opened 2017-02-02
qemu instances alive after 'docker rm -f'


aduenasd:
qemu instances are not being shutdown after 'docker rm -f'

root@clr-skylake # pidof qemu-system-x86_64
root@clr-skylake # docker run -tid clearlinux bash
7085933e09105204369356be0bbdf2681b22a782b566312d7f2b0b0e51ba331f
root@clr-skylake # docker rm -f $(docker ps -aq)
7085933e0910
root@clr-skylake # pidof qemu-system-x86_64
19956
root@clr-skylake  #

I noticed this at:
commit 7188b1c3e22dc4aec016635add6b1a4a6e62ae46
Merge: 2e94d20 19c4503
Author: James O. D. Hunt <james.o.hunt@intel.com>
Date:   Thu Feb 2 09:24:38 2017 +0000

    Merge pull request #613 from amshinde/check-race-support

    make: Add test-program to check -race is supported with go

dlespiau:
Hi,

Which docker version are you using for this test?

aduenasd:
root@clr-skylake ~ # docker version
Client:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.7.3
 Git commit:   7392c3b0ce0f9d3e918a321c66668c5d1ef4f689
 Built:        Wed Jan 11 05:25:37 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.7.3
 Git commit:   7392c3b0ce0f9d3e918a321c66668c5d1ef4f689
 Built:        Wed Jan 11 05:25:37 2017
 OS/Arch:      linux/amd64

devimc:
guilty 2b585d9c8a2b1295479faecddea2cd54df1c3a7f

running
```
for i in `seq 0 20`; do  docker run -dti clearlinux sh ; docker rm -f $(docker ps -aq) ; sleep 1; if pidof qemu-lite-system-x86_64; then break ; fi ; done
```
cc-proxy crashes

```
I0202 10:54:34.707926   24817 vm.go:114] [vm 83e7bddb hyperstart] pid 183 exit normally, status 0
I0202 10:54:34.707957   24817 vm.go:114] [vm 83e7bddb hyperstart] exec pid 185, pid 183
I0202 10:54:34.708003   24817 vm.go:114] [vm 83e7bddb hyperstart] can not find exec whose pid is 183
[...]

PR #639, opened 2017-02-02
proxy: Don't quit the hyper->client IO goroutine when a shim dies

dlespiau:
That's a bug we have had for the longest time. There is one such
goroutine per VM and we where quitting it whenever we had an error
writing data to a shim.

However, we may have a number of shim per-VM, one for the initial
container and one for each exec. We can't just quite the whole goroutine
if the write to one of the shim is failing.

This bug was discovered because we know signal "VM lost" when this
goroutine exits, which was causing havoc.

Fixes: https://github.com/01org/cc-oci-runtime/issues/638

### DIFF
--- a/proxy/vm.go
+++ b/proxy/vm.go
@@ -150,13 +150,15 @@ func (vm *vm) ioHyperToClients() {
 
 		err = hyperstart.SendIoMessageWithConn(session.client, msg)
 		if err != nil {
-			fmt.Fprintf(os.Stderr,
-				"error writing I/O data to client: %v\n", err)
-			break
+			// When the shim is forcefully killed, it's possible we
+			// still have data to write. Ignore errors for that case.
+			vm.infof(1, "io", "error writing I/O data to client:", err)
+			continue
 		}
 	}
 
-	// Having an error on read/write is interpreted as having lost the VM.
+	// Having an error on the IO channel read is interpreted as having lost
+	// the VM.
 	vm.signalVMLost()
 	vm.wg.Done()
 }


dlespiau:
If the reviewer is fully awake, he/she will notice we may have a tiny tiny here! Related to https://github.com/01org/cc-oci-runtime/issues/622. The case where it happens:

- One or several shims are running for that pod.
- One of the shim goes away, because it has been killed with SIGKILL
- The process in the VM the shim was shadowing still has data to send to that shim but the proxy can't write to the shim, it's gone
- We leak the session.client fd (which is the proxy end of the shim<->proxy socketpair())

Note that, in such occurrences, the VM is being destroyed soon after and vm.Close() will clean up those dangling I/O sessions.

So, this is a leak if the shim of an exec'ed process goes away while the VM process still has data for it. To make sure that's all true, it's time to bite the bullet and do a better job at unit testing this thing and will be for another PR as this one should fix a really annoying regression we have today.

chavafg:
qa-failed

devimc:
lgtm

chavafg:
unfortunately I am still seeing some qemu process after running docker integration tests:

```
All tests successful.
Files=23, Tests=52, 312 wallclock secs ( 0.06 usr  0.02 sys +  6.06 cusr  2.22 csys =  8.36 CPU)
Result: PASS
$ pidof qemu-lite-system-x86_64
8000 7536 2871
$ sudo docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

dlespiau:
@chavafg well, it could be another issue, some tests execute a sleep 30 and leave a container behind IIRC. Do you know if that's a regression or have we recently hardened the check?

proxy/vm.go
```diff
@@ -150,9 +150,10 @@ func (vm *vm) ioHyperToClients() {
 
 		err = hyperstart.SendIoMessageWithConn(session.client, msg)
 		if err != nil {
-			fmt.Fprintf(os.Stderr,
-				"error writing I/O data to client: %v\n", err)
-			break
+			// When the shim is forcefully killed, it's possible we
+			// still have data to write. Ignore errors for that case.
+			vm.infof(1, "io", "error writing I/O data to client:", err)
+			continue
```
jodh-intel: The change looks like it might lead to an infinite loop - is it guaranteed that `vm.hyperHandler.ReadIoMessage()` will eventually kick us out of the loop when it returns an error?

dlespiau:
I've been running the docker tests with this branch quite a bit this morning. I do notice qemu processes left behind, but there are also tests that fail.

qemu process can be left out when:
  - the runtime doesn't issue a destroypod command
  - destroypod doesn't work for some reason

With bug #640, the proxy was crashing. Because all commands go through the proxy, the runtime couldn't issue a destroypod command. With this PR, the proxy doesn't crash anymore.

At least I think bug #6378 (a regression introduced in 2b585d9 as @devimc says) is fixed. qemu processes are, I believe, caused by other bugs.

proxy/vm.go
dlespiau: vm.hyperHandler.ReadIoMessage() is a blocking read() on the hyperstart socket. It will error out with EOF when the qemu process dies.

If the qemu process doesn't go away, this goroutine doesn't either, but that's the intended behaviour: at the moment the proxy is only slave to the runtime/qemu and reacts to command/events from those.

proxy/vm.go
```diff
@@ -270,6 +272,21 @@ func (session *ioSession) Close() {
...
+	if session == nil {
+		vm.Unlock()
+		return
+	}
+	for i := 0; i < session.nStreams; i++ {
+		delete(vm.ioSessions, seq+uint64(i))
+	}
+	vm.Unlock()
+
+	session.Close()
```
jodh-intel: Missing check on returned error?

dlespiau:
I've also filed an issue which I think should help pinpoint where/why we leave processes behind: https://github.com/01org/cc-oci-runtime/issues/647

proxy/vm.go
dlespiau: there's no returned error here :) Actually, the whole chunk is spurious and should be in that commit. Let me update that.

dlespiau: ah, actually, when I rebased I also pushed a commit that shouldn't have been there. Fixed now.

jodh-intel:
lgtm.

chavafg:
moving to qa-passed, got some swarm failures unrelated to this.

chavafg:
qa-passed

Issue #640, opened 2017-02-02
proxy: fd leak if exec shim goes away unexpectedly 


dlespiau:
follow up of:
  - https://github.com/01org/cc-oci-runtime/issues/622
  - https://github.com/01org/cc-oci-runtime/issues/638
  - https://github.com/01org/cc-oci-runtime/pull/639

The case where it happens:

- One or several shims are running for that pod.
- One of the shim goes away, because it has been killed with SIGKILL
- The process in the VM the shim was shadowing still has data to send to that shim but the proxy can't write to the shim, it's gone
- We leak the session.client fd (which is the proxy end of the shim<->proxy socketpair())

Note that, in such occurrences, the VM is being destroyed soon after and vm.Close() will clean up those dangling I/O sessions.

So, this is a leak if the shim of an exec'ed process goes away while the VM process still has data for it.

PR #641, opened 2017-02-02
proxy: Close the connection when detecting the shim is gone

dlespiau:
See https://github.com/01org/cc-oci-runtime/issues/640 for details about the issue.

The main part of this PR is really the test coverage and being able to trigger that condition. I was expected it to be a bit tricky, but re-using my mock hyperstart and a similar setup than the proxy tests with the fd leak detector makes it manageable.

### DIFF
--- a/proxy/vm_test.go
+++ b/proxy/vm_test.go
@@ -1,0 +1,157 @@
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"net"
+	"os"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/containers/virtcontainers/hyperstart/mock"
+
+	hyper "github.com/hyperhq/runv/hyperstart/api/json"
+	"github.com/stretchr/testify/assert"
+)
+
+// vmRig maintains a test environment for vm objects
+type vmRig struct {
+	t  *testing.T
+	wg sync.WaitGroup
+
+	// hyperstart mocking
+	Hyperstart      *mock.Hyperstart
+	ctlPath, ioPath string
+
+	// fd leak detection
+	detector          *FdLeakDetector
+	startFds, stopFds *FdSnapshot
+}
+
+func newVMRig(t *testing.T) *vmRig {
+	return &vmRig{
+		t: t,
+	}
+}
+
+func (rig *vmRig) Start() {
+	var err error
+
+	rig.startFds, err = rig.detector.Snapshot()
+	assert.Nil(rig.t, err)
+
+	// Start hyperstart go routine
+	rig.Hyperstart = mock.NewHyperstart(rig.t)
+	rig.Hyperstart.Start()
+
+	// Explicitly send READY message from hyperstart mock
+	rig.wg.Add(1)
+	go func() {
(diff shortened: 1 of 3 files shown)


chavafg:
qa-failed

proxy/vm_test.go
```diff
@@ -0,0 +1,157 @@
...
+	shim.Close()
+
+	// Simulate the VM process sending data for a shim that is no longer there.
+	rig.Hyperstart.SendIoString(seq, "foo\n")
+
+	// Wait until the vm goroutine has read the IO data from hyperstart
+	// sent in the line above.
+	// Because that goroutine should cleanup vm.ioSessions on error, we
+	// use that condition for the wait. There isn't really an easier to
+	// synchronise here.
```
jodh-intel: I don't really understand this. If this really is the only approach we have for now though, don't we need a warning/error message plus call to `t.Fatal()` if this wait time failed to produced the expected outcome?

proxy/vm_test.go
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) 2016 Intel Corporation
```
jodh-intel: Yay for new tests! :-)

proxy/vm_test.go
dlespiau: Yes we do need an assert, for instance assert(t, 0, l). But I also think there's a better way to fix this (see comment).

dlespiau:
So, to be clear let's describe the part of the system in question:

we have 3 threads of execution:
  - a goroutine copying data from the VM to shims. There's one such goroutine per VM. That's `ioHyperToClients()`
  - a goroutine copying data from shim to VM. There's one such goroutine per shim. That's `ioClientToHyper()~
  - The test thread

The first goroutine does:

```
data, err :-= read(hyper_io_channel)
if err != nil {
    return
}

// finds the corresponding shim for the sequence number we have received
session := findSession() 

// session.client is the connection to the shim
err = write(session.client, data)
if err != nil {
   vm.closeSession(session) // will close session.client
   continue
}
```

So, in this PR, I'm handling the case where the shim is going away, session.client is a valid fd to a dead connection. What the test does is:

1. close the other end of session.client to simulate a dead shim 
2. make the mock hyperstart send some data for that shim
3. *some time later* `read(hyper_io_channel)` unblocks with the data from the mock hyperstart
4.  write(session.client, data) returns an error, we can't write to the shim, and go through the code handling that error

What the wait does: it waits for the ioHyperToClients() goroutine to be scheduled and read data from the mock hyperstart through the side effect of the code run when handling the write(session.client) error.

And we need to wait for that to happen or we'll have a race between receiving the hyperstart message that unblocks `read(hyper_io_channel)` and the code cleanup the proxy<->hyperstart connection (done in the test thread) which will also unblock ``read(hyper_io_channel)` but with an EOF error and so the code path we wanted to exercise isn't (the error handling of `read(hyper_io_channel)` terminates the goroutine.

dlespiau:
Ok, that we have more context, there's a better way to do that. If the shim dies, we have another read() that will return an EOF, the read() in the goroutine copying data from the shim to hyperstart. This one will always produce an error if the shim goes away, while the one I touched will only produce an error if hyperstart sends data to the dead shim.

We should either close() the connection to the dead shim:
  - in the per-shim goroutine
  - both in the per-shim goroutine and in the erorr path fixed in this PR. Even if we close it twice (the second close will return an error, but we can ignore it.).

proxy/vm_test.go
jodh-intel: I'm still confused - this is test code, so you can assert the behaviour, so why not? The comment says "`Because that goroutine should cleanup vm.ioSessions on error...`" but what if that behaviour doesn't happen?

All I'm suggesting is that you add a check to ensure that `len(vm.ioSessions) == 0` when that loop has exited and error if not since it appears that that block could fail, but it would be a silent failure.

jodh-intel:
This branch now needs a rebase.

dlespiau:
I still need to revisit the fix to be more correct as well, will do after some more 3.0 work in proxy so I can unblock more of 3.0.

jodh-intel:
@dlespiau - can you give an update on this PR please?

jodh-intel:
@dlespiau - can we close this?

dlespiau:
Sure, we have the issue to track it anway.

PR #642, opened 2017-02-02
Test: Delete function from the setup that is not longer used and add …

GabyCT:
…--rm flag

- This will delete the use of the function of clean_docker_ps in the setup as
this one is no longer used in our test-common.bash
- Add the flag of --rm when running a container

### DIFF
--- a/tests/integration/docker_images/popular-containers-images.bats
+++ b/tests/integration/docker_images/popular-containers-images.bats
@@ -25,7 +25,6 @@ SRC="${BATS_TEST_DIRNAME}/../../lib/"
 
 setup() {
 	source $SRC/test-common.bash
-	clean_docker_ps
 	runtime_docker
 }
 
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

devimc:
lgtm

chavafg:
still having some issues with docker swarm, not related to this change, moving to qa-passed

chavafg:
qa-passed

jodh-intel:
HI @GabyCT - good catch. It looks like c8b933adb193d611fc0998368f866d8028efe99a removed the `clean_docker_ps` function. But there still other files calling this function:

```
$ grep -rlE "\<clean_docker_ps\>" | wc -l
7
```

GabyCT:
@jodh-intel, thanks for the observation. I noticed that the other files that are calling the function are part of the metrics as the integration docker tests are now fixed.

tests/integration/docker_images/popular-containers-images.bats
```diff
@@ -289,7 +290,7 @@ setup() {
 }
 
 @test "start instance in percona container" {
-	if $DOCKER_EXE run -i percona perl -e 'print "Clear Containers\n"' | grep LANG; then false; else true; fi
+	if $DOCKER_EXE run --rm -i percona perl -e 'print "Clear Containers\n"' | grep LANG; then false; else true; fi
```
dlespiau: This construct "if xxx | grep foo; then false; else true; fi" could have been done with "if xxx | grep -v foo" (no need to do anything about this, for next time you're around that code).

dlespiau:
lgtm, but those remaining occurrences of clean_docker_ps will have to be removed. Filed https://github.com/01org/cc-oci-runtime/issues/652.

Issue #647, opened 2017-02-03
Fail a test if a container, cc-shim or qemu is left behind


dlespiau:
We need to tighten a bit our tests by failing if something unexpected happens to better understand where the bugs are. It's not enough to say "at the end of the test run, we have processes left behind".

Tests need to ensure:

- setup(): no qemu(-lite) process is running on the system
- setup(): no cc-shim is running on the system
- setup(): there's no running container on the system

- teardown(): no qemu(-lite) process is running on the system
- teardown(): no cc-shim is running on the system
- teardown(): there's no running container on the system

at teardown(), if we detect a problem, we fail the test. That way we'll know which test failed. We should then try to clean up the environment for the next test (kill container, qemu, cc-shim).

jodh-intel:
Agreed, although we can tighten up the checks further:

* "`function()`: no `$thing` is running on the system *as the user running the tests*".

PR #648, opened 2017-02-03
tests: Use the number of running tests in info.bats

dlespiau:
Since:

  commit c8b933adb193d611fc0998368f866d8028efe99a
  Author: Julio Montes <julio.montes@intel.com>
  Date:   Thu Jan 26 18:16:44 2017 -0600

      tests: fix docker functional tests

We don't run docker rm in setup(), which is a good thing and avoid hiding bugs.

However it's possible the system we are running on still has some
(stopped) containers not yet deleted. It's slightly more robust to use
the number of running containers for the check we do here.

### DIFF
--- a/tests/integration/docker/info.bats
+++ b/tests/integration/docker/info.bats
@@ -31,6 +31,12 @@ setup() {
 @test "Container info" {
 	container=$(random_name)
 	$DOCKER_EXE run -itd --name $container busybox
-	$DOCKER_EXE info| grep "Containers: 1"
+	# Checks one container is running (relies on having no previous docker
+	# container running).
+	$DOCKER_EXE info| grep "^ Running: 1$"
+	# As an additional measure, make sure $container is indeed running
+	run $DOCKER_EXE inspect --type container --format '{{ .State.Status }}' $container
+	[ "$status" -eq 0 ]
+	[ "$output" = "running" ]
 	$DOCKER_EXE rm -f $container
 }


tests/integration/docker/info.bats
```diff
@@ -31,6 +31,6 @@ setup() {
 @test "Container info" {
 	container=$(random_name)
 	$DOCKER_EXE run -itd --name $container busybox
-	$DOCKER_EXE info| grep "Containers: 1"
+	$DOCKER_EXE info| grep "^ Running: 1"
```
jodh-intel: This needs anchoring:

```
$DOCKER_EXE info| grep "^ Running: 1$"
```

dlespiau: it does!

jodh-intel: at the end (see the dollar ;)

dlespiau: I meant it does need anchoring!

jodh-intel: ah! :smiley:

chavafg:
qa-failed

tests/integration/docker/info.bats
chavafg: what about using a regex instead of a fixed number, so in case we have other containers running, the test does not fail.

jodh-intel: I was thinking the same, but I don't think we can reliably test this since `info` only gives a count so it's all going to be horribly racy.

dlespiau: If don't check anything if we just check for a number. We can actually do much better, just ask docker for the status of $container:

```
$ docker inspect --type container --format '{{ .State.Status }}' $container
running
```

dlespiau:
Updated PR using docker inspect and template strings to get the status of the created container.

tests/integration/docker/info.bats
```diff
@@ -31,6 +31,8 @@ setup() {
 @test "Container info" {
 	container=$(random_name)
 	$DOCKER_EXE run -itd --name $container busybox
-	$DOCKER_EXE info| grep "Containers: 1"
+	run $DOCKER_EXE inspect --type container --format '{{ .State.Status }}' $container
+	[ "$status" -eq 0 ]
+	[ "$output" = "running" ]
```
jodh-intel: Nice idea, but this test is supposed to be testing `docker info` ;-)

Maybe add back in the call to `docker info` and keep the extra tests so we have some degree of assurance that the container state we're checking in `info` relates to the one we created?

dlespiau: argh! ah well.

dlespiau:
Ok, Updated things again, checking both the number of running containers and that the created container is indeed running.

Of course, as said by everyone on this PR, this relies on having no docker container running when starting this test, that's a slightly better constraint than today (having no stopped container either) and should be made better by https://github.com/01org/cc-oci-runtime/issues/647

amshinde:
lgtm

jodh-intel:
lgtm

dlespiau:
@chavafg can you confirm this last version info info.bats passes in the CI environment?

chavafg:
info.bats passes

chavafg:
qa-passed

PR #653, opened 2017-02-07
Swarm tests fixes 

chavafg:
This PR fixes:
- replicas obtained from non-running containers:
    The swarm tests are verifying the interfaces and IP addresses
    from the containers listed with 'docker ps -a -q', but this fails
    if there is a container that is not running (which maybe was created
    before running the tests).
    This commit changes the command to 'docker ps -q' so that we only
    get the list of containers that are running.

- a typo in swarm.bats
- indentation issues.

### DIFF
--- a/tests/integration/docker/swarm.bats
+++ b/tests/integration/docker/swarm.bats
@@ -119,7 +119,7 @@ setup() {
 }
 
 @test "check that replicas has two interfaces" {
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

amshinde:
lgtm

dlespiau:
Nitpick, for next time: white space changes are usually put in a separate commit to ease the review.

lgtm

PR #655, opened 2017-02-07
Add memory consumption tool

MarioCarrilloA:
This tool allows to measure the average memory
consumption of a Clear Containers component during
a determined Clear Containers stress test.

The parameters that modify the stress level are:
* number of clear containers.
* workload for all instances.
can be customized.

### DIFF
--- a/tests/metrics/collect_mem_consmd.sh
+++ b/tests/metrics/collect_mem_consmd.sh
@@ -1,0 +1,302 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to measure the average memory consumption
+#  of a Clear Containers component during a determined Clear Containers
+#  stress flow.
+#
+#  Note: It is recommended to launch the same number of Clear Containers per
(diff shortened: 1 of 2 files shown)


gorozco1:
lgtm

tests/metrics/collect_mem_consmd.sh
```diff
@@ -0,0 +1,299 @@
...
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to measure the average memory consumption
+#  of a Clear Containers component during a determined Clear Containers
+#  stress flow.
+
+# set -x
+# set -e
```
dlespiau: leaving set -e always on seems like a good thing to me so we can catch errors that could be silent otherwise.

tests/metrics/collect_mem_consmd.sh
```diff
@@ -0,0 +1,299 @@
...
+EOF
+)
+}
+
+# This function launch collectl tool
+# in order to get metrics about a specific tool.
+function start_collect()
+{
+	collectl -oTm -sZm -i1:1 \
+		--procfilt f$PROC \
```
dlespiau: We have one cc-shim process per container. When PROC=cc-shim, are all the cc-shim processes monitored? if so that means the results is actually the total memory consumed by all the cc-shim processes, not just one. Maybe we should tell that to the user and/or divide the result by the number of shims?

MarioCarrilloA: The result is an average in KB

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: Done :)

tests/metrics/collect_mem_consmd.sh
dlespiau: Sure, but is the result the sum of all cc-shim running on the system, just one?

MarioCarrilloA: The results is the sum of all

dlespiau: I guess that's fine, but that also means we can't compare cc-shim numbers directly between two runs launching a different number of containers. We'd need to divide the result by the number of shims for that.

We'll need to keep that in mind, could you at least put a comment in the code?

MarioCarrilloA: yes, the best way is to compare 2 executions launching the same number of CC, I will add a comment in the script description

MarioCarrilloA: @dlespiau Done

dlespiau:
Alright:

lgtm

Issue #154, opened 2016-08-23
shim errors appeared when running several containers 


chavafg:
When trying to launch100 containers, I got several of these errors:
Also the docker-cor.service went to failure state. 
[docker-journal.txt](https://github.com/01org/cc-oci-runtime/files/431621/docker-journal.txt)
[cc-oci-runtime.log.txt](https://github.com/01org/cc-oci-runtime/files/431623/cc-oci-runtime.log.txt)

```
6db699944d5c106a580ba9ed20081c5f3c3626e763fa4f10df6785628562d642
docker: Error response from daemon: shim error: open pid: no such file or directory.
01fd7a382ca1b867cd654fb9c5742c274df22072c7f2c664dd0d8a9adc805f2b
docker: Error response from daemon: fork/exec /usr/bin/cc-oci-runtime.sh: resource temporarily unavailable: "".
```

Taking a look at docker-cor service, I got these errors regarding to container 6db69994:  
```
Aug 23 02:44:30 clr-containers dockerd[1413]: time="2016-08-23T02:44:30.510237936Z" level=debug msg="Calling POST /v1.24/containers/6db699944d5c106a580ba9ed20081c5f3c3626e763fa4f10df6785628562d642/start"
Aug 23 02:44:30 clr-containers dockerd[1413]: time="2016-08-23T02:44:30.510940554Z" level=debug msg="devmapper: activateDeviceIfNeeded(cf80ef447e9ae005edad0f40997b2cd6c78484b6a90819ba1c2d6d6e9c00a2b3)"
Aug 23 02:44:30 clr-containers dockerd[1413]: time="2016-08-23T02:44:30.544955238Z" level=debug msg="container mounted via layerStore: /var/lib/docker/devicemapper/mnt/cf80ef447e9ae005edad0f40997b2cd6c78484b6a90819ba1c2d6d6e9c00a2b3/rootfs"
Aug 23 02:44:30 clr-containers dockerd[1413]: time="2016-08-23T02:44:30.545266939Z" level=debug msg="Assigning addresses for endpoint compassionate_einstein's interface on network bridge"
Aug 23 02:44:30 clr-containers dockerd[1413]: time="2016-08-23T02:44:30.545302902Z" level=debug msg="RequestAddress(LocalDefault/172.17.0.0/16, <nil>, map[])"
[...]

PR #682, opened 2017-02-17
documentation: add fedora installation details

egernst:
Signed-off-by: Eric Ernst <eric.ernst@intel.com>

### DIFF
--- a/documentation/Installing-Clear-Containers-on-Fedora-25.md
+++ b/documentation/Installing-Clear-Containers-on-Fedora-25.md
@@ -1,0 +1,76 @@
+#Install the Clear Containers runtime on Fedora 25
+
+##Install the Clear Containers runtime
+```
+$ sudo -E dnf config-manager --add-repo \
+http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo
+$ sudo dnf install cc-oci-runtime linux-container
+```
+
+##Install docker 1.12
+
+For compatibility reasons, docker 1.12.1 is required. This is not available
+in Fedora 25, so we will pull from the Fedora 24 repository:
+
+```
+$ sudo dnf config-manager --add-repo \
+	https://yum.dockerproject.org/repo/main/fedora/24
+```
+Append gpg key information to end of docker's fedora 24 repo config:
+```
+$ sudo vi /etc/yum.repos.d/yum.dockerproject.org_repo_main_fedora_24.repo
+gpgcheck=1
+gpgkey=https://yum.dockerproject.org/gpg
+```
+Install docker engine:
+```
+$ sudo dnf install docker-engine-1.12.1-1.fc24
+```
+
+##Configure docker to use Clear Containers by default
+```
+$ sudo mkdir -p /etc/systemd/system/docker.service.d/
+$ sudo vi /etc/systemd/system/docker.service.d/clr-containers.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd -D --add-runtime cor=/usr/bin/cc-oci-runtime --default-runtime=cor
+```
+##Restart the docker systemd service
+```
+$ sudo systemctl daemon-reload
+$ sudo systemctl restart docker
+```
+
+##Run a Clear Container
+You are now ready to run Clear Containers. For example
+```
+$ sudo docker run -ti fedora bash
+```
+##Notes
+You can override the default runtime docker will use by specifying
+`--runtime $runtime` (`$runtime` can be either `cor` or `runc` (dockers
+default)). For example, to specify explicitly that you wish to use the
+Clear Containers runtime:
+```
+$ sudo docker run --runtime cor -ti ubuntu bash
+```
+Note: this should not be necessary if you update `--default-runtime` setting in `clr-containers.conf` as described above.
+
+##Increase limits (optional)
+If you are running a recent enough kernel (4.3+), you should consider
+increasing the `TasksMax=` systemd setting. Without this, the number of
+Clear Containers you are able to run will be limited.  A sample of the error which may manifest can be seen at [issue 154] (https://github.com/01org/cc-oci-runtime/issues/154).
+
+
+Run the commands below and if they display `OK`, proceed, else skip this
+section
+```
+$ grep -q ^pids /proc/cgroups && echo OK
+```
+Assuming you got an `OK`
+```
+$ cd /etc/systemd/system/docker.service.d/
+$ sudo vi /etc/systemd/system/docker.service.d/clr-containers.conf
+[Service] # Allow maximum number of containers to run.
+TasksMax=infinity
+```


documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,60 @@
...
+##Install the Clear Containers runtime
+```
+$ sudo -E dnf config-manager --add-repo
+http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo
+$ sudo dnf install cc-oci-runtime linux-container
+```
+
+##Install docker 1.12
+```
+$ sudo -E dnf config-manager --add-repo https://yum.dockerproject.org/repo/testing/fedora/25
```
gorozco1: According to documentation of docker: https://docs.docker.com/engine/installation/linux/fedora/

This is the URL of the repo, is there any reason to add the testing repo?

2. sudo dnf config-manager \
    --add-repo \
    https://docs.docker.com/engine/installation/linux/repo_files/fedora/docker.repo

egernst: I simply pulled the installation directions as is which were used in wiki before (and have been used/verified).  Having said that, I can verify and update this.

gorozco1: Oh, I just updated the wiki too

chavafg:
qa-passed

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,60 @@
...
+
+##Run a Clear Container
+You are now ready to run Clear Containers. For example
+```
+$ sudo docker run -ti fedora bash
+```
+##Notes
+You can override the default runtime docker will use by specifying
+`--runtime $runtime` (`$runtime` can be either `cor` or `runc` (dockers
+default)). For example, to specify explicitly that you wish to use the
```
jodh-intel: This is slightly unclear. You are correct that `runc` is dockers default, but what isn't clear is that the command above has now changed the default to `cor` making `--runtime cor` optional.

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,60 @@
...
+Clear Containers runtime
+```
+$ sudo docker run --runtime cor -ti ubuntu bash
+```
+##Increase limits (optional)
+If you are running a recent enough kernel (4.3+), you should consider
+increasing the `TasksMax=` systemd setting. Without this, the number of
+Clear Containers you are able to run will be limited.
```
jodh-intel: It might be a good idea to reference #154 here since the error you get if you *don't* specify `TasksMax` is not very helpful.

documentation/Installing-Clear-Containers-on-Fedora-25.md
egernst: Sure -- I was just pulling in explicitly what exists on the wiki today.  I can clarify that this isn't necessary if you updated the config...

documentation/Installing-Clear-Containers-on-Fedora-25.md
egernst: Okay

egernst:
Updates made to take James and  Geronimo's feedback into account.

Perhaps I misunderstood, but should we also have a pull to remove the wiki pages, so we aren't just duplicating data?

jodh-intel:
lgtm

grahamwhaley:
@egernst we don't currently drive the wiki through the git/PR flow - we just go edit it freeform. I suggest we land this PR first, and then we can remove any outdated wiki pages by hand

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,63 @@
...
+$ sudo -E dnf config-manager --add-repo
+http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo
+$ sudo dnf install cc-oci-runtime linux-container
+```
+
+##Install docker 1.12
+```
+$ sudo dnf config-manager --add-repo \
+	https://docs.docker.com/engine/installation/linux/repo_files/fedora/docker.repo
+$ sudo dnf install docker-engine-1.12.1-1.fc25
```
grahamwhaley: Just to check @egernst - is this 1.12.1 for fc25 still valid? - I had a browse of the repo and did not see it - and #687 implies that maybe it is not... I'm not set up to easily try this - anybody else verify??

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,63 @@
...
+```
+##Notes
+You can override the default runtime docker will use by specifying
+`--runtime $runtime` (`$runtime` can be either `cor` or `runc` (dockers
+default)). For example, to specify explicitly that you wish to use the
+Clear Containers runtime:
+```
+$ sudo docker run --runtime cor -ti ubuntu bash
+```
+Note: this should not be necessary if you update --default-runtime setting in clr-containers.conf as described above.
```
grahamwhaley: probably code quote the arguments like `--default-runtime` and maybe `clr-containers.conf` as well?

grahamwhaley:
@egernst generally looking pretty good. I like the relative simplicity of it :-)  Couple of style notes and a query about 1.12 availability on f25 - will wait for answers etc...

documentation/Installing-Clear-Containers-on-Fedora-25.md
egernst: Indeed it is not -- I am pushing an update to use the fc24 docker engine instead.

documentation/Installing-Clear-Containers-on-Fedora-25.md
egernst: Sure.

egernst:
Updates made based on feedback from Graham.  Also updated to grab docker engine from FC24 repo instead (tested booting using our released cc-oci-runtime).

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,63 @@
+#Install the Clear Containers runtime on Fedora 25
+
+##Install the Clear Containers runtime
+```
+$ sudo -E dnf config-manager --add-repo
```
chavafg: backslash missing at the end of the line

jodh-intel:
lgtm

amshinde:
lgtm

gorozco1:
With this do we fix issue #687 ? @chavafg

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,63 @@
...
+$ sudo -E dnf config-manager --add-repo \
+http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo
+$ sudo dnf install cc-oci-runtime linux-container
+```
+
+##Install docker 1.12
+```
+$ sudo dnf config-manager --add-repo \
+	https://yum.dockerproject.org/repo/main/fedora/24
+$ sudo dnf install docker-engine-1.12.1-1.fc24
```
gorozco1: Could you verify the changes addresses and fixes #687

chavafg: We need to add the step documented here, before installing `docker-engine`:
https://github.com/01org/cc-oci-runtime/issues/687#issuecomment-283209057

so we do not get this error: 

`Error: Public key for docker-engine-1.12.1-1.fc24.x86_64.rpm is not installed`

egernst: Done

egernst:
Added GPG key handling instructions.

chavafg:
qa-passed

gorozco1:
lgtm

documentation/Installing-Clear-Containers-on-Fedora-25.md
```diff
@@ -0,0 +1,72 @@
...
+```
+$ sudo -E dnf config-manager --add-repo \
+http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo
+$ sudo dnf install cc-oci-runtime linux-container
+```
+
+##Install docker 1.12
+```
+$ sudo dnf config-manager --add-repo \
+	https://yum.dockerproject.org/repo/main/fedora/24
```
jodh-intel: I think we need a clear comment explaining that, yes, we really do mean Fedora 24 here and everywhere else in the doc.

Without that, it's not unreasonable for the user to think that we only partially updated this doc from F24->F25 and the user may then disregard all mention of F24, go with F25 packages and have a bad outcome.

gorozco1: We have to add an statement that says we are going to install docker and its dependencies from a fedora 24 repo, because compatibility.

dlespiau: @egernst So, it seems we are just a little precision away from merging this PR. Mind taking another quick look at it?

egernst: @dlespiau -- thanks for the poke.  Update made to take jodh's feedback into account.

dlespiau:
lgtm

rcaballeromx:
Please make sure that the steps are easy to follow, even and especially, if you have never done them before. The steps should be numbered if they need to be performed in a specific order. Optional or alternative steps should be added as notices. The way this content is written, the steps are headings which is not correct.  The best practice is to have a step for every independent action. Simple actions, such as creating a directory and moving inside, may not require their own step.
Please get in contact with me if you have any questions regarding this information and to discuss what would be your preference regarding me supplying comments on specific pieces of the content.

dlespiau:
@rcaballeromx Those are good points. We have a few submissions around installation instructions pending and I think it'd be quite hard to converge towards a good and coherent whole without seeing all of those pieces together.

Any objection if we merge them, and then, have another pass with a more holistic view than just the fedora installation instructions?

rcaballeromx:
@dlespiau I have commented along the same lines on the other submissions since they have the same problem. Each document should have its own set of steps and contain all the steps needed to complete the task. If that is not the case, as you hint, we have a much larger problem. Readers get frustrated if, after following a series of steps, they discover that the task is not done and have to move to the next set of steps to complete it.
Generally, I treat documentation exactly the same as I would code. Therefore, I would not merge this content, as is, since it does not follow best practices, does not comply with industry standards, and does not use the right markup. Yes, even in MD there are different markups needed for different things.
If there is urgent need to include the content, we could compromise and merge it but only if we have a specific date in the near future by which the contributor would perform the second pass you describe to fix it.

chavafg:
qa-passed

grahamwhaley:
We have a need to release, as the code release these documents enable installation of was done a few weeks ago and we'd like to voice the availability further.
I would go with the solution of opening a new Issue to note 'docs need cleanup' (please word it better than that and with more details :-), and **as long as these instructions work**, merge them.
I think we have to take the pragmatic 'does it work' choice over the 'is it perfectly pretty yet'. We do that with code sometimes as well... when necessary.

dlespiau:
@rcaballeromx I'm afraid it's not just about the pending PRs, we already have some installation instructions in-tree that follow the same model, eg. https://github.com/01org/cc-oci-runtime/blob/master/documentation/Installing-Clear-Containers-on-Ubuntu.md

So, a staged approach is appealing to me:
  - We don't delay the crux of the information too much longer, It's been 24 days already the current state does have some useful information.
  - We don't have an incoherent state where some instructions follow your recommendations and others don't,
  - I've opened issue https://github.com/01org/cc-oci-runtime/issues/739 to track the changes you suggest across all the installation instructions we have and assigned it to @GabyCT. That's the best I can do, a completion date is hard to commit to.

dlespiau:
@rcaballeromx Thanks, that's much appreciated. Merging then.

PR #683, opened 2017-02-20
Misc fixes

jodh-intel:


grahamwhaley:
Found with your new linter tool per chance?
lgtm

src/process.c
```diff
@@ -226,9 +226,7 @@ cc_oci_setup_shim (struct cc_oci_config *config,
 	ret = true;
 
 out:
-	if (tty_fd > 2) {
```
dlespiau: This seems to have been done on purpose in commit f1c21527e9056c3f629baaa98849c67edd866cbb

chavafg:
qa-failed

src/process.c
jamesodhunt: Indeed, but I don't think that is correct behaviour. I've just updated my test which should guarantee the fd gets closed in all scenarios.

@amshinde - could you review this?

grahamwhaley: /me admits it rang a little bell when he read it... iirc, this was so we do not close the host stdin/stdout/stderr iirc - and if so, I will nag that there was not enough comments in either the commit message nor the code... @amshinde - can you remember ? (and if so, maybe submit a new PR clarifying in the code etc. etc.)

dlespiau: I think the problem was: what happens if tty_fd is, say, 2? (and the fact that @amshinde had to introduce that case makes me think it's not a theoretical question).

- open(config->console, O_RDWR |  O_NOCTTY) returns 2
- dup2() on 0 and 1 work as usual, closing the previous 0 and 1
- dup2(tty_fd, 2) does nothing, from the man page:

```
    If  oldfd  is  a  valid  file descriptor, and newfd has the same value as oldfd, then dup2()
    does nothing, and returns newfd
```
close(tty_fd) will close fd 2, and we don't want to do that.

jodh-intel: Good catch! Test updated.

dlespiau:
lgtm

src/process.c
amshinde: lgtm. Yes tty_fd has been found to have values  <=2.

amshinde:
lgtm

jodh-intel:
Hi @chavafg - could you respin the tests for this PR?

chavafg:
qa-passed

devimc:
lgtm

PR #688, opened 2017-02-21
docs: Add Clear Linux and Centos 7 installation instructions

chavafg:
This commit adds instructions on how to install and run
Clear Containers 2.1 on Clear Linux and Centos 7.

### DIFF
--- a/documentation/Installing-Clear-Containers-on-Centos-7.md
+++ b/documentation/Installing-Clear-Containers-on-Centos-7.md
@@ -1,0 +1,203 @@
+# Install the Clear Containers runtime on Centos 7
+
+
+## Introduction
+
+Installing Clear Containers on CentOS is slightly more difficult than on other distributions mainly due to the fact that it does not provide current versions of gcc and glibc.
+
+Hence, this guide explains how to install updated packages in `/usr/local`, to avoid conflicting with the base packages.
+
+## Warning
+
+Note carefully the comment in the section [Install container kernel and rootfs](#install-container-kernel-and-rootfs). You MUST update the packages listed regularly to ensure your Clear Containers benefit from performance improvements, new features and most importantly CVE security fixes.
+
+## Install updated glib and json-glib
+
+Note: these packages need to be built with the default CentOS compiler since they don't build by default with gcc 6.
+
+```
+$ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:$PKG_CONFIG_PATH
+$ sudo yum groupinstall 'Development Tools'
+$ sudo yum install zlib-devel libffi-devel
+
+$ curl -L -O "https://download.gnome.org/sources/glib/2.46/glib-2.46.2.tar.xz"
+$ tar -xvf glib-2.46.2.tar.xz
+$ pushd glib-2.46.2
+$ ./configure --prefix=/usr/local --disable-silent-rules && make -j5 && sudo make install
+$ popd
+
+$ curl -L -O "https://download.gnome.org/sources/json-glib/1.2/json-glib-1.2.2.tar.xz"
+$ tar -xvf json-glib-1.2.2.tar.xz
+$ pushd json-glib-1.2.2
+$ ./configure --prefix=/usr/local --disable-silent-rules && make -j5 && sudo make install
+$ popd
+
+```
+
+## Install updated gcc
+
+This is required for `qemu-lite`.
+
+```
+$ sudo yum install gmp-devel mpfr-devel libmpc-devel gcc-c++
+$ curl -L -O http://www.mirrorservice.org/sites/ftp.gnu.org/gnu/gcc/gcc-6.2.0/gcc-6.2.0.tar.gz
+$ tar xvf gcc-6.2.0.tar.gz
+$ pushd gcc-6.2.0
+$ ./configure --enable-languages=c --disable-multilib --prefix=/usr/local/gcc-6.2.0
+$ make -j5 && sudo make install
+$ popd
+```
+
+## Install qemu-lite
+
+Although this package installs the `qemu-lite` version of `qemu`, the binary will still be called `qemu-system-x86_64`.
+Requires at least gcc 5.
+
+```
+$ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:$PKG_CONFIG_PATH
+$ export PATH=/usr/local/gcc-6.2.0/bin:$PATH
+$ sudo yum install libcap-ng-devel pixman-devel libattr-devel libcap-devel
+$ curl -L -O https://github.com/01org/qemu-lite/archive/3f889234fcd31ac4119aae57559b1286e09e1568.tar.gz
+$ tar xvf 3f889234fcd31ac4119aae57559b1286e09e1568.tar.gz
+$ pushd qemu-lite-3f889234fcd31ac4119aae57559b1286e09e1568
+$ CC=/usr/local/gcc-6.2.0/bin/gcc ./configure \
+				--disable-bluez 	--disable-brlapi \
+				--disable-bzip2 	--disable-curl \
+				--disable-curses 	--disable-debug-tcg \
+				--disable-fdt 		--disable-glusterfs \
+				--disable-gtk 		--disable-libiscsi \
+				--disable-libnfs 	--disable-libssh2 \
+				--disable-libusb 	--disable-linux-aio \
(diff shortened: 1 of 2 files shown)


chavafg:
qa-passed

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,171 @@
+# Install the Clear Containers runtime on Centos 7
+
+
+## Introduction
+
+Installing Clear Containers on CentOS is slightly more difficult than on other distributions mainly due to the fact that it does not provide current versions of gcc and glibc.
+
+Hence, this guide explains how to install updated packages in /usr/local, to avoid conflicting with the base package
```
jodh-intel: - `/usr/local`.
- "to avoid conflicting with the base packages."

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,171 @@
...
+$ sudo yum install libcap-ng-devel pixman-devel libattr-devel libcap-devel
+$ curl -L -O https://github.com/01org/qemu-lite/archive/3f889234fcd31ac4119aae57559b1286e09e1568.tar.gz
+$ tar xvf 3f889234fcd31ac4119aae57559b1286e09e1568.tar.gz
+$ pushd qemu-lite-3f889234fcd31ac4119aae57559b1286e09e1568
+$ CC=gcc ./configure --disable-bluez --disable-brlapi --disable-bzip2 --disable-curl --disable-curses --disable-debug-tcg --disable-fdt --disable-glusterfs --disable-gtk --disable-libiscsi --disable-libnfs --disable-libssh2 --disable-libusb --disable-linux-aio --disable-lzo --disable-opengl --disable-qom-cast-debug --disable-rbd --disable-rdma --disable-sdl --disable-seccomp --disable-slirp --disable-snappy --disable-spice --disable-strip --disable-tcg-interpreter --disable-tcmalloc --disable-tools --disable-tpm --disable-usb-redir --disable-uuid --disable-vnc --disable-vnc-{jpeg,png,sasl} --disable-vte --disable-xen --enable-attr --enable-cap-ng --enable-kvm --enable-virtfs --target-list=x86_64-softmmu --extra-cflags="-fno-semantic-interposition -O3 -falign-functions=32" --datadir=/usr/local/share/qemu-lite --libdir=/usr/local/lib64/qemu-lite --libexecdir=/usr/local/libexec/qemu-lite
```
jodh-intel: That initial `CC=gcc` can be dropped I think as we've already set `$PATH` above to ensure gcc 6 is used.

Alternatively, to make it 100% clear, you could specify `CC=/usr/local/gcc-6.2.0/bin/gcc ./configure ...`.

jodh-intel: I wonder if we should fold this very long line using backslashes; it makes my eyes bleed just a little ;-)

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,171 @@
...
+$ sudo yum install gmp-devel mpfr-devel libmpc-devel gcc-c++
+$ curl -L -O http://www.mirrorservice.org/sites/ftp.gnu.org/gnu/gcc/gcc-6.2.0/gcc-6.2.0.tar.gz
+$ tar xvf gcc-6.2.0.tar.gz
+$ pushd gcc-6.2.0
+$ ./configure --enable-languages=c --disable-multilib --prefix=/usr/local/gcc-6.2.0
+$ make -j5 && sudo make install
+$ popd
+```
+
+## Install qemu-lite
```
jodh-intel: To avoid possible confusion, it might be an idea to add a comment here stating that although this installs the qemu-lite version of qemu, the *binary* will still be called `qemu-system-x86_64`.

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,171 @@
...
+$ ./autogen.sh --prefix=/usr/local --disable-tests --disable-functional-tests --with-cc-image=/usr/share/clear-containers/clear-containers.img --with-cc-kernel=/usr/share/clear-containers/vmlinux.container --with-qemu-path=/usr/local/bin/qemu-system-x86_64
+$ make -j5 && sudo make install
+$ popd
+
+```
+
+## Install docker 1.12
+
+```
+$ sudo tee /etc/yum.repos.d/docker.repo <<-'EOF'
```
jodh-intel: The quotes around `EOF` aren't needed, nor is the dash, so this can become:

```
$ sudo tee /etc/yum.repos.d/docker.repo <<EOF
```

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,171 @@
...
+## Run a Clear Container
+You are now ready to run clear containers. For example:
+
+```
+$ sudo docker run -ti fedora bash
+```
+
+## Known Issues
+
+Currently `cc-oci-runtime v2.1` only works correctly with `docker-engine=1.12.1-1.el7.centos.x86_64`. While we update `cc-oci-runtime` to work with later versions of docker, we recommend to mark `docker-engine` to not being updated on the next upgrade.
```
jodh-intel: "While we update `cc-oci-runtime` to work with later versions of docker, we recommend running the command below to stop the `docker-engine` package from being updated when new package versions become available. Note that there is a risk in doing this since newer versions of Docker may contain security fixes."

documentation/Installing-Clear-Containers-on-ClearLinux.md
```diff
@@ -0,0 +1,49 @@
...
+$ sudo swupd check-update
+swupd-client software update checker 3.8.0
+   Copyright (C) 2012-2016 Intel Corporation
+
+Attempting to download version string to memory
+Current OS version: 13050
+There is a new OS version available: 13440
+
+```
+If your version is **13310** or newer, skip updating your host.
```
jodh-intel: It might just be simpler to say "run `swupd update`" as that will ensure the host is using the latest release and avoids the `check-update` step.

grahamwhaley: I don't think we should say 'skip updating' - as CLR is a rolling update distro :-) - but, we can say it is 'not necessary to update' :-)

jodh-intel: That was my concern exactly. At the risk of sounding like Dupont et Dupond, I'd say even more - "it is not necessary to update just for CC."

chavafg:
Thanks for your comments, all have been addressed.

jodh-intel:
lgtm! :smile:

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,203 @@
+# Install the Clear Containers runtime on Centos 7
+
+
+## Introduction
+
+Installing Clear Containers on CentOS is slightly more difficult than on other distributions mainly due to the fact that it does not provide current versions of gcc and glibc.
+
+Hence, this guide explains how to install updated packages in `/usr/local`, to avoid conflicting with the base packages.
+
+## Warning
```
grahamwhaley: I like this section btw :-)

documentation/Installing-Clear-Containers-on-Centos-7.md
```diff
@@ -0,0 +1,203 @@
...
+
+```
+$ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:$PKG_CONFIG_PATH
+$ sudo yum groupinstall 'Development Tools'
+$ sudo yum install zlib-devel libffi-devel
+
+$ curl -L -O "https://download.gnome.org/sources/glib/2.46/glib-2.46.2.tar.xz"
+$ tar -xvf glib-2.46.2.tar.xz
+$ pushd glib-2.46.2
+$ ./configure --prefix=/usr/local --disable-silent-rules && make -j5 && sudo make install
```
grahamwhaley: Just a side note - `-j5` is somewhat system dependent. Not a blocker, and I think we'll leave them in - but more generic would be no `-j` at all or just `-j` with no argument - both of which also have their downsides.

grahamwhaley:
lgtm

PR #694, opened 2017-02-23
runtime: Fix docker file descriptor leaks

### LINKED ISSUE #616: Docker is leaking file descriptors
It seems as though docker-containerd and docker-containerd-shim are leaking file descriptors.
I ran a 4K repetition loop that starts and stops 10 concurrent containers, and docker failed to continue launching more containers at iteration 3276.

**Steps to reproduce the issue:**

1. To reproduce this issue, I ran the following script to a reasonable number of iterations i.e. 4000

```
REPS=4000
MAX_CONTAINERS=10

docker_run()
{
  local ITER=$1
  echo -e "\nIteration $ITER"
  for i in $(seq 1 $MAX_CONTAINERS); do
        docker run -d clearlinux
        bash -c "while true; do sleep 1; done"
  done
}

docker_stop()
{
        docker stop $(docker ps -qa)
        docker rm -f $(docker ps -qa)
}

run_test()
{
for j in $(seq 1 $REPS); do
        docker_run $j
        sleep 2
        docker ps -a
        docker_stop
done
}
run_test
```

2. Count how many file descriptors are open in each iteration for the processes: qemu-lite-system-x86_64, dockerd, dockerd-containerd and docker-containerd-shim:

`lsof -n -p $(echo $(pidof  "${PID}") | awk '{$1=$1}1' OFS=" -p ") | wc -l"`

3. The count of file descriptors increased with the number of iterations as is shown in the following charts:

docker-containerd
docker-containerd-shim

4. Docker hangs after iteration 3275 once the number of docker-containerd opened files was: 65535
```
# lsof -n -p $(pidof docker-containerd) | wc -l
655535

```
5. max open files for docker-containerd process
```
cat /proc/$(pidof docker-containerd)/limits | grep "Max open files"
Max open files            65536
```

The complete data collection with count of file descriptors for the processes listed in 3 is below:

[output_23012017_0036_csv.txt](https://github.com/01org/cc-oci-runtime/files/725252/output_23012017_0036_csv.txt)

6. docker info:
```
Containers: 7
 Running: 5
 Paused: 0
 Stopped: 2
Images: 1
Server Version: 1.12.6
Storage Driver: overlay
 Backing Filesystem: extfs
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: host bridge null overlay
Swarm: inactive
Runtimes: cor runc
Default Runtime: cor
Security Options:
Kernel Version: 4.9.5-294.native
Operating System: Clear Linux Software for Intel Architecture
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 15.63 GiB
Name: clrContainersDev
ID: 3FHD:CE6K:AEUC:WLZS:RSSE:MEZU:2LSP:GODG:BSCC:XMA6:Q4SB:SSAN
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): true
 File Descriptors: 50
 Goroutines: 52
 System Time: 2017-01-23T17:32:04.880737686Z
 EventsListeners: 0
```

devimc:
Docker provides a cgroup path where the workload MUST be
added before create pid file otherwise docker will create a new
cgroup and add the workload to it.
When workload process ends the runtime MUST delete the cgroup path
notifying docker to free its open event file descriptors otherwise
docker won't close them causing file descriptors leaks.

fixes #616

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -60,6 +60,7 @@ struct subcommand *subcommands[] =
  */
 static struct spec_handler *stop_spec_handlers[] = {
 	&hooks_spec_handler,
+	&linux_spec_handler,
 
 	/* terminator */
 	NULL
@@ -153,6 +154,7 @@ handle_command_stop (const struct subcommand *sub,
 	gchar             *config_file = NULL;
 	gboolean           ret;
 	GNode*             root = NULL;
+	gchar             *cgroup_dir = NULL;
 
 	g_assert (sub);
 	g_assert (config);
@@ -207,6 +209,16 @@ handle_command_stop (const struct subcommand *sub,
 	g_print ("stopped container %s\n", config->optarg_container_id);
 
 out:
+	if (config->oci.oci_linux.cgroupsPath) {
+		cgroup_dir = g_strdup_printf ("%s/%s", CGROUP_MEM_DIR,
+			config->oci.oci_linux.cgroupsPath);
+		/* removing cgroup path will notify docker to close its event fds */
+		if (g_rmdir(cgroup_dir) != 0) {
+			g_critical("failed to remove cgroup dir: %s", cgroup_dir);
+		}
+		g_free_if_set (cgroup_dir);
+	}
+
 	g_free_if_set (config_file);
 	cc_oci_state_free (state);
 
--- a/src/oci.h
+++ b/src/oci.h
@@ -109,6 +109,11 @@
 /** Mode for \ref CC_OCI_WORKLOAD_FILE. */
 #define CC_OCI_SCRIPT_MODE		0755
 
+/** Mode for cgroup directory.
+ * 0755 is the mode used to create directories in /sys/fs/cgroup/memory/
+ */
+#define CC_OCI_CGROUP_MODE		0755
+
 /** Mode for all created directories. */
 #define CC_OCI_DIR_MODE                0750
 
@@ -136,6 +141,9 @@
 
 /* Path to the stateless passwd file. */ 
 #define STATELESS_PASSWD_PATH "/usr/share/defaults/etc/passwd"
+
+/* Path to memory cgroup directory */
+#define CGROUP_MEM_DIR "/sys/fs/cgroup/memory"
 
 /* Offset to add to the interface index for assigning the pci slot.
  * First 3 slots are in use for pc-lite machine type
@@ -254,6 +262,9 @@ struct oci_cfg_process {
 struct oci_cfg_linux {
 	/** List of \ref oci_cfg_namespace namespaces */
 	GSList          *namespaces;
+
+	/** cgroup path */
+	gchar           *cgroupsPath;
 };
 
 /** Representation of the OCI runtime schema embodied by
--- a/src/process.c
+++ b/src/process.c
@@ -788,6 +788,12 @@ cc_oci_vm_launch (struct cc_oci_config *config)
 	GSocketConnection *shim_socket_connection = NULL;
 	GError            *error = NULL;
 	int                status = 0;
+	char              *pid_str = NULL;
+	int                fd_tasks = -1;
+	int                fd_procs = -1;
+	gchar             *cgroup_dir = NULL;
+	gchar             *cgroup_tasks = NULL;
+	gchar             *cgroup_procs = NULL;
 
 	if (! config) {
 		return false;
@@ -1186,6 +1192,38 @@ child_failed:
 	 */
 	ret = cc_proxy_disconnect (config->proxy);
 
+	/* Before create pid file!
+	 *
+	 * Docker provides a cgroup path that MUST be created before pid file
+	 * workload pid MUST be copied to task and cgroup.procs notifying to docker
+	 * that the workload is part of a cgroup.
+	 * With this change docker WILL NOT create a new cgroup and WILL NOT copy
+	 * the workload pid to this new cgroup avoiding file descriptor leaks
+	 */
+	if (config->oci.oci_linux.cgroupsPath) {
+		cgroup_dir = g_strdup_printf ("%s/%s", CGROUP_MEM_DIR,
+			config->oci.oci_linux.cgroupsPath);
+		g_mkdir_with_parents (cgroup_dir, CC_OCI_CGROUP_MODE);
+
+		pid_str = g_strdup_printf ("%d", config->state.workload_pid);
+
+		cgroup_tasks = g_strdup_printf ("%s/tasks", cgroup_dir);
+		fd_tasks = open (cgroup_tasks, O_WRONLY);
+		if (write (fd_tasks, pid_str, strlen(pid_str)) <= 0) {
+			g_critical ("failed to copy workload pid to cgroup tasks: %s",
+				strerror(errno));
+			goto out;
+		}
+
+		cgroup_procs = g_strdup_printf ("%s/cgroup.procs", cgroup_dir);
+		fd_procs = open (cgroup_procs, O_WRONLY);
+		if (write (fd_procs, pid_str, strlen(pid_str)) <= 0) {
+			g_critical ("failed to copy workload pid to cgroup procs: %s",
+				strerror(errno));
+			goto out;
+		}
+	}
+
 	/* Finally, create the pid file.
 	 *
 	 * This MUST be done after all setup since containerd
@@ -1207,6 +1245,14 @@ out:
 	if (shim_err_fd != -1) close (shim_err_fd);
 	if (shim_args_fd != -1) close (shim_args_fd);
 	if (shim_socket_fd != -1) close (shim_socket_fd);
+
+	close_if_set (fd_procs);
+	close_if_set (fd_tasks);
+
+	g_free_if_set (cgroup_dir);
+	g_free_if_set (cgroup_tasks);
+	g_free_if_set (cgroup_procs);
+	g_free_if_set (pid_str);
 
 	if (setup_networking) {
 		netlink_close (hndl);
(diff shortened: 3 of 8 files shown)


src/command.c
```diff
@@ -153,6 +154,7 @@ handle_command_stop (const struct subcommand *sub,
 	gchar             *config_file = NULL;
 	gboolean           ret;
 	GNode*             root = NULL;
+	gchar             *cgroup_dir;
```
jodh-intel: Needs to be set to `NULL` for the `g_free_if_set()` call.

src/oci.h
```diff
@@ -109,6 +109,9 @@
 /** Mode for \ref CC_OCI_WORKLOAD_FILE. */
 #define CC_OCI_SCRIPT_MODE		0755
 
+/** Mode for cgroup directory. */
+#define CC_OCI_CGROUP_MODE		0755
```
jodh-intel: Does it have to be `0755`? If not, can we use `CC_OCI_DIR_MODE`?

src/command.c
jodh-intel: Ah - I see that it can't actually be `NULL` now, so this change isn't actually required (but might be good anyway from a maintenance perspective in case the code changes later).

src/command.c
```diff
@@ -207,6 +209,13 @@ handle_command_stop (const struct subcommand *sub,
 	g_print ("stopped container %s\n", config->optarg_container_id);
 
 out:
+	cgroup_dir = g_strdup_printf ("/sys/fs/cgroup/memory/%s",
```
jcvenegas: if (config->oci.oci_linux.cgroupsPath) ?

src/process.c
```diff
@@ -1186,6 +1193,33 @@ cc_oci_vm_launch (struct cc_oci_config *config)
...
 
+	/* Before create pid file!
+	 *
+	 * Docker provides a cgroup path that MUST be created before pid file
+	 * workload pid MUST be copy to task and cgroup.procs notifying to docker
+	 * that the workload is part of a cgroup.
+	 * With this change docker WILL NOT create a new cgroup and WILL NOT copy
+	 * the workload pid to this new cgroup avoiding file descriptor leaks
+	 */
+	cgroup_dir = g_strdup_printf ("/sys/fs/cgroup/memory/%s",
```
jcvenegas: check if cgroupsPath is not null

src/process.c
```diff
@@ -788,6 +788,13 @@ cc_oci_vm_launch (struct cc_oci_config *config)
 	GSocketConnection *shim_socket_connection = NULL;
 	GError            *error = NULL;
 	int                status = 0;
+	char               buf[64] = { 0 };
```
jcvenegas: why this magic number?

src/process.c
```diff
@@ -1186,6 +1193,33 @@ cc_oci_vm_launch (struct cc_oci_config *config)
...
+	cgroup_dir = g_strdup_printf ("/sys/fs/cgroup/memory/%s",
+		config->oci.oci_linux.cgroupsPath);
+	g_mkdir_with_parents (cgroup_dir, CC_OCI_CGROUP_MODE);
+
+	buf_len = g_snprintf (buf, sizeof(buf), "%d", config->state.workload_pid);
+
+	cgroup_tasks = g_strdup_printf ("%s/tasks", cgroup_dir);
+	fd_tasks = open (cgroup_tasks, O_WRONLY);
+	if (write (fd_tasks, buf, (size_t)buf_len) <= 0) {
+		g_critical ("failed to copy workload pid to cgroup tasks");
```
jodh-intel: I'd personally check the return from `open` but if that fails the `write` will detect it. However, I'd add a call to `strerror(errno)`.

src/process.c
```diff
@@ -1186,6 +1193,33 @@ cc_oci_vm_launch (struct cc_oci_config *config)
...
+		goto out;
+	}
+
+	cgroup_procs = g_strdup_printf ("%s/cgroup.procs", cgroup_dir);
+	fd_procs = open (cgroup_procs, O_WRONLY);
+	if (write (fd_procs, buf, (size_t)buf_len) <= 0) {
```
jodh-intel: Same comment as above.

src/process.c
```diff
@@ -1186,6 +1193,33 @@ cc_oci_vm_launch (struct cc_oci_config *config)
...
+		g_critical ("failed to copy workload pid to cgroup procs");
```
jodh-intel: Same comment as above.

src/command.c
jodh-intel: There are two occurrences of `"/sys/fs/cgroup/memory/%s"` so might be worth a macro for the static bit (or all of it), or even an macro / inline function for this.

src/process.c
jodh-intel: Why `64`?

src/process.c
```diff
@@ -1186,6 +1193,33 @@ cc_oci_vm_launch (struct cc_oci_config *config)
...
+	 * workload pid MUST be copy to task and cgroup.procs notifying to docker
+	 * that the workload is part of a cgroup.
+	 * With this change docker WILL NOT create a new cgroup and WILL NOT copy
+	 * the workload pid to this new cgroup avoiding file descriptor leaks
+	 */
+	cgroup_dir = g_strdup_printf ("/sys/fs/cgroup/memory/%s",
+		config->oci.oci_linux.cgroupsPath);
+	g_mkdir_with_parents (cgroup_dir, CC_OCI_CGROUP_MODE);
+
+	buf_len = g_snprintf (buf, sizeof(buf), "%d", config->state.workload_pid);
```
jodh-intel: Would `g_strdup_printf()` be easier to use here?

src/oci.h
devimc: it have to be 0755

src/spec_handlers/linux.c
```diff
@@ -102,6 +102,8 @@ handle_linux_section (GNode *root, struct cc_oci_config *config)
 		g_node_children_foreach(root, G_TRAVERSE_ALL,
 			(GNodeForeachFunc)handle_namespaces_section,
 			config);
+	} else if (! g_strcmp0 (root->data, "cgroupsPath")) {
+		config->oci.oci_linux.cgroupsPath = g_strdup (root->children->data);
```
jodh-intel: Could you add any new tests for this change (for example to `tests/spec_handlers/linux_test.c`)?

chavafg:
qa-passed

src/oci.h
jodh-intel: OK, in which case, I'd add a comment in the code explaining that's why you can't use `CC_OCI_DIR_MODE` (otherwise, someone might come along later and -- trying to be helpful -- "fix" this assuming it's a valid change to make ;-).

src/command.c
devimc: good catch!
thanks

src/spec_handlers/linux.c
devimc: sure

jodh-intel:
kudos for tracking this fixing this btw! :smile:

src/oci.h
devimc: ok, thanks

src/process.c
devimc: using g_strdup_printf

src/process.c
devimc: good catch,
thanks

devimc:
@jodh-intel  @jcvenegas thanks for your review

sboeuf:
@devimc Nice fix :)

1. Could you slightly change your commit message with something like that:
"Docker provides a cgroup path, and it expects related files to be filled with the proper PID. In case corresponding files are not filled as expected, it does it by itself, leading to file descriptors leaks.
In order to comply with Docker workflow and to make sure it does not try to fill those files by itself, we MUST create them before the PID workload file. PID MUST be copied to task and cgroup.procs files so as to notify to Docker that the workload is part of the expected cgroup."

2. Maybe a dumb question but where does the leak comes from exactly ? I mean, if we don't create those files for the specified cgroup, you say that Docker will create them, but where is the root issue ? Because I expect Docker to properly close the file descriptor he created, right ?

devimc:
@sboeuf 
1. sure, I can improve commit message
2. Docker creates a new cgroup (/sys/fs/cgroup/memory/xxxx) and adds the workload process to it
when the workload ends docker expects that runtime deletes these files, but we do not know where they are, for this reason docker keeps they open

sboeuf:
@devimc Thanks for the explanation. You say we don't know where those files are, but don't you think we can find them from the cgroup path provided by Docker ?
I ask because it would allow us to avoid the creation of those files, only the removal at the end.

devimc:
@sboeuf before this patch, docker created a new cgroup because we did not use the path provided by it,
we did not know where this new cgroup was created hence we cannot notify docker to free its open fds

sboeuf:
Ah okay, it relies on a different one in case we don't use the one provided....
Thanks

jcvenegas:
lgtm

src/process.c
```diff
@@ -1186,6 +1192,38 @@ cc_oci_vm_launch (struct cc_oci_config *config)
 	 */
 	ret = cc_proxy_disconnect (config->proxy);
 
+	/* Before create pid file!
+	 *
+	 * Docker provides a cgroup path that MUST be created before pid file
+	 * workload pid MUST be copy to task and cgroup.procs notifying to docker
```
amshinde: s/copy/copied/

devimc: good catch!  thanks

amshinde:
lgtm.
Nice fix Julio!

PR #696, opened 2017-02-23
create: Kill the VM along with shim on create error.

amshinde:
When "create" fails, we forcefully kill the shim with SIGKILL.
The vm may still be around at this point, kill the vm as well.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -768,7 +768,7 @@ gboolean
 cc_oci_vm_launch (struct cc_oci_config *config)
 {
 	gboolean           ret = false;
-	GPid               pid;
+	GPid               pid = -1;
 	ssize_t            bytes;
 	char               buffer[2] = { '\0' };
 	int                hypervisor_args_pipe[2] = {-1, -1};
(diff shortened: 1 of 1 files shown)


dlespiau:
That looks like something we want to do indeed.

The vm_launch() function could do with some trimming. Also, if we introduce something that can fail *after* vm_launch() at some point in the future, we won't kill the shim/qemu processes then. That said, as 2.1 is in maintenance mode:

lgtm.

src/process.c
```diff
@@ -1225,6 +1225,18 @@ cc_oci_vm_launch (struct cc_oci_config *config)
 		kill(config->state.workload_pid, SIGKILL);
 	}
 
+	/* We have force killed the shim if it is running at this pointi,
```
dlespiau: s/pointi/point/

amshinde: Fixed.

dlespiau:
Oh, IIRC, delete() is still called if create() fails, so I was wondering if we shouldn't cleanup things in delete() instead. But I guess create() is ok as well, we know what we need to kill at the end of create(), so that's a bonus.

chavafg:
qa-failed

amshinde:
@dlespiau IIRC, we rely on "destroypod" to shut down the VM in delete, for which the runtime needs to send attach to the proxy with containerid. This may not work in cases where vm_launch() errors out before creating any pod/container.

amshinde:
Its possible that the VM is launched and the proxy is not aware of the VM  at all.

chavafg:
qa-passed

jcvenegas:
lgtm

PR #697, opened 2017-02-23
fdleak: Don't let Compare() return true when it shouldn't

dlespiau:
*Note: This bug was discovered in clearcontainers/proxy when testing against go tip (what will become 1.9). It's a real bug though that could have happened even with earlier go versions.*

When we've already failed the comparison for a previous fd, nothing
should be able to make us return true.

Unfortunately, we were overriding unconditionally the equal boolean
with no regard of the past.

This was found testing the proxy on go tip (what will become 1.9), the
runtime has changed and exposed that bug.


chavafg:
qa-passed

jcvenegas:
lgtm

devimc:
lgtm

PR #699, opened 2017-02-27
Metrics/bootchart

jcvenegas:
This PR adds a new service called cc-bootchart for systemd based images.  If a bootchart was created at boot time ( by systemd-bootchart)  it will copy  the svg to the container /run directory. 

This service should be used just for measures and would not be installed by default in a clear container image. A target is added to insert the service in the container image called insert-bootchart.

### DIFF
--- a/data/cc-bootchart.service
+++ b/data/cc-bootchart.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=Copy systemd-bootchart to container rootfs
+
+[Service]
+Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+Type=simple
+#wait some time before copy the bootchart svg
+ExecStart=/usr/bin/bash -c 'sleep 2; cp /run/log/*.svg /tmp/hyper/shared/run'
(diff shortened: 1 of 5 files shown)


chavafg:
qa-passed

data/cc-bootchart.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Copy systemd-bootchart to container rootfs
+
+[Service]
+Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+Type=simple
+#wait some time before copy the bootchart svg
+ExecStart=/usr/bin/bash -c 'sleep 2; cp /run/log/*.svg /tmp/hyper/shared/run'
```
gorozco1: sleep 2

I remember @dlespiau had some comments regarding that sleep is not reliable

jodh-intel: Yes - this `sleep` needs to be replaced with something that copies the output file *once it's been fully generated*.

devimc: not sure if declaring this service as ```idle``` will do the magic

jcvenegas: @devimc it is not bad idea, and elegant if it works, let me try

chavafg:
qa-passed

jcvenegas:
Closing this one , a different approach was used in #730

PR #704, opened 2017-02-27
runtime: add --all option to kill command

### LINKED ISSUE #693: Add --all option for kill command
`runc` added a `--all`l option to the `kill` command on commit https://github.com/opencontainers/runc/commit/e58671e530eb04a233d953c6a56e56804703f80c. This is being used by docker 1.13 so we also need to support `--all`.

devimc:
With this patch users will can send signal to
all processes running inside the container

fixes #693

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -222,17 +222,25 @@ err:
  * \param config \ref cc_oci_config.
  * \param state \ref oci_state.
  * \param signum Signal number to send to hypervisor.
+ * \param allProcesses \c true to send a signal to all container's processes,
+ *  \c false to send a signal to container's workload.
  *
  * \return \c true on success, else \c false.
  */
 gboolean
 cc_oci_kill (struct cc_oci_config *config,
 		struct oci_state *state,
-		int signum)
+		int signum,
+		gboolean allProcesses)
 {
 	enum oci_status last_status;
 
 	if (! (config && state)) {
+		return false;
+	}
+
+	if (! cc_oci_get_signame (signum)) {
+		g_critical ("signal %d is not supported", signum);
 		return false;
 	}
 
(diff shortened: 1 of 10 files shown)


chavafg:
qa-passed

src/oci.c
```diff
@@ -261,35 +267,31 @@ cc_oci_kill (struct cc_oci_config *config,
...
-		if (! cc_proxy_hyper_kill_container (config, signum)) {
-			g_critical ("failed to kill container");
+	if (signum == SIGKILL) {
+		/* send signal to cc-shim because SIGKILL will end
+		 * the VM and cc-shim is not able to catch it
+		 */
+		if (kill (state->pid, signum) < 0) {
+			g_critical ("failed to stop cc-shim %u: %s",
+					(unsigned)state->pid,
+					strerror (errno));
```
jodh-intel: Did you change the order of the `kill()` and `cc_proxy_hyper_kill_container()` calls to minimise the chance that the container would still have running processes *after* `cc-shim` was killed? If so, might be worth a comment explaining that.

Presumably, if there were any `exec`'s at the time of this call, all the other `cc-shim` processes would also detect this and exit themselves?

src/oci.c
```diff
@@ -228,14 +228,20 @@ cc_oci_get_config_and_state (gchar **config_file,
 gboolean
 cc_oci_kill (struct cc_oci_config *config,
 		struct oci_state *state,
-		int signum)
+		int signum,
+		gboolean allProcesses)
```
jodh-intel: Needs a `\param` in the function header above.

src/util.c
```diff
@@ -119,6 +119,30 @@ cc_oci_get_signum (const gchar *signame)
 }
 
 /*!
+ *
```
jodh-intel: "Convert the specified signal number into its symbolic name".

src/util.c
```diff
@@ -119,6 +119,30 @@ cc_oci_get_signum (const gchar *signame)
...
+ * \param signum Numer of signal.
+ *
+ * \return signal name, or NULL on error.
```
jodh-intel: "\return static signal name string, or NULL on error."

jodh-intel:
This looks great - just one query and a few comments to add.

src/oci.c
devimc: I kill container's processes before cc-shim because cc-shim will catch those signals (exit codes) and will end.
cc-shim is not able to catch SIGKILL hence I have to kill it too
Adding comments ..
Thank you!

src/oci.c
devimc: good catch! thanks

jodh-intel:
lgtm.

sameo:
LGTM

PR #714, opened 2017-03-02
tests/metrics: Add support to measure RSS and PSS

MarioCarrilloA:
This script will show an average about RSS memory
and PSS memory. RSS memory is measured by collectl
tool and PSS memory by psstop tool.

### DIFF
--- a/tests/metrics/collect_mem_consmd.sh
+++ b/tests/metrics/collect_mem_consmd.sh
@@ -47,11 +47,22 @@ CONT_GATE="$CONT_SHARED_PATH/gate"
 HOST_RDM_FILE="$HOST_SHARED_PATH/out.data"
 CONT_RDM_FILE="$CONT_SHARED_PATH/out.data"
 CSV_FILE="mem.csv"
+PSS_FILE="pss.log"
 
 declare WORKLOAD
 declare COLLECT_PID
+declare SMEM_PID
 declare F_OPTS
-declare AVG
+declare DATA_RES
+declare RSS_RES
+declare PSS_RES
+
+# columns in collect output
+declare -A COLLECT_MEM=(
+	["VmStk"]=13
+	["VmData"]=12
+	["VmRSS"]=11
+	["VmLck"]=10)
 
 # Shows help about this script
 function help()
@@ -117,7 +128,8 @@ EOF
 # in order to get metrics about a specific tool.
 function start_collect()
 {
-	collectl -oTm -sZm -i1:1 \
+	echo "Launching collectl"
+	collectl -oTm -sZm -i0.5:1 \
 		--procfilt f$PROC \
 		--procopts tm -f $PROC-mem &
 
@@ -128,38 +140,105 @@ function start_collect()
 # It kills the collectl process.
 function stop_collect()
 {
+	echo "Stop collectl"
 	kill "$COLLECT_PID"
 }
 
-# Get the average about a "csv" file produced
-# by collectl tool.
-function get_average()
+# It launches a instance of smem tool
+# in order to get metrics about proportional
+# set size.
+function start_pss_monitor()
 {
-	sum=0
-	count=0
+	bash ./smem_monitor.sh $PROC &
+	SMEM_PID=$!
+}
+
+# It terms smem process
+function stop_pss_monitor()
+{
+	kill "$SMEM_PID"
+}
+
+# Get PSS memory metrics
+function get_pss_metrics()
+{
+	pss_sum=0
+	pss_count=0
+	samples=""
+
+	while read line; do
+		pss_mem=$(echo "$line" | sed 's/[A-Za-z]*//g')
+		res=$(echo "$pss_mem > 0" | bc )
+		if [ $res -eq 1 ];then
+			pss_sum=$(echo "$pss_sum + $pss_mem"| bc -l )
+			pss_count=$(( $pss_count + 1 ))
+			if [ ! -z "$samples" ];then
+				samples="$samples ,"
+			fi
+			samples="$samples $pss_mem"
+		fi
+	done < "$PSS_FILE"
+
+	stdev=$(python3 -c "import statistics; print (statistics.stdev([$samples]))")
+	avg=$(echo "$pss_sum / $pss_count" | bc -l)
+	pstdev=$(echo "($stdev / $avg) * 100" | bc -l)
+	echo "avg: $avg stdev: $stdev pstdev: $pstdev"
+}
+
+# Get metrics about a "csv" file produced
+# by collectl tool.
+function get_collect_metrics()
+{
+	data_sum=0
+	data_count=0
+	samples=""
 
 	while read line; do
 		cmd=$(echo "$line" | awk -F "," '{print $30}' | grep $PROC)
-		if [ ! -z "$cmd" ];then
-			binary="$(echo "$cmd" | cut -d " " -f1 | \
-				xargs basename -z | grep $F_OPTS $PROC)"
-			if [ ! -z "$binary" ];then
-				mem_data=$(echo "$line" | awk -F "," '{print $12}')
-				if (( $mem_data > 0 ));then
-					sum=$(( $sum + $mem_data ))
-					count=$(( $count + 1 ))
-				fi
+		if [ -z "$cmd" ];then
+			continue
+		fi
+
+		binary="$(echo "$cmd" | cut -d " " -f1 | \
+			xargs basename -z | grep $F_OPTS $PROC)"
+
+		if [ -z "$binary" ];then
+			continue
+		fi
+
+		data_mem=$(echo "$line" | awk -v column=$1 -F "," '{print $column}')
+
+		if (( $data_mem > 0 ));then
+			data_sum=$(( $data_sum + $data_mem ))
+			data_count=$(( $data_count + 1 ))
+			if [ -n "$samples" ];then
+				samples="$samples ,"
 			fi
+			samples="$samples $data_mem"
 		fi
 
 	done < "$CSV_FILE"
 
-	if (( $count == 0 ));then
+		if (( $data_count == 0 ));then
 		echo -e "\e[31m$MYSELF: error: no collected samples\e[39m"
 		exit 1;
 	fi
-	AVG=$(echo "$sum / $count" | bc)
-	echo -e "\e[33mMemory consumption average: $AVG\e[39m"
+
+	stdev=$(python3 -c "import statistics; print(statistics.stdev([$samples]))")
+	avg=$(echo "$data_sum / $data_count" | bc -l)
+	pstdev=$(echo "($stdev / $avg) * 100" | bc -l)
+
+	echo "avg: $avg stdev: $stdev pstdev: $pstdev"
+}
+
+# Print the metrics memory
+function print_results()
+{
+	echo -e "\e[33m==== Results ===="
+	echo "$PROC VmData: $DATA_RES"
+	echo "$PROC RSS:    $RSS_RES"
+	echo "$PROC PSS:    $PSS_RES"
+	echo -e "=================\e[39m"
 }
 
 # It creates a "csv" file using a raw information
@@ -193,13 +272,9 @@ function check_proxy_service()
 		echo -e "\e[31merror: cc-proxy could not start\e[39m"
 		exit 1
 	fi
-
-	sleep 1
 }
 
-# This function will check the necessary requirements
-# in the environment for running Clear Containers.
-function set_environment()
+function clean_docker_waste()
 {
 	echo "Cleaning environment"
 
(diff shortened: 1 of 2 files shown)


tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
+	echo "Launching psstop"
+	psstop_monitor=$(cat << EOF
+	#!/bin/bash
+
 	sum=0
 	count=0
 
+	while [ 1 ];do
+		if [ ! -f "$STOP_FILE" ];then
+			data=\$(psstop $PROC | awk '{print \$1}')
```
jodh-intel: `psstop` is packaged for Clear, but is not available in the Ubuntu and Fedora repos. An alternative tool that shows PSS that *is* available in all 3 distros is `smem`. In fact, https://github.com/01org/cc-oci-runtime/blob/master/tests/metrics/density/docker_memory_usage.sh.in is already using that.

What is the overlap between this script and the existing one?

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
-# Get the average about a "csv" file produced
-# by collectl tool.
-function get_average()
+# It launches a instance of psstop tool
+# in order to get metrics about proportional
+# set size.
+function start_psstop()
 {
+	echo "Launching psstop"
+	psstop_monitor=$(cat << EOF
```
jodh-intel: It would be simpler if this script didn't automatically generate another script, but just called another script passing the arguments it needs.

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
+			data=\$(psstop $PROC | awk '{print \$1}')
+			for i in \$data;do
+				if (( \$i > 0 ));then
+					sum=\$(( \$i + \$sum ))
+					count=\$(( \$count + 1))
+				fi
+			done
+
+			if (( \$count > 0 ));then
+				avg=\$(( \$sum / \$count ))
```
jodh-intel: This will give an integer result. If that's not what's expected, you could use:

```
avg=$(echo "scale=4; $sum / $count"|bc -l)
```

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
+				echo "\$avg" >> $PSS_FILE
+			fi
+
+			sleep 0.5
+		else
+			break;
```
jodh-intel: It's simpler to have the test to break out of the loop immediately after the `while`. That way, you don't need the indent.

tests/metrics/collect_mem_consmd.sh
```diff
@@ -47,11 +47,16 @@ CONT_GATE="$CONT_SHARED_PATH/gate"
 HOST_RDM_FILE="$HOST_SHARED_PATH/out.data"
 CONT_RDM_FILE="$CONT_SHARED_PATH/out.data"
 CSV_FILE="mem.csv"
+PSS_FILE="pss.log"
+STOP_FILE="$(pwd)/stop.log"
```
jodh-intel: You can't assume that `$(pwd)` is writable, so this should really be:

```
STOP_FILE=$(mktemp)
```

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
+
+	while read line; do
+		pss_mem=$(echo "$line" | sed 's/[A-Za-z]*//g')
+		if (( $pss_mem > 0 ));then
+			pss_sum=$(( $pss_sum + $pss_mem ))
+			pss_count=$(( $pss_count + 1 ))
+		fi
+	done < "$PSS_FILE"
+
+	average=$(( $pss_sum / $pss_count ))
```
jodh-intel: Again, this will give an integer result for `bash` (`zsh` doesn't have that limitation! :smile:)

tests/metrics/collect_mem_consmd.sh
```diff
@@ -197,9 +280,7 @@ function check_proxy_service()
 	sleep 1
 }
 
-# This function will check the necessary requirements
-# in the environment for running Clear Containers.
-function set_environment()
+function clean_docker_waste()
```
jodh-intel: I :heart: the name of this function! :smile:

tests/metrics/collect_mem_consmd.sh
```diff
@@ -274,24 +366,32 @@ function main()
 
 	echo "Active gate"
 	echo "True" > $HOST_GATE
+	sleep 0.5
```
jodh-intel: What's this for?

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +134,115 @@ function start_collect()
...
+	done < "$PSS_FILE"
+
+	average=$(( $pss_sum / $pss_count ))
+	echo "$average"
+}
+
+
+# Get the average about a "csv" file produced
+# by collectl tool.
+function get_collect_average()
```
jodh-intel: Needs an explanation of what `$1` is.

tests/metrics/collect_mem_consmd.sh
```diff
@@ -274,24 +366,32 @@ function main()
...
 
 	echo "Workload finishes"
+	sleep 1
 	stop_collect
 	create_csv_file
 
 	echo "Making a total"
-	get_average
+	RSS_AVG=$(get_collect_average 11)
+	DATA_AVG=$(get_collect_average 12)
```
jodh-intel: Why `11` and `12`?

tests/metrics/collect_mem_consmd.sh
```diff
@@ -274,24 +366,32 @@ function main()
...
 		if (( $cont_num == 1 ));then
+			stop_psstop
 			break
 		fi
 
 		sleep 0.1
 	done
 
 	echo "Workload finishes"
+	sleep 1
```
jodh-intel: What's this for?

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: Me too 🙂

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: Ups!, magic numbers, i will fix that

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: It was for taking some samples after workload finishes, but at this moment it is unnecessary

MarioCarrilloA:
I will send a new version 😉

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +140,100 @@ function start_collect()
...
+			pss_sum=$(echo "$pss_sum + $pss_mem"| bc -l )
+			pss_count=$(( $pss_count + 1 ))
+			if [ ! -z "$samples" ];then
+				samples="$samples ,"
+			fi
+			samples="$samples $pss_mem"
+		fi
+	done < "$PSS_FILE"
+
+	stdev=$(python3.6 -c "import statistics; print (statistics.stdev([$samples]))")
```
jodh-intel: This looks fragile - does it need to be Python 3.6+ or could you just specify `python3`?

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +140,100 @@ function start_collect()
...
+# Get PSS memory metrics
+function get_pss_metrics()
+{
+	pss_sum=0
+	pss_count=0
+	samples=""
+
+	while read line; do
+		pss_mem=$(echo "$line" | sed 's/[A-Za-z]*//g')
+		res=$(echo "$pss_mem > 0" | bc )
```
jodh-intel: It would be a good idea to check that `bc` is available before starting this script. Something like:

```
cmd=bc
command -v "$cmd" 2>/dev/null || { echo "ERROR: command $cmd not available"; exit 1; }

```

You could make that clearer if you defined a `die` function:

```
die()
{
    local msg="$*"
    echo >&2 "ERROR: $msg"
    exit 1
}

    :
    :

cmd=bc
command -v "$cmd" 2>/dev/null || die "command $cmd not available"
```

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +140,100 @@ function start_collect()
...
+		if [ $res -eq 1 ];then
+			pss_sum=$(echo "$pss_sum + $pss_mem"| bc -l )
+			pss_count=$(( $pss_count + 1 ))
+			if [ ! -z "$samples" ];then
```
jodh-intel: This is fine, but a more succinct method (that doesn't need the `if ... fi`) is:

```
[ -n "$samples" ] && samples="$samples ,"
```

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +140,100 @@ function start_collect()
...
 		cmd=$(echo "$line" | awk -F "," '{print $30}' | grep $PROC)
 		if [ ! -z "$cmd" ];then
 			binary="$(echo "$cmd" | cut -d " " -f1 | \
 				xargs basename -z | grep $F_OPTS $PROC)"
 			if [ ! -z "$binary" ];then
-				mem_data=$(echo "$line" | awk -F "," '{print $12}')
-				if (( $mem_data > 0 ));then
-					sum=$(( $sum + $mem_data ))
-					count=$(( $count + 1 ))
+				data_mem=$(echo "$line" | awk -v colum=$1 -F "," '{print $colum}')
```
jodh-intel: Typo: `colum` should probably be `column`?

jodh-intel: Another comment: this deeply indented code is very difficult to read (and thus more likely to contain errors). I'd try to simplify it by doing things like:

```
while read line; do
    # If the line doesn't match, ignore immediately.
    # This means you save 1 leve of indentation - easier to read ;-)
    [ -n "$cmd" ] || continue

    # Again, test the line and ignore it if you don't get the result you want.
    [ -n "$binary" ] || continue

    # The main code is now here, but doesn't need to be deeply nested.
   
done
```

tests/metrics/collect_mem_consmd.sh
```diff
@@ -128,38 +140,100 @@ function start_collect()
...
 
-	if (( $count == 0 ));then
+		if (( $data_count == 0 ));then
 		echo -e "\e[31m$MYSELF: error: no collected samples\e[39m"
 		exit 1;
 	fi
-	AVG=$(echo "$sum / $count" | bc)
-	echo -e "\e[33mMemory consumption average: $AVG\e[39m"
+
+	stdev=$(python3.6 -c "import statistics; print(statistics.stdev([$samples]))")
```
jodh-intel: `python3`?

tests/metrics/smem_monitor.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
```
jodh-intel: 2017?

tests/metrics/smem_monitor.sh
```diff
@@ -0,0 +1,58 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to measure the PSS average memory.
+
+# set -x
```
jodh-intel: Please remove commented-out code.

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/smem_monitor.sh
MarioCarrilloA: +1

tests/metrics/smem_monitor.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: Hi, thank you for the comments and regarding to this point, I know that is not necessary to use “if [ ] ... fi” however I think is easier to read the code using them, I like when a user does not need to be an expert  in a specific language (in this case bash)in order to understand a source code. For that reason I think is better to be clear on each line. What do you think?

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
MarioCarrilloA: +1

tests/metrics/collect_mem_consmd.sh
jodh-intel: Sure - to be clear, this comment was only a thought - I'm not suggesting you need to change your code to remove the `if... fi` as both are perfectly valid :smile:

MarioCarrilloA: perfect, I will send new version :)

jodh-intel:
lgtm

dlespiau:
lgtm

PR #722, opened 2017-03-06
mount: Remove bogus parameter check

sameo:
We only want to fail if our config is NULL. The list of OCI mounts being
empty is a perfectly valid use case.


jcvenegas:
Hi @sameo , could you explain a bit more why is bogus? I see that it perform an usual check

chavafg:
qa-passed

amshinde:
lgtm

grahamwhaley:
@sameo thanks for the explanation in the commit now - with that,
lgtm

PR #745, opened 2017-03-14
tests/metrics: Add tool for memory mapping

MarioCarrilloA:
This tool will report the memory map of a process
in JSON file. Each mapping has the next information:

- Proportional set size
- Shared memory
- Private memory

### DIFF
--- a/tests/metrics/README.md
+++ b/tests/metrics/README.md
@@ -48,3 +48,18 @@ collectl
 ```bash
 # ./collect_mem_consmd.sh -c 50 -m 1000 -p cc-shim -x
 ```
+
+### Mapping tool
+
+The `map_mem.sh` tool will report the memory map of a process in a JSON file.
+Each mapping has the following information:
+
+- Proportional set size
+- Shared memory
+- Private memory
+
+**Usage example:**
+
+```bash
+# ./map_mem.sh cc-proxy
+```
--- a/tests/metrics/map_mem.sh
+++ b/tests/metrics/map_mem.sh
@@ -1,0 +1,351 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to report the memory map of a process in
+#  a JSON file
+
+declare MYSELF=${0##*/}
+declare SAMPLES_FILE=$(mktemp)
+declare DEFAULT_SAMPLES=10
+
+declare NUM_SAMPLES
+declare JSON_FILE
+declare CC_ID
+
+# shows help about this script
+function help()
+{
+usage=$(cat << EOF
+Usage: $MYSELF [-h] [--help] [-v] [--version]
+  Description:
+       This tool is to report the memory map of a process in
+       a JSON file.
+  Options:
+       -h, --help             Help page
+       -p, --process-name     Process Name
+       -n, --num-samples      The number of samples that will be
+                              collected
+       -v, --version          Show version.
+EOF
+)
+	echo "$usage";
+}
+
+# exit with error
+function die()
+{
+	local msg="$*"
+	echo >&2 "error: $msg"
+	exit 1
+}
+
+# Checking env
+function init()
+{
+	req_tools=(
+		pmap
+		awk
+		bc)
+
+	if [ -z "$PROC" ];then
+		die "process name no found"
+	fi
+
+	if [ -z "$NUM_SAMPLES" ];then
+		NUM_SAMPLES="$DEFAULT_SAMPLES"
+	fi
+
+	# check available tools
+	echo "check required tools"
+	for cmd in ${req_tools[*]};do
+		command -v "$cmd" &> /dev/null
+		res=$?
+		if [ $res -ne 0 ];then
+			die "command $cmd not available"
+		fi
+		echo "command: $cmd: yes"
+	done
+}
+
+# dump memory avg to JSON file
+function dump_avg_JSON()
+{
+	local pss=$1
+	local shared=$2
+	local private=$3
+	local count=$4
+
+	pss_avg=$(echo "scale=2; $pss / $count" | bc)
+	shared_avg=$(echo "scale=2; $shared / $count" | bc -l )
+	private_avg=$(echo "scale=2; $private / $count" | bc -l)
+
+	echo -e "\t\t\"pss\": $pss_avg," >> $JSON_FILE
+	echo -e "\t\t\"shared\": $shared_avg," >> $JSON_FILE
+	echo -e "\t\t\"private\": $private_avg" >> $JSON_FILE
+}
+
+# it creates JSON file
+function create_json()
+{
+	local process_name="$1"
+	local samples_file="$2"
+
+	local mapping=""
+	local pss_mem=0
+	local shared_mem=0
+	local private_mem=0
+	local count=0
+
+	# accumulators
+	local shared_sum=0
+	local private_sum=0
+	local pssmem_sum=0
+
+	declare -A memory=(
+		["mapping"]=1
+		["pssmem"]=2
+		["shared"]=3
+		["private"]=4)
+
+	JSON_FILE="${process_name}.json"
+	output="$(cat "$samples_file" | sort)"
+
+	# start JSON file creation
+	echo "{" > "$JSON_FILE"
+
+	while read line;do
+		mapping=$(echo "$line" | awk -F "," -v column=${memory["mapping"]} '{print $column}')
+		pss_mem=$(echo "$line" | awk -F "," -v column=${memory["pssmem"]} '{print $column}')
+		shared_mem=$(echo "$line" | awk -F "," -v column=${memory["shared"]} '{print $column}')
+		private_mem=$(echo "$line" | awk -F "," -v column=${memory["private"]} '{print $column}')
+		count=$(( $count + 1 ))
+
+		# last state
+		if [ -z "$last_mapping" ];then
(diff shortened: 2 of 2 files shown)


tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+	echo "check required tools"
+	for cmd in ${req_tools[*]};do
+		command -v "$cmd" &> /dev/null
+		res=$?
+		if [ $res -eq 0 ];then
+			echo "command: $cmd: yes"
+			continue
+		fi
+
+		echo  "error: command $cmd not available"
```
jodh-intel: Shouldn't this be an error condition?

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to report the memory map of a process in
+#  a JSON file
+
+declare PROC="$1"
+declare NUM_SAMPLES=10
+declare SAMPLES_FILE="rmap.log"
+declare JSON_file
```
jodh-intel: This variable uses a strange combination of capitals and lower-case letters ;-) Maybe `JSON_FILE` would be more consistent with the rest of this script?

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+			continue
+		fi
+
+		echo  "error: command $cmd not available"
+	done
+
+	# clean env
+	echo "cleaning env"
+	rm -rf *.log
+	rm -rf *.json
```
jodh-intel: These commands are too loose - the script will delete files that the user might want to keep. It would be better to create a new (empty) temporary directory to store logs in (so that the clean operation isn't required), or only delete named logfiles that this script creates. I think the temporary directory option is better.

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+}
+
+# dump memory avg to JSON file
+function dump_avg_JSON()
+{
+	pss=$1
+	shared=$2
+	private=$3
+	count=$4
```
jodh-intel: Shouldn't these variables by `local` for consistency and to avoid polluting the global variable namespace?

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+		["mapping"]=1
+		["pssmem"]=2
+		["shared"]=3
+		["private"]=4)
+
+	JSON_file="${PROC}.json"
+	output="$(cat "$SAMPLES_FILE" | sort)"
+
+	# start JSON file creation
+	echo "{" > "$JSON_file"
```
jodh-intel: It would be clearer I think if you collected up all the data you need in variables, then did something like:

```
cat >>"$JSON_file" <<EOT
{
    "foo": "$foo",
    "bar": "$bar"
}
EOT
```

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+
+	# Finish JSON file creation
+	echo "}" >> $JSON_file
+}
+
+# take one sample of mapping
+function take_map_sample()
+{
+	local ps_name=$1
+	local format_columns=23
```
jodh-intel: This could do with a comment explaining what it's used for.

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+	echo -e "\t}" >> $JSON_file
+
+	# Finish JSON file creation
+	echo "}" >> $JSON_file
+}
+
+# take one sample of mapping
+function take_map_sample()
+{
+	local ps_name=$1
```
jodh-intel: Best to quote all values (same comment to the other functions in this script):
```
local ps_name="$1"
```

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+		["shugep"]=17
+		["pclean"]=11
+		["pdirty"]=12
+		["phugep"]=18)
+
+	# getting pid of the process
+	pid=$(pidof $ps_name | awk '{print $1}')
+	if [ -z "$pid" ];then
+		docker rm -f  "$CC_ID"
+		echo "process does not exis"
```
jodh-intel: s/exis/exist/. But also, can you include `"$pid` in the error message.

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+		pssmem_sum=$(( $pssmem_sum + $pssmem ))
+
+	done <<< $(echo -e "$output")
+
+	# dump buffer data"
+	echo "$mapping,$pssmem_sum,$shared_sum,$private_sum" >> "$SAMPLES_FILE"
+	echo "[anon],$anon_pssmem,$anon_shared,$anon_private" >> "$SAMPLES_FILE"
+}
+
+# main
```
jodh-intel: I'd remove this comment as it doesn't add much.

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to report the memory map of a process in
+#  a JSON file
+
+declare PROC="$1"
+declare NUM_SAMPLES=10
```
jodh-intel: Might be useful to make this a command-line option (but default to `10`)?

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+	local pssmem_sum=0
+
+	declare -A memory=(
+		["mapping"]=1
+		["pssmem"]=2
+		["shared"]=3
+		["private"]=4)
+
+	JSON_file="${PROC}.json"
+	output="$(cat "$SAMPLES_FILE" | sort)"
```
jodh-intel: Not required, but it might be better to pass in `$PROC` and `$SAMPLES_FILE` as parameters to this function.

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  The objective of this tool is to report the memory map of a process in
+#  a JSON file
+
+declare PROC="$1"
+declare NUM_SAMPLES=10
+declare SAMPLES_FILE="rmap.log"
```
jodh-intel: If you do something like this, this script will fail:
```
cd /read/only/mount/point
/path/to/cc-oci-runtime/tests/metrics/map_mem.sh
```

To make this script more robust, it would be better to create the samples file in a temporary location:

```
declare SAMPLES_FILE=$(mktemp)
```

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,287 @@
...
+	# accumulators
+	local shared_sum=0
+	local private_sum=0
+	local anon_pssmem=0
+	local anon_shared=0
+	local anon_private=0
+
+	# types of memory
+	declare -A memory=(
+		["pssmem"]=8
```
jodh-intel: Again, an explanatory comment on what these numbers are would be useful from a maintenance perspective.

tests/metrics/map_mem.sh
MarioCarrilloA: I made some experiments during this script, one of them was to collect different quantity of samples (10, 100, ...500 etc), I noticed that the value of the variance does not affect the values, I got very similar values taking 10 samples and N samples (N > 10  and N < 500), if you consider that it is a good idea to add support for changing this value, I can add it, what do you think?

tests/metrics/map_mem.sh
MarioCarrilloA: +1

tests/metrics/map_mem.sh
jodh-intel: On this point, I don't mind either way (it just looked like a value that users might want to vary).

tests/metrics/map_mem.sh
MarioCarrilloA: I used that way because the body of the JSON file could have "N" lines, but yes I will look for a better option

tests/metrics/map_mem.sh
MarioCarrilloA: Ops!  +1

tests/metrics/map_mem.sh
MarioCarrilloA: +1

tests/metrics/map_mem.sh
MarioCarrilloA: +1

tests/metrics/map_mem.sh
MarioCarrilloA: agree +1

MarioCarrilloA:
I will send a version 2

tests/metrics/map_mem.sh
MarioCarrilloA: perfect, I think it could be a better idea to add that option, nice

tests/metrics/map_mem.sh
MarioCarrilloA: $pid could not be included because it does not exist, but the typo will be fixed :)

tests/metrics/map_mem.sh
```diff
@@ -0,0 +1,351 @@
...
+		["shugep"]=17
+		["pclean"]=11
+		["pdirty"]=12
+		["phugep"]=18)
+
+	# getting pid of the process
+	pid=$(pidof $ps_name | awk '{print $1}')
+	if [ -z "$pid" ];then
+		docker rm -f  "$CC_ID"
+		die "process does not exist"
```
jodh-intel: Yes, sorry, I meant `$ps_name`, not `$pid` in this error message. Not a blocker though ;)

jodh-intel:
lgtm

dlespiau:
lgtm

chavafg:
qa-passed

gorozco1:
lgtm

tests/metrics/README.md
```diff
@@ -48,3 +48,18 @@ collectl
 ```bash
 # ./collect_mem_consmd.sh -c 50 -m 1000 -p cc-shim -x
 ```
+
+### Mapping tool
+
+The `map_mem.sh` tool will report the memory map of a process in JSON file.
```
rcaballeromx: ... in a JSON file.

tests/metrics/README.md
```diff
@@ -48,3 +48,18 @@ collectl
...
+Each mapping has the next information:
```
rcaballeromx: ... has the following information:

jcvenegas:
@rcaballeromx could you please review this PR again, @MarioCarrilloA  added the changes requested?

sboeuf:
lgtm

PR #753, opened 2017-03-16
runtime: check if cc-shim is running before stop it

devimc:
This patch is to avoid failures when cor tries to
stop cc-shim when it was already stopped

### DIFF
--- a/src/oci.c
+++ b/src/oci.c
@@ -378,7 +378,7 @@ cc_oci_kill (struct cc_oci_config *config,
 		 * finished with SIGKILL signal hence cc-shim MUST receive
 		 * this signal too
 		 */
-		if (kill (state->pid, signum) < 0) {
+		if (kill (state->pid, signum) < 0 && errno != ESRCH) {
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

src/oci.c
```diff
@@ -378,7 +378,7 @@ cc_oci_kill (struct cc_oci_config *config,
 		 * finished with SIGKILL signal hence cc-shim MUST receive
 		 * this signal too
 		 */
-		if (kill (state->pid, signum) < 0) {
+		if (kill (state->pid, 0) == 0 && kill (state->pid, signum) < 0) {
```
jodh-intel: It would be more conventional (and saves a syscall) to write this as:

```
if (kill (state->pid, signum) < 0 && errno != ESRCH) {
```

devimc: Hi @jodh-intel 
you're right , thanks :)

dlespiau:
lgtm

jodh-intel:
lgtm

PR #761, opened 2017-03-22
Add check for hook path.

amshinde:


chavafg:
qa-passed

src/process.c
```diff
@@ -287,6 +287,11 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 		return false;
 	}
 
+	if (! hook->path) {
```
jodh-intel: Good catch. Could you add a test for this scenario?

amshinde: @jodh-intel  I checked and there exists a test already to handle this. Looking further, the spec handler for hooks requires path to be present in the config json to complete successfully.
So this is a redundant check.

jodh-intel: Wait a sec - having looked more closely here, this change is bogus: `hook->path` is an array, so must be valid if `hook` is. This is alluded to below where we check that the first element is the expected slash to denote a full path.

(I don't see the test you are referring to either. We could do with a new test for the case where `hook->path[0] = '\0'`).

devimc: @jodh-intel  you're right (```hook->path[0] == '\0'```)

PR #774, opened 2017-03-24
Fix cc-oci-runtime list for pod containers

sameo:
cc-oci-runtime list would always return `stopped` for pod containers.
Pod containers are shim controlled containers, but they're not associated to a VM, so `cc_oci_vm_running` would always return false on them.

This PR generalizes `cc_oci_vm_running` to `cc_oci_container_running` and consider both VM linked containers (Docker containers and pod sandbox containers) as well as pod containers.

### DIFF
--- a/src/mount.c
+++ b/src/mount.c
@@ -243,7 +243,7 @@ cc_handle_mounts(struct cc_oci_config *config, GSList *mounts, gboolean volume)
 			continue;
 		}
 
-		if ((! cc_pod_is_vm(config) || cc_pod_is_sandbox(config)) && volume) {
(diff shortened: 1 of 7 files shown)


jodh-intel:
Nice!

lgtm.

sboeuf:
Sounds good, LGTM

chavafg:
qa-passed

grahamwhaley:
@sboeuf  - to note, I believe your L G T M will only register with pullapprove if it is at the front of (or possibly on its own) line :-)

sboeuf:
@grahamwhaley thanks for the tip, I didn't notice ;)

sboeuf:
LGTM

PR #775, opened 2017-03-24
Makefile: Add build hardening with -fPIE flag

sboeuf:
Adding the -fPIE -pie flags in order to harden the security. It creates our binary as a Position Independant Executable, and takes advantage of ASLR, making ROP attacks much harder to perform.

See https://wiki.debian.org/Hardening for further details.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -137,7 +137,10 @@ AM_CFLAGS = -std=gnu99 -fstack-protector -Wall -pedantic \
 	-Wstrict-prototypes -Wundef -fno-common \
 	-Werror-implicit-function-declaration \
 	-Wformat -Wformat-security -Werror=format-security \
-	-Wconversion -Wunreachable-code
+	-Wconversion -Wunreachable-code \
+	-fPIE
+
+AM_LDFLAGS = -pie
 
 # We set --with-systemdunitdir here so make distcheck can run make install as a
 # normal user and not fail.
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

dlespiau:
Seems like the build with clang is failing and needs to be looked at.

sboeuf:
@dlespiau yes I have noticed clang build is not passing. It is due to following warnings considered as errors:
clang: warning: argument unused during compilation: '-pie'

Any idea how to properly solve this ?

dlespiau:
We can always adapt the flags depending on the compiler we configure the compilation with (CC=gcc or CC=clang). `-fPIE -pie` for gcc, `-fPIE` for clang.

sboeuf:
Yes makes sense. Supposed to do that from configure.ac ?

dlespiau:
Yes, configure.ac. Test CC in configure.ac, set a variable, eg. PIE_CFLAGS, to either `-fPIE -pie` or `-fPIE` and call AC_SUBST on it. USe $(PIE_CFLAGS) in Makefile.am. Turns out -pie seems to be a linker option. Not sure if that is needed in the link phase with clang.

sboeuf:
@dlespiau I think I have just figured this out. No need to make a check according to the compiler used because clang actually support -pie link option. The warning/error message was about wrong use of -pie which is a link option as a compiler option.

sboeuf:
@dlespiau build is passing now ;)

chavafg:
qa-passed

sboeuf:
@grahamwhaley @dlespiau maybe LGTM this one ?

grahamwhaley:
Yep, given the qa-passed and CI passed, and we are pretty sure about the -pie being a linker only cmd,
lgtm

dlespiau:
I've checked locally and the resulting binaries seem to be relocatable and ready for ASLR.

dlespiau:
lgtm? (maybe editing doesn't trigger pullaprove?)

PR #776, opened 2017-03-24
Compatibility with docker 17

devimc:


chavafg:
qa-passed

src/oci.c
```diff
@@ -60,6 +60,9 @@
 #include "pod.h"
 #include "namespace.h"
```
amshinde: Pertaining to the commit message itself,  I think the docker flow in 1.12.x was the same,(create->start->kill->delete), delete was called if the signal itself was a terminating signal i.e if the shim to be terminated.  I think what has changed is that docker now expects all processes (shim+VM) to be terminated for the delete to be called. At least this was my understanding regarding the change from earlier discussions.

src/oci.c
```diff
@@ -372,8 +404,8 @@ cc_oci_kill (struct cc_oci_config *config,
 		return true;
 	}
 
-	/* stopping container */
-	config->state.status = OCI_STATUS_STOPPING;
+	/* signaling container, not all signals stop the container */
+	config->state.status = OCI_STATUS_SIGNALING;
```
amshinde: Nice, I always thought it was a bug, that we mark the container as stopping for any signal.
Is the string for OCI_STATUS_SIGNALING part of the OCI spec?

src/oci.c
```diff
@@ -382,8 +414,7 @@ cc_oci_kill (struct cc_oci_config *config,
 	}
 
 #ifndef UNIT_TESTING
-	/* kill container's processes, cc-shim will catch exit code
-	 * and will end
+	/* kill container's processes if they are still running
```
amshinde: Why has this comment changed, I think the earlier commit is still relevant

src/oci.c
```diff
@@ -394,31 +425,72 @@ cc_oci_kill (struct cc_oci_config *config,
...
-		 */
-		if (kill (state->pid, signum) < 0 && errno != ESRCH) {
-			g_critical ("failed to stop cc-shim %u: %s",
-					(unsigned)state->pid,
-					strerror (errno));
-			return false;
+	int tries = 1;
+	/* termination signals */
+	const int term_signals[] = {
+		SIGTERM, SIGINT, SIGQUIT, SIGKILL, SIGHUP, -1
```
amshinde: One cant really say that the signals listed here besides SIGKILL are termination signals, they may very well be handled by the container process in the VM to not exit or be masked completely. SIGHUP may just cause configuration for a process to just be reloaded. 
One can say that a signal is terminating, if the shim itself has received exit code from the container process and exited.

src/oci.c
```diff
@@ -394,31 +425,72 @@ cc_oci_kill (struct cc_oci_config *config,
...
+	 */
+	if (kill (state->pid, signum) < 0 && errno != ESRCH) {
+		g_critical ("failed to stop cc-shim %u: %s",
+				(unsigned)state->pid,
+				strerror (errno));
+		goto error;
+	}
 
-	config->state.status = OCI_STATUS_STOPPED;
+	for (int i=0; i<tries; i++) {
```
amshinde: I dont particularly like the idea of looping here. I thought we had a cleaner mechanism implemented earlier to check if shim has exited.

src/oci.c
```diff
@@ -60,6 +60,9 @@
 #include "pod.h"
 #include "namespace.h"
 
+#define MAX_TRIES_KILL 5
+#define MAX_SLEEP_KILL 100000
```
amshinde: Would be useful to add comments for these #defines are used for.

src/oci.c
```diff
@@ -394,31 +425,72 @@ cc_oci_kill (struct cc_oci_config *config,
...
+			if (! cc_oci_state_file_create (config, state->create_time)) {
+				g_critical ("failed to recreate state file");
+				goto error;
+			}
+			break;
+		}
+		/* This sleep is needed since cc-oci-runtime is not
+		 * the reaper of cc-shim
+		 */
+		usleep (MAX_SLEEP_KILL);
```
amshinde: Using sleep does not seem clean. Why is this sleep required, at this point the state would have been updated if the signal was a terminating signal.

src/oci.c
```diff
@@ -954,7 +1002,7 @@ cc_oci_start (struct cc_oci_config *config,
 
 		/* If the VM was stopped then *do not* cleanup */
 		if (config->state.status != OCI_STATUS_STOPPED &&
-			config->state.status != OCI_STATUS_STOPPING) {
+			config->state.status != OCI_STATUS_SIGNALING) {
```
amshinde: Do we need the second check?

src/oci.h
```diff
@@ -143,7 +143,7 @@ enum oci_status {
 	OCI_STATUS_RUNNING,
 	OCI_STATUS_PAUSED,
 	OCI_STATUS_STOPPED,
-	OCI_STATUS_STOPPING,
+	OCI_STATUS_SIGNALING,
```
amshinde: Again is this value part of OCI spec? Might be useful to have this in its own commit with explanation for introducing this value.

tests/integration/docker/exit-code.bats
```diff
@@ -47,3 +47,11 @@ teardown() {
 	[ "${status}" -eq "$exit_status" ]
 	$DOCKER_EXE rm -f $container
 }
+
+@test "Verifying exit code with docker ps" {
+	container=$(random_name)
+	run $DOCKER_EXE run --name $container -ti ubuntu bash -c "exit $exit_status"
+	[ "${status}" -eq "$exit_status" ]
+    run $($DOCKER_EXE ps -a | grep $container | grep "Exited ($exit_status)")
```
amshinde: Indentation needs to be fixed.

shim/shim.c
```diff
@@ -94,30 +94,6 @@ signal_handler(int signal_no)
...
 
-/*!
- * Assign signal handler for all the signals that should be
- * forwarded by the shim to the proxy.
- *
- * \param sa Signal handler
- * \return true on success, false otherwise
- */
-bool
-assign_all_signals(struct sigaction *sa)
```
devimc: cc-oci-runtime will send signals to hyper directly because '--all' option was added to kill command hence cc-shim doesn't know when kill just one process or when kill all processes

src/oci.c
devimc: nop, but according to oci spec we can add new states
```
Additional values MAY be defined by the runtime, however, they MUST
be used to represent new runtime states ...
```

chavafg:
qa-failed

src/oci.h
devimc: no, it isn't

src/oci.c
devimc: yes, because we do not know what signal was send to the container

src/oci.c
devimc: because we will send the signal to cc-shim either way

src/oci.c
devimc: I agree, but cc-oci-runtime is not the reaper of cc-shim so cc-oci-runtime has to decide if it must kill the container or not

chavafg:
qa-failed

chavafg:
qa-failed

src/process.c
```diff
@@ -657,25 +658,29 @@ cc_shim_launch (struct cc_oci_config *config,
...
+		args[i++] = g_strdup_printf ("%d", proxy_io_base);
 		if ( ! config->oci.process.terminal) {
-			args[9] = g_strdup ("-e");
-			args[10] = g_strdup_printf ("%d", proxy_io_base + 1);
+			args[i++] = g_strdup ("-e");
+			args[i++] = g_strdup_printf ("%d", proxy_io_base + 1);
+		}
+		if (initial_workload && ! cc_pod_is_pod_container (config)) {
+			/* cc-shim will destroy the VM when initial workload ends */
+			args[i++] = g_strdup ("-w");
```
jodh-intel: It's not strictly part of this PR, but now you've improved the code to use a variable rather than hard-coded index values, it would be a good idea to add a check to ensure that `i <= 12` (or even better add a `#define SHIM_ARG_COUNT 12` and then ensure `i <= SHIM_ARG_COUNT`) to avoid any possibility of an overrun for any future changes.

chavafg:
qa-passed

src/process.c
devimc: @jodh-intel fixed
thanks

jodh-intel:
lgtm.

src/process.c
```diff
@@ -683,6 +685,12 @@ cc_shim_launch (struct cc_oci_config *config,
 			args[i++] = g_strdup ("-w");
 		}
 
+		if (i > SHIM_ARG_COUNT) {
+			g_critical("args index exceeds SHIM_ARG_COUNT: %d > %i",
+				SHIM_ARG_COUNT, i);
```
dlespiau: shouldn't it be `"...: %d > %d", i, SHIM_ARG_COUNT);` ? (btw, %d and %i are the same for printf)

src/oci.c
```diff
@@ -356,78 +355,65 @@ cc_oci_kill (struct cc_oci_config *config,
...
+			config->state.status = OCI_STATUS_STOPPED;
+			/* update state file */
+			if (! cc_oci_state_file_create (config, state->create_time)) {
+				g_critical ("failed to recreate state file");
+				return false;
+			}
+		}
+	} else {
+		/* is VM running ?*/
+		if (kill (config->vm->pid, 0) == 0 &&
```
amshinde: You dont know if it is terminating signal. Doesnt seem right to stop the VM here. You should be doing this only for sigkill and sigterm.

src/process.c
devimc: oops yep, fixing

src/oci.c
amshinde: After taking a second look  realize that it makes sense to kill the VM if the shim is not running.

shim/shim.c
```diff
@@ -534,25 +536,27 @@ parse_numeric_option(char *input) {
...
-        printf("  -d,  --debug          Enable debug output\n");
-        printf("  -h,  --help           Display this help message\n");
+        printf("  -c,  --container-id     Container id\n");
+        printf("  -p,  --proxy-sock-fd    File descriptor of the socket connected to cc-proxy\n");
+        printf("  -o,  --proxy-io-fd      File descriptor of I/0 fd sent by the cc-proxy\n");
+        printf("  -s,  --seq-no           Sequence no for stdin and stdout\n");
+        printf("  -e,  --err-seq-no       Sequence no for stderr\n");
+        printf("  -d,  --debug            Enable debug output\n");
+        printf("  -h,  --help             Display this help message\n");
+        printf("  -w,  --initial-workload This instance represents the initial workload and will destroy the VM when it finish\n");
```
amshinde: s/finish/finishes/

src/oci.c
```diff
@@ -356,78 +355,65 @@ cc_oci_kill (struct cc_oci_config *config,
...
+		{
+			if (kill (state->pid, signum) < 0 && errno != ESRCH) {
+				g_critical ("failed to stop cc-shim %u: %s",
 					(unsigned)state->pid,
 					strerror (errno));
-			return false;
+				return false;
+			}
+		}
+		if (signum == SIGKILL || signum == SIGTERM) {
```
amshinde: Might be useful to add a comment pointing to the docker code where it lists these signals as terminating.

chavafg:
qa-passed

chavafg:
qa-failed

chavafg:
qa-failed

shim/shim.c
```diff
@@ -455,6 +454,9 @@ handle_proxy_output(struct cc_shim *shim)
 		shim->exiting = true;
 		goto out;
 	} else if (shim->exiting && stream_len == (STREAM_HEADER_SIZE+1)) {
+		if (shim->initial_workload) {
+			send_proxy_hyper_message(shim->proxy_sock_fd, "destroypod", "\"\"");
```
amshinde: Does this work with CRIO with multiple containers in a pod?

chavafg:
qa-failed

shim/shim.c
devimc: yes, it's working

chavafg:
qa-passed

chavafg:
qa-passed

jodh-intel:
lgtm

jcvenegas:
lgtm , nice work @devimc :D probably we would like to document the current kill process somewhere, at least the expected behavior  for pods and docker containers.

amshinde:
lgtm
Agree with @jcvenegas.

amshinde:
@sameo Can you take a look at this PR as well since it involves crio changes?

sameo:
LGTM

PR #778, opened 2017-03-27
configure.ac: Add -D_FORTIFY_SOURCE=2 optimization

sboeuf:
In order to harden the build and the overall security of the project, this patch enables FORTIFY_SOURCE optimization in case the flag --enable-debug is not detected.

Indeed, when debug is enabled, the -O flag is set to -O0, and the option FORTIFY_SOURCE expect at least -O1 to be properly enabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,6 +75,9 @@ AS_IF([test "x$enable_debug" = "xyes"],
        LDFLAGS=`echo "$LDFLAGS" | sed -e "s/ -g\b/ -ggdb3/g"`
       ]])
 
+AS_IF([test "x$enable_debug" = "xno"],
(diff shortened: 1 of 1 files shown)


sboeuf:
@dlespiau could you take a look ?

chavafg:
qa-passed

amshinde:
lgtm

grahamwhaley:
I'm not sure I'd call it an 'optimisation' - more of a 'feature' :-) , but - nice!
lgtm

sboeuf:
@grahamwhaley I called it optimization because it is a compiler optimization that can only be used if at least -O1 is enabled.

PR #779, opened 2017-03-27
Makefile: Ensure noexecstack and relro are enabled

sboeuf:
As we want to increase the security of the project, we want to make sure we have both noexecstack and relro flags enabled. By default on several distributions, gcc will enable those flags, but we don't want to end up with a case where those flags are not enabled.

About the details of these flags, noexecstack protects the stack from being executed, while relro protects against data relocation.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -140,7 +140,7 @@ AM_CFLAGS = -std=gnu99 -fstack-protector -Wall -pedantic \
 	-Wconversion -Wunreachable-code \
 	-fPIE
 
-AM_LDFLAGS = -pie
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

grahamwhaley:
@sboeuf I see you also added 'now'. Can you add a comment about that to the commit message as well please? Other than that,
lgtm

chavafg:
qa-passed

sboeuf:
@amshinde @dlespiau could you lgtm this one if you agree with ?

amshinde:
lgtm

Issue #152, opened 2016-08-22
networkstatic/iperf3 containers fails to run with cors


mcastelino:
If you try to run iperf3 with runc as the runtime it works. 
However if the same container is run with cor as a runtime it will fail with the error

`iperf3: error - unable to create a new stream: No such file or directory`

This may have something to do with the setup of /tmp or an environment variable in cor.

jodh-intel:
This is looking like a 9p bug. I created a test program (attached) that makes the following calls (similar to what `iperf_new_stream()` in iperf3 is doing. The program works fine in a normal environment, but fails in a CC on the call to `ftruncate()`. Here's some sample strace output:

```
execve("./test_iperf3-manohar", ["./test_iperf3-manohar"], [/* 9 vars */]) = 0
uname({sysname="Linux", nodename="7e55c4113e73", ...}) = 0
brk(NULL)                               = 0x186b000
brk(0x186c1c0)                          = 0x186c1c0
arch_prctl(ARCH_SET_FS, 0x186b880)      = 0
readlink("/proc/self/exe", "/shared/test_iperf3-manohar", 4096) = 27
brk(0x188d1c0)                          = 0x188d1c0
brk(0x188e000)                          = 0x188e000
fstat(1, {st_mode=S_IFREG|0644, st_size=468, ...}) = 0
gettimeofday({1471958259, 200678}, NULL) = 0
getpid()                                = 360
open("/tmp/foo.O2LCdw", O_RDWR|O_CREAT|O_EXCL, 0600) = 3
unlink("/tmp/foo.O2LCdw")               = 0
ftruncate(3, 0)                       
[test_iperf3-manohar.c.txt](https://github.com/01org/cc-oci-runtime/files/432478/test_iperf3-manohar.c.txt)
  = -1 ENOENT (No such file or directory)
write(2, "ERROR: ftruncate(fd=3, 0) failed"..., 42ERROR: ftruncate(fd=3, 0) failed: errno=2
) = 42
write(1, "DEBUG: template='/tmp/foo.XXXXXX"..., 115DEBUG: template='/tmp/foo.XXXXXX'
SUCCESS: mkstemp(/tmp/foo.O2LCdw) worked
SUCCESS: unlink(/tmp/foo.O2LCdw) worked
) = 115
exit_group(1)                           = ?
+++ exited with 1 +++
```

A workaround is to run the following in the container before starting the app:

```
# mount -t ramfs -o size=20M ramfs /tmp
```

[...]

PR #781, opened 2017-03-29
Metrics: Network Measurement using Iperf

GabyCT:
This script will measure network bandwidth, jitter, network bandwidth
for multiple TCP connections, UDP network bandwidth and bi-directional
network bandwidth using iperf in a local setup (1 client - 1 server in
the same hardware).

### DIFF
--- a/tests/metrics/network/network-metrics.sh
+++ b/tests/metrics/network/network-metrics.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Port number where the server will run
+port=5001:5001
+# Image name
+image=gabyct/network
+# Measurement time (seconds)
+time=5
+
+function setup {
+        runtime_docker
+        kill_processes_before_start
+}
+
+# This script will perform all the measurements using a local setup
+
+function bandwidth {
+        setup
+        bandwidth_result=$(mktemp)
+        $DOCKER_EXE run -d --name=iperf-server ${image} bash -c "iperf -p ${port} -s" > /dev/null && \
+	server_address=`$DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql)` && \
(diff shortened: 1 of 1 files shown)


chavafg:
qa-passed

grahamwhaley:
Looking pretty good to me. A couple of notes:
- we are using an image from a user on the docker hub, which is probably fine for now, but we should be aware that maybe that will go away one day, and we should have our own image setup
- that image is using iperf2 I think. I wonder if iperf3 is actually better - @mcastelino any thoughts on that?
- I ran the test locally and the UDP test bandwidth is 3000x less than the TCP (1mbit vs 3gbit) - is that expected, do you see that locally as well etc. Just feels like a big gap, so would be good if we verified that is the expected result.

otherwise,
lgtm

GabyCT:
@grahamwhaley the main reason why we are using iperf2 is because iperf3 is not working with cc-oci-runtime (issue #152). On the other hand, I saw the same behaviour in UDP (locally).

egernst:
I see same delta in performance when I run TCP v UDP locally as well.  I expect this is a result of packet sizes being used in tcp (1500) vs in UDP (not sure what's used, but guessing much smaller, individual packets).  Perhaps Manohar can comment here.  Also depends if we are getting packet drops, etc.

mcastelino:
@GabyCT @grahamwhaley We should use our own image.
iperf2 has one feature that is missing in iperf3 which is the reason I still want this to be used for that one test case. So we need both iperf3 and 2.

> Bi-directional simultaneous (test the speed both ways at the same time)
> Use the -d switch to test in the network bandwidth in both directions. This will perform two tests; one from local host to remote host, and another from the remote host to the local host

We use/should test this. 

iperf3 is better at UDP testing as iperf2 does not reliably support testing UDP beyond 2Gbps

> Note that iperf 2.x results are inconsistent for UDP above 2Gbps. We recommend using nuttcp or iperf3 for high-speed UDP testing.


@GabyCT can you use iperf UDP test itself to measure jitter vs ping as this allows for better control

> The UDP tests with the -u argument 

@egernst @GabyCT 

If we really want to measure UDP bandwidth then you need to scale it up till we see loss with iperf3.
But that will require that this script be modified to perform a scaled search.

Better yes let us switch to 

> nuttcp 

for UDB bandwidth testing

GabyCT:
@mcastelino, I did the implementation in this script for iperf2 and create an image in docker hub. Currently, this script measures bi-directional (local-remote) and (remote-local), also it is measuring jitter using -u option, bandwidth and multiple tcp connections. @mcastelino, @grahamwhaley, @egernst, I will create another PR for the implementation of iperf3 and the use of nuttcp.

egernst:
Sounds good to me. Thanks!

chavafg:
qa-passed

chavafg:
qa-passed

grahamwhaley:
Just a check then - are we waiting for changes to this PR, or are we going to merge this one and then add improvements on top? @GabyCT @egernst

GabyCT:
@grahamwhaley improvements and new tests will be on top

grahamwhaley:
OK, then given the 'promise' we will do updates (maybe you could open an Issue for that @GabyCT ?),
lgtm

GabyCT:
@grahamwhaley done :) #791

amshinde:
lgtm

grahamwhaley:
Hi @mcastelino @GabyCT - (comment on old thread, which may lead us to open a new issue...) - now we have iperf3 and nuttcp tests in place, I get the feeling that all but one of the iperf2 tests/measures are probably now redundant and outdated? If so, can we state which iperf2 test is the one we need to keep, and then rework the iperf2 script to remove (or disable by default) all the tests we don't need any more?
It is quite confusing having multiple tests that report say jitter or UDP bandwidth *if* they are duplications (that is, unless they add something unique - in which case I think we need to document that much more clearly).

mcastelino:
@grahamwhaley @GabyCT The only iperf2 test we need to retain is the directional test where both ends act as server and client. All remaining tests should be run just with iperf3. There is no value running them with iperf2.

Bidirectional testing (-d / -r)

GabyCT:
@mcastelino and @grahamwhaley so I will submit a PR where I will remove the other tests and only retain the bidirectional test

jodh-intel:
Related: #848.

Issue #760, opened 2017-03-21
Telemetry crash related to cc_run_hook()


sboeuf:
Telemetry crash 774.
Something wrong happened after the runtime forked to do a re-exec of the OCI hook provided.

sboeuf:
@devimc I have assigned this to you as you're driving cc-oci-runtime. Feel free to assign it to someone else who could take some time to investigate this.
@dlespiau any ideas what could be the reason for such kind of failure ?

jodh-intel:
@sboeuf - Are there any associated logs to help with debugging here (docker, bats, runtime, etc)?

I'm only seeing a summary of the backtrace, which only seems tho show that a crash occurred in `cc_run_hook()`, but not exactly where nor what the inputs were.

sboeuf:
@jodh-intel I agree this is really complicated to understand the root cause. I have asked about test setup or extra logs we could use to help us, but apparently there is no such thing. Telemetry seems to report only crashes but we cannot know what was the reason for this...

icarus-sparry:
The only obvious call to abort() in the path including cc_run_hooks() is the one in cc_run_hook() (Presumably inlined as it is static and only has one call site).  This will be called if the system fails to execvpe the program given by hook->path.

(Failures to execvpe are usually caused by the program not being on the PATH, but can be due to missing resources. symbolic link loops, I/O errors and so on)

The code does do a call to g_critical to log an error message in the child process. 

I think it would make more sense to just call exit() with the value of errno rather than abort(). The parent already has code to handle the process failing. The core does draw attention to the issue, which is good.

amshinde:
There are a few crashes originating from my machine in the telemetry logs. I had changed my telemetry machine-id to a recognizable name a few days back to track any crashes. Comparing the pids in the telemetry logs to the cc-oci-runtime logs, I can say that the issue was with exec-ing the hooks, specifically :
"failed to exec hook /usr/bin/dockerd (deleted): No such file or directory"

This was caused due to missing files during the the container bundle split (containers-basic and containers-virt), causing certain files to not be deleted.

I do agree with @icarus-sparry, it would make sense to exit() with errno rather than calling abort() here.

sboeuf:
@sameo @jodh-intel @dlespiau @grahamwhaley What do you think about that ? We can simply get rid of every abort() in the code and replace them with exit(errno) ?

jodh-intel:
[...]

PR #784, opened 2017-03-29
hooks: If the exec call for a hook fails, exit with failure.

amshinde:
If the execvpe call for invoking hook fails, we log the failure.
We should exit cleanly with non-zero exit status here, instead of
calling abort here which causes a core dump.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -282,6 +282,7 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 	char c = 0;
 	int status = 1;
 	char **args = NULL;
+	int saved_errno = 0;
 
 	if (! hook) {
 		return false;
@@ -319,11 +320,13 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 
 		/* waiting for parent setup */
 		if (read (pipe_parent_error[0], &c, sizeof(c)) > 0) {
+			saved_errno = errno;
 			g_critical ("parent setup failed");
 			goto fail_child;
 		}
 
 		if (dup2 (stdin_pipe[0], STDIN_FILENO) < 0) {
+			saved_errno = errno;
 			g_critical ("failed to dup hook stdin");
 			goto fail_child;
 		}
@@ -335,6 +338,7 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 		}
 
 		if (execvpe (hook->path, args, hook->env) < 0) {
+			saved_errno = errno;
 			g_critical ("failed to exec hook %s: %s",
 					hook->path, strerror (errno));
 		}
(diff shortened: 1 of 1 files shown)


chavafg:
qa-failed

chavafg:
there was an unusal failure:	
```
not ok 6 quick ping among replicas on their overlay ip 
# (from function `setup' in test file swarm.bats, line 53) 
# `$DOCKER_EXE swarm init ${swarm_interface_arg}' failed 
# Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again. 
# Error response from daemon: error updating cluster settings: rpc error: code = 2 desc = update out of sequence 
# Error response from daemon: service testswarm not found 
# ID HOSTNAME STATUS AVAILABILITY MANAGER STATUS 
# 66p90amjh8m5u7w8pllihjex0 * ubuntu-cc-ci Ready Active Leader 
# Node left the swarm. 
# Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again.
```
Retesting...

chavafg:
qa-passed

grahamwhaley:
@amshinde would this be for fixing #760 ?  If so, pulling in @icarus-sparry for comment/review as well.
just FMI, does the OCI spec have anything to say on how to handle 'missing hooks' at all?

src/process.c
```diff
@@ -347,7 +347,7 @@ cc_run_hook(struct oci_cfg_hook* hook, const gchar* state,
 		close_if_set (stdin_pipe[0]);
 		close_if_set (pipe_child_error[1]);
 		close_if_set (pipe_parent_error[0]);
-		abort();
+		exit(EXIT_FAILURE);
```
icarus-sparry: I would prefer to call exit() with the reason so there is more discrimination on why the hook failed, rather than just "it failed". Of course errno can be altered by things like the close_if_set() calls, so really need to create a new variable and assign it in all the failure paths to capture the first failure.

So for example put an assignment before line 338 - the call to g_critical.

The idea is sound - just give a little more info.

amshinde: @icarus-sparry you mean something like exit(saved_errno) ?

icarus-sparry: yes, exactly exit(saved_errno).

amshinde: @icarus-sparry I have pushed an updated commit. Please have a look.

chavafg:
qa-passed

devimc:
lgtm

grahamwhaley:
Heh, just fyi for @icarus-sparry - your LGTM appreciated :-) - we use 'pullapprove' to verify some of our required checks/info, including requiring acks from at least two folks from the formal 'list of reviewers', not including the author of the PR :-)

dlespiau:
lgtm

jodh-intel:
Hi @icarus-sparry - further to @grahamwhaley's comments, you can find further details on the wiki here: https://github.com/01org/cc-oci-runtime/wiki/Release-Process#pr-reviews

PR #804, opened 2017-04-07
Add 'make metrics-tests' rule to execute all metrics

chavafg:
This PR contains 3 commits:

- metrics - Add Makefile rule to run metrics tests
This new Makefile rule will run all metrics tests in
tests/metrics/run_docker_metrics and will generate the files
needed to execute them

- metrics doc - update metrics README
This change documents how to execute the new rule to
execute the metrics tests.

- metrics - abort metrcis tests if generated files are missing
This commit will verify if the test/lib/test-common.bash is
already generated before running the metrics tests. It will
also tell you to use the Makefile rule 'make metrics-test' so
the missing files are generated.

### DIFF
--- a/tests/metrics/README.md
+++ b/tests/metrics/README.md
@@ -4,19 +4,21 @@
 Some tests require you to enable the debug mode of cc-oci-runtime. To enable it, please take a look at:
 https://github.com/01org/cc-oci-runtime#18running-under-docker
 
-To run the metrics, just run:
+
+### Run all metrics tests
+To run the metrics, please use the Makefile rule `metrics-tests` on the top level of this repository:
 
 ```bash
-$ cd tests/metrics
-$ ./run_docker_metrics
+$ sudo -E PATH=$PATH make metrics-tests
 ```
 
 This will run all metrics tests and generete a `results` directory.
 
-Each file in the `result` directory contains the result for a test and has the format of a CSV.
+Each file in the `results` directory contains the result for a test as a comma separated values or CSV.
 At the end of each file you will find the result average of all the data collected by the test.
 
-You can also run each tests script separately. e.g.
+Execute `sudo -E PATH=$PATH make metrics tests` to generate the necessary files to run individual tests.
+For example:
 
 ```bash
 $ cd tests/metrics
--- a/tests/metrics/run_docker_metrics
+++ b/tests/metrics/run_docker_metrics
@@ -1,23 +1,19 @@
 #!/bin/bash
 
 source ./run_docker_metrics.dat
+
+# Verify that test-common.bash is generated.
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+[ -f "${SCRIPT_PATH}/../lib/test-common.bash" ] || \
+	{ echo "Required files are not generated, please run 'sudo make metrics-tests' from top level of this repo" \
+	&& exit 1; }
+
+
 RESULT_DIR="./results"
 
 if [ ! -d "$RESULT_DIR" ]; then
 	mkdir "$RESULT_DIR"
 fi
-
-IN_FILES=(workload_time/cor_create_time.sh.in)
-
-#generate files if they do not exist
-for inf in ${IN_FILES[@]}; do
-	outf="`dirname \"${inf}\"`/`basename -s '.in' \"${inf}\"`"
-	if [ ! -f "${outf}" ]; then
-		PACKAGE_NAME=cc-oci-runtime \
-		../../data/genfile.sh "${inf}" "${outf}"
-		chmod +x "${outf}"
-	fi
-done
 
 # complete workload : docker run --runtime $runtime -tid $image $cmd
 bash workload_time/docker_workload_time.sh true ubuntu runc "$TIMES"
@@ -37,5 +33,5 @@ bash density/docker_cpu_usage.sh "$TIMES" "$CPU_WAIT_TIME"
 bash density/docker_memory_usage.sh "$MEM_CONTAINERS" "$MEM_WAIT_TIME"
 
 # kernel boot time
-bash workload_time/kernel_boot_time.sh $TIMES
+bash workload_time/kernel_boot_time.sh "$TIMES"
 bash workload_time/kernel_boot_time_stress.sh
--- a/Makefile.am
+++ b/Makefile.am
@@ -411,6 +411,15 @@ crio-tests: cc-oci-runtime cc-proxy cc-shim tests/lib/test-crio.bats
 	@timeout 4m $(BATS_PATH) -t $(srcdir)/tests/integration/cri-o
 endif
 
+if METRICS_TESTS
+METRICS_PATH := "$(abs_top_srcdir)/tests/metrics"
+metrics-tests: cc-oci-runtime cc-proxy cc-shim $(GENERATED_FILES)
+	@pushd $(METRICS_PATH); \
+	./run_docker_metrics; \
+	popd ; \
+	echo "Metrics results can be found on $(METRICS_PATH)/results"
+endif
+
 #### tests ####
 if BUILD_TESTS
 
--- a/configure.ac
+++ b/configure.ac
@@ -122,6 +122,19 @@ AS_IF([test x"$run_docker_tests" = x"yes"],
     [AC_SUBST([DOCKER_TESTS], [0])])
 AM_CONDITIONAL([DOCKER_TESTS], [test x"$run_docker_tests" = x"yes"])
 
+# Metrics tests
+AC_ARG_ENABLE(metrics-tests, AS_HELP_STRING([--disable-metrics-tests],
+              [disable metrics tests @<:@default=no@:>@]))
+
+AS_IF([test x"$enable_metrics_tests" = x"yes" -o x"$enable_metrics_tests" = x""],
+    [run_metrics_tests=yes],
+    [run_metrics_tests=no])
+
+AS_IF([test x"$run_metrics_tests" = x"yes"],
+    [AC_SUBST([METRICS_TESTS], [1])],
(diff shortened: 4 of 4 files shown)


chavafg:
qa-passed

jodh-intel:
lgtm

tests/metrics/run_docker_metrics
```diff
@@ -1,24 +1,19 @@
 #!/bin/bash
 
 source ./run_docker_metrics.dat
+
+# Verify that test-common.bash is generated.
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+[ -f "${SCRIPT_PATH}/../lib/test-common.bash" ] || \
```
grahamwhaley: Hi. Can you double check this file check line for me. It seems to always exit on my system (ubuntu under ciao-down) and doesn't run the tests (when invoked with the `make metrics-tests` from the top level dir).
To check I ran the script in the subdir with -x. Output:

```
gwhaley@singlevm:/work/gwhaley/gopath/src/github.com/01org/cc-oci-runtime/tests/metrics$ sudo bash -x ./run_docker_metrics
+ source ./run_docker_metrics.dat
++ TIMES=100
++ MEM_CONTAINERS=10
++ MEM_WAIT_TIME=30m
++ CPU_WAIT_TIME=2m
+++ readlink -f ./run_docker_metrics
++ dirname /work/gwhaley/gopath/src/github.com/01org/cc-oci-runtime/tests/metrics/run_docker_metrics
+ SCRIPT_PATH=/work/gwhaley/gopath/src/github.com/01org/cc-oci-runtime/tests/metrics
+ '[' -f /work/gwhaley/gopath/src/github.com/01org/cc-oci-runtime/tests/metrics/../lib/test-common.bash ']'
+ exit 1
gwhaley@singlevm:/work/gwhaley/gopath/src/github.com/01org/cc-oci-runtime/tests/metrics$ ls ../lib
test-common.bash  test-common.bash.in  test-crio.bats  test-crio.bats.in
```

I also checked - it does print the warning and exit when the tests-common.bash file is removed.

I'd probably just open-code the line as an if/fi line :-)

tests/metrics/README.md
```diff
@@ -4,25 +4,19 @@
...
 
-To run the metrics, just run:
+
+### Run all metrics tests
+To run the metrics, please use Makefile rule `metrics-tests` on the top level of this repository:
 
 ```bash
-$ cd tests/metrics
-$ ./run_docker_metrics
+$ sudo make metrics-tests
```
grahamwhaley: Just a note - this does not quite work on my 'ubuntu under ciao-down' instance as 'go' is not available to root, and this make line tries to build the whole source base (from a clean checkout), including the proxy which is in `go`. Just a niggle/annoyance.

tests/metrics/README.md
```diff
@@ -4,25 +4,19 @@
...
 ```
 
 This will run all metrics tests and generete a `results` directory.
 
 Each file in the `result` directory contains the result for a test and has the format of a CSV.
```
grahamwhaley: nitpick not directly related to your patches (but shows up in view)... here `result` should also be `results` as per above. If you do re-work this file then maybe you can fix that as well? :-)

tests/metrics/README.md
```diff
@@ -4,25 +4,19 @@
...
 At the end of each file you will find the result average of all the data collected by the test.
 
-You can also run each tests script separately. e.g.
```
grahamwhaley: I'm OK deleting this section - or alternatively - we could just note that individual tests can be run once `make metrics-tests` has been once to generate the necessary files. I know it's a bit imperfect, but being able to run individual tests is useful.

tests/metrics/README.md
chavafg: Added `sudo` because if user is not part of the docker group, it will not be able to run docker command successfully, but I can remove it and add a note for this. Does that sounds good?

tests/metrics/run_docker_metrics
chavafg: Yes, I missed `( )` to group the echo and the exit command, sorry about that, fixing...

tests/metrics/README.md
chavafg: Sure

tests/metrics/README.md
chavafg: Sounds good

tests/metrics/README.md
grahamwhaley: Sadly, it is slightly more complicated - at least one of the tests (docker_memory_usage.sh) requires to be running as `root` to be able to run `smem` whilst looking at the docker processes it does not own - so, removing the `sudo` I expect will make that test not generate the correct data. Maybe we just make a note saying if the build may fail if `root` does not have a correct environment set up, so you may need to first run as non-root to get a valid build, and then again as root to get a full set of valid results. Not great. @jodh-intel , any thoughts on a better solution - non are immediately springing to my mind.
 And, of course (such as in my case), we don't actually do an `install` step here, so we are actually testing the system installed CC components, not the ones we just built afaik :-)  Maybe the make `metrics-tests' rule should not rely on the CC binaries being built, as it is not actually testing those binaries??

chavafg: This is what works for me: 
```
./autogen.sh --with-qemu-path=/usr/bin/qemu-lite-system-x86_64 --with-cc-image=/usr/share/clear-containers/clear-containers.img --with-cc-kernel=/usr/share/clear-containers/vmlinux.container
make 
sudo make install 
sudo -E make metrics-tests
```
But yes, I have configured `go` to use it with `sudo`.

Currently docker and cri-o tests work with the installed binaries, so the `make install` should be a needed step before running the metrics.

chavafg:
qa-passed

tests/metrics/README.md
jodh-intel: I've raised #811 to capture the need for `root`.

Best practise says we should run everything with the minimum set of privs and only elevate privs when we need to using `sudo(8)`.

@chavafg - We could land this change and then immediately fix #811, but as a minimum we need to document the steps you've outlined there (including the crucial `-E` flag for `sudo(8)`).

tests/metrics/run_docker_metrics
grahamwhaley: Hi again. That still doesn't seem to work for me. I think what is happening is that the (xxx) starts a subshell, so the 'exit 1' exits that subshell only, and not the 'app'. What I think you need to do is replace the () with {} to create 'group command', like:

```
        { echo "Required files are not generated, please run 'sudo make metrics-tests' from top level of this repo" \
        && exit 1; }
```

I tried this, and that does work for me.

tests/metrics/README.md
```diff
@@ -4,19 +4,21 @@
...
 
-To run the metrics, just run:
+
+### Run all metrics tests
+To run the metrics, please use Makefile rule `metrics-tests` on the top level of this repository:
 
 ```bash
-$ cd tests/metrics
-$ ./run_docker_metrics
+$ sudo -E make metrics-tests
```
grahamwhaley: Ah, 'sudo -E' - that feels better. Oddly, it still doesn't actually work for me (even though a `sudo -e echo $PATH` shows the go binary path in there???). I think we can run with this as is for the minute though - it is a step in the right direction.

grahamwhaley:
Oh, and a nitpick if we are doing some edits. In the last commit message s/metrcis/metrics/ :-)

tests/metrics/README.md
jodh-intel: @grahamwhaley - Does a "`sudo -E PATH=$PATH make metrics-tests`" work for you?

grahamwhaley: @jodh-intel it does!!! @chavafg , if we can add that `PATH=$PATH` bit, that'd be good.

chavafg:
@grahamwhaley all changes applied, can you please check?

chavafg:
qa-passed

devimc:
lgtm

grahamwhaley:
Tested-By: me
lgtm

grahamwhaley:
@rcaballeromx some minor docs changes here for you to review :-)

tests/metrics/README.md
```diff
@@ -4,19 +4,21 @@
 Some tests require you to enable the debug mode of cc-oci-runtime. To enable it, please take a look at:
 https://github.com/01org/cc-oci-runtime#18running-under-docker
 
-To run the metrics, just run:
+
+### Run all metrics tests
+To run the metrics, please use Makefile rule `metrics-tests` on the top level of this repository:
```
rcaballeromx: Change to:
... please use the Makefile rule ...

tests/metrics/README.md
```diff
@@ -4,19 +4,21 @@
...
 ```bash
-$ cd tests/metrics
-$ ./run_docker_metrics
+$ sudo -E PATH=$PATH make metrics-tests
 ```
 
 This will run all metrics tests and generete a `results` directory.
 
-Each file in the `result` directory contains the result for a test and has the format of a CSV.
+Each file in the `results` directory contains the result for a test and has the format of a CSV.
```
rcaballeromx: Each file in the `results` directory contains the result for a test as comma separated values or CSV.

tests/metrics/README.md
```diff
@@ -4,19 +4,21 @@
...
 At the end of each file you will find the result average of all the data collected by the test.
 
-You can also run each tests script separately. e.g.
+You will be able to run individual tests after `make metrics-tests` has been exectued (so necessary
```
rcaballeromx: Execute `make metrics-tests` to generate the necessary files needed to run individual tests. For example:

chavafg:
@rcaballeromx thanks for your comments, already updated. Can you please check?

PR #808, opened 2017-04-10
Metrics: Improve readability

GabyCT:
This will try to improve readability of the commands of this script

### DIFF
--- a/tests/metrics/network/network-metrics-iperf3.sh
+++ b/tests/metrics/network/network-metrics-iperf3.sh
@@ -30,6 +30,7 @@ port=5201:5201
 image=gabyct/network
 # Measurement time (seconds)
 time=5
+set -e
 
 function setup {
 	runtime_docker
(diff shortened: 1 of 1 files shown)


tests/metrics/network/network-metrics-iperf3.sh
```diff
@@ -42,10 +42,12 @@ function setup {
 function iperf3_bandwidth {
 	setup
 	bandwidth_result=$(mktemp)
-	$DOCKER_EXE run -d --name=iperf3-server ${image} bash -c "mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s" > /dev/null && \
-	server_address=`$DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql)` && \
-	$DOCKER_EXE run -ti --rm --name=iperf3-client ${image} bash -c "mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c ${server_address} -t ${time}" > "$bandwidth_result"
-	total_bandwidth=`cat $bandwidth_result | tail -n 3 | head -1 | awk '{print $(NF-2), $(NF-1)}'`
+	server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s"
+	$DOCKER_EXE run -d --name=iperf3-server ${image} bash -c "${server_command}" > /dev/null && \
```
devimc: Q.  are ```&&``` needed ?

GabyCT:
@devimc to avoid that the client and the server goes to connection timeout as everything must be synchoronized

tests/metrics/network/network-metrics-iperf3.sh
GabyCT: @devimc to avoid that the client and the server goes to connection timeout as everything must be synchoronized

GabyCT: @devimc a note has been added

devimc:
lgtm

chavafg:
qa-passed

chavafg:
qa-passed

tests/metrics/network/network-metrics-iperf3.sh
grahamwhaley: @GabyCT I'm not sure the && gives you the synchronisation you are hoping for - I don't believe the sub commands 'wait for each other' due to && or are run in parallel - the `bash` manual states I believe they have 'left associativity', and for a command `command1 && command2`, command2 will execute iff command1 succeeds. Thus, && gives you sequence - but so will placing the commands on separate sequential lines. It will catch errors and not run subsequent sub-commands, which is good, but I suspect setting `-e` at the top of the script may actually be a better solution for error handling here.
 I might be missing something here though.
Think I'll ask @jodh-intel if he has any thoughts as well.

jodh-intel:
Hi @GabyCT - I'm in favour of anything that makes the code more readable but I think we can make this simpler still: rather than using the `&& \`, just put a `set -e` at the top of the file. That way you can remove the `&& \`'s (which tend to make the file more difficult to update).

grahamwhaley:
I like the breakout of the commands into strings - makes those easier to read. And I think there is little else you can do to make those docker lines much clearer.
Once we've worked out the && discussion then I'll 'l g t m' it :-)

GabyCT:
@jodh-intel and @grahamwhaley changes applied

jodh-intel:
This is looking much better. Could you just do one final thing: add some blank lines to space out those commands into more logical units (think of them as paragraphs) so that each group of lines represents a action/activity. Something like:

```
server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s"
$DOCKER_EXE run -d --name=iperf3-server ${image} bash -c "${server_command}" > /dev/null

server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c ${server_address} -t ${time}"
$DOCKER_EXE run -ti --rm --name=iperf3-client ${image} bash -c "${client_command}" > "$bandwidth_result"

 total_bandwidth=$(cat $bandwidth_result | tail -n 3 | head -1 | awk '{print $(NF-2), $(NF-1)}')
    :
    :
```

GabyCT:
@jodh-intel done :)

jodh-intel:
@GabyCT - great, thanks.

lgtm.

grahamwhaley:
Tested-By: me
lgtm

PR #812, opened 2017-04-11
proxy: Restrict the socket and parent directory modes

dlespiau:
We don't need those sockets to be read/writable to the whole world. Only
root is enough.


jodh-intel:
lgtm

chavafg:
qa-passed

amshinde:
lgtm

PR #819, opened 2017-04-12
fdleak: Handle nil snapshots in Compare

dlespiau:
When a test fails, we may not generate a fd snapshot and end up in
Compare with nil snapshot(s).

Compare() should handle that case gracefully instead of generating a
panic.


devimc:
lgtm

jodh-intel:
lgtm

PR #821, opened 2017-04-12
docs: update clear containers image version

### LINKED ISSUE #798: docker hello-world hangs up with -t
Running the standard Docker hello-world example (https://hub.docker.com/_/hello-world/) with:
``` docker run -ti hello-world' ```
hangs up (run does not return).
If I then kill the run (has to be done from a separate terminal), and then try to `docker rm -f` the container that was created, that also hangs up.

This was under `ciao-down` (https://github.com/01org/ciao/tree/master/testutil/ciao-down), which installs Ubuntu 16.04.2 LTS, Docker 1.12.1 and Clear Containers 2.1.2

This is similar to the previous ticket #222 - but this hangs up rather than errors out. Note, on #222 2.1 'worked' with `docker run --rm -i hello-world`.  That command runs for me, but I do not see the output from the `hello-world` image, so I would classify that as not working? I would expect the `hello-world` functionality to be the same as when using the `runc` runtime.

@jodh-intel verified a hang up here. @jodh-intel , can you provide some system details if they are different from mine, for completeness.

devimc:
fixes #798


jodh-intel:
lgtm

dlespiau:
Usually, a reason in given in the commit message so we can trace the reasons of the updates in 6 months time :)

lgtm

jodh-intel:
@devimc - yes, is this to fix #798? If so, could you tweak the commit details?

devimc:
@jodh-intel yep

chavafg:
qa-passed

Issue #815, opened 2017-04-11
Failure when trying to run pod create


chavafg:
There seems to be a random issue when trying to launch containers.
As example when running this test: 
```
not ok 27 Kill container with SIGUSR1
# (in test file /home/cc-tester/go/src/github.com/01org/cc-oci-runtime/cc-oci-runtime-2.1.3/tests/integration/docker/kill.bats, line 55)
#   `$DOCKER_EXE ps -a | grep $container | grep "Exited (${exit_code})"' failed
# 9a1dc3f385305255200413d47bd57242591203eb383e864aced317883a9ab712
# Mg51LRvmYTRN1nIXiTQFsY60cw7qUMO4
```

Looking at the logs, cc-oci-runtime fails to run `pod create `
```
2017-04-11T13:17:57.678926Z:21266:cc-oci-runtime:debug:CTL created event: 1
2017-04-11T13:17:57.696571Z:21019:cc-oci-runtime:debug:proxy msg length: 37
2017-04-11T13:17:57.696662Z:21019:cc-oci-runtime:debug:message read from proxy socket: {"success":true,"data":{"version":1}}
2017-04-11T13:17:57.696781Z:21019:cc-oci-runtime:debug:msg received: {"success":true,"data":{"version":1}}
2017-04-11T13:17:57.696957Z:21019:cc-oci-runtime:debug:communicating with proxy
2017-04-11T13:17:57.697004Z:21019:cc-oci-runtime:debug:sending message (length 598) to proxy socket
2017-04-11T13:17:57.697049Z:21019:cc-oci-runtime:debug:writing message data to proxy socket: {"id":"hyper","data":{"hyperName":"startpod","data":{"hostname":"11c29395d10c","containers":[],"shareDir":"rootfs","in
terfaces":[{"macAddr":"02:42:0a:ff:00:07","newDeviceName":"eth0","mtu":1450,"ipAddresses":[{"ipAddress":"10.255.0.7","netMask":"255.255.0.0"},{"ipAddress":"10.255.0.4","netMask":"255.255.255.255"}]},{"macAddr":"
02:42:ac:12:00:03","newDeviceName":"eth1","mtu":1500,"ipAddresses":[{"ipAddress":"172.18.0.3","netMask":"255.255.0.0"}]}],"routes":[{"dest":"default","gateway":"172.18.0.1","device":"eth1"},{"dest":"10.255.0.0",
[...]

PR #823, opened 2017-04-12
Metrics: UDP Network Bandwidth

GabyCT:
UDP network bandwidth using nuttcp with two different
buffer sizes

### DIFF
--- a/tests/metrics/network/network-metrics-nuttcp.sh
+++ b/tests/metrics/network/network-metrics-nuttcp.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Description:
+# Test inter (docker<->docker) udp network bandwidth
+# using nuttcp
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+set -e
+# Currently default nuttcp has a bug
+# see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=745051
+# Image name
+image=gabyct/nuttcp
+# Measurement time (seconds)
+time=5
+
+function setup {
+	runtime_docker
+	kill_processes_before_start
+}
+
+# This script will perform all the measurements using a local setup
+
+# Test single docker->docker udp bandwidth
+
+function udp_bandwidth {
+	setup
+	bandwidth_result=$(mktemp)
+	server_command="/root/nuttcp -u -S"
+	$DOCKER_EXE run -tid --name=nuttcp-server ${image} bash > /dev/null
+	server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
+	
+	client_command="/root/nuttcp -T${time} -u -Ru -i1 -l${bl} ${server_address}"	
+	$DOCKER_EXE exec nuttcp-server bash -c "${server_command}"
+	$DOCKER_EXE run -ti --rm --name=nuttcp-client ${image} bash -c "${client_command}" > "$bandwidth_result"
+
+	total_bandwidth=$(cat "$bandwidth_result" | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat "$bandwidth_result" | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f "$bandwidth_result"
+}
+
+function udp_default_buffer_size {
+	# Test UDP Jumbo (9000 byte MTU) packets
+	# Packet header (ICMP+IP) is 28bytes, so maximum payload is 8972 bytes
+	# See the nuttcp documentation for more hints:
+	# https://fasterdata.es.net/performance-testing/network-troubleshooting-tools/nuttcp/
+	bl=8972
+	udp_bandwidth
+}
+
+function udp_specific_buffer_size {
+	# Test UDP standard (1500 byte MTU) packets
+	# Even though the packet header (ICMP+IP) is 28 bytes, which would
+	# result in 1472 byte frames, the nuttcp documentation recommends
+	# use of 1200 byte payloads, so let's run with that for now.
+	# See the nuttcp examples.txt for more information:
+	# http://nuttcp.net/nuttcp/5.1.3/examples.txt
+	bl=1200
+	udp_bandwidth
+}
+
+udp_default_buffer_size
+echo "UDP bandwidth is (${bl} buffer size) : $total_bandwidth"
+echo "UDP % of packet loss is (${bl} buffer size) : $total_loss"
+
+udp_specific_buffer_size
+echo "UDP bandwidth is (${bl} buffer size) : $total_bandwidth"
+echo "UDP % of packet loss is (${bl} buffer size) : $total_loss"


chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
got an error on this test:
```
not ok 2 Kill container with SIGUSR1 
# (in test file kill.bats, line 55) 
# `$DOCKER_EXE ps -a | grep $container | grep "Exited (${exit_code})"' failed 
# 65cd9720256a3353ddc8c1f5ba8a985c98ddfe187df59c527c7ab1fd1e3fb755 
# JfcVUpg06YtmYIIAV1L74shil377x41I
```
which is now reported on #815 and it is unrelated to this PR 
moving to qa-passed

chavafg:
qa-passed

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+}
+
+# This script will perform all the measurements using a local setup
+
+# Test single docker->docker udp bandwidth
+
+function udp_bandwidth {
+	setup
+	bandwidth_result=$(mktemp)
+	server_command="/root/nuttcp -u -S"
```
jodh-intel: It would be slightly clearer if you move this line down to the next block (where it is used).

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+	server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
+	
+	client_command="/root/nuttcp -T${time} -u -Ru -i1 -l${bl} ${server_address}"	
+	$DOCKER_EXE exec nuttcp-server bash -c "${server_command}"
+	$DOCKER_EXE run -ti --rm --name=nuttcp-client ${image} bash -c "${client_command}" > "$bandwidth_result"
+
+	total_bandwidth=$(cat $bandwidth_result | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat $bandwidth_result | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f $bandwidth_result
```
jodh-intel: Can you quote this file "just in case".

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# This script will perform all the measurements using a local setup
+
+# Test single docker->docker udp bandwidth
+
+function udp_bandwidth {
+	setup
+	bandwidth_result=$(mktemp)
+	server_command="/root/nuttcp -u -S"
+	$DOCKER_EXE run -tid --name=nuttcp-server ${image} bash > /dev/null
```
jodh-intel: Can you add a variable for `nuttcp-server` as it's referred to multiple times.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+	server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
+	
+	client_command="/root/nuttcp -T${time} -u -Ru -i1 -l${bl} ${server_address}"	
+	$DOCKER_EXE exec nuttcp-server bash -c "${server_command}"
+	$DOCKER_EXE run -ti --rm --name=nuttcp-client ${image} bash -c "${client_command}" > "$bandwidth_result"
+
+	total_bandwidth=$(cat $bandwidth_result | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat $bandwidth_result | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
```
jodh-intel: Why isn't `nuttcp-client` deleted too?

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+function udp_bandwidth {
+	setup
+	bandwidth_result=$(mktemp)
+	server_command="/root/nuttcp -u -S"
+	$DOCKER_EXE run -tid --name=nuttcp-server ${image} bash > /dev/null
+	server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
+	
+	client_command="/root/nuttcp -T${time} -u -Ru -i1 -l${bl} ${server_address}"	
+	$DOCKER_EXE exec nuttcp-server bash -c "${server_command}"
+	$DOCKER_EXE run -ti --rm --name=nuttcp-client ${image} bash -c "${client_command}" > "$bandwidth_result"
```
jodh-intel: Again, define a variable for `nuttcp-client`.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+
+	total_bandwidth=$(cat $bandwidth_result | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat $bandwidth_result | tail -1 | awk '{print $(NF-1), $(NF)}')
```
jodh-intel: Quote `bandwidth_result` everywhere.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f $bandwidth_result
+}
+
+function udp_default_buffer_size {
+	bl=8972
```
jodh-intel: Is there something special about this number (and the `1200` below)? How were they chosen? A comment would be helpful here.

Also, I'd suggest passing this parameter into the function as a parameter rather than making the function rely on a global variable.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function udp_specific_buffer_size {
+	bl=1200
+	udp_bandwidth
+}
+
+udp_default_buffer_size
+echo "UDP bandwidth is (8972 buffer size) : $total_bandwidth"
+echo "UDP % of packet loss is (8972 buffer size) : $total_loss"
```
jodh-intel: Please don't hardcode numbers like these (above and below) - use `$bl`.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+# Description:
+# Test inter (docker<->docker) udp network bandwidth
+# using nuttcp
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Image name
+image=gabyct/nuttcp
```
jodh-intel: If the isn't a standard `nuttcp` docker image we can use, please include the `Dockerfile` for this image in this PR.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+# Measurement time (seconds)
+time=30
+set -e
```
jodh-intel: This could be moved to the top of the script.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
...
+# using nuttcp
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Image name
+image=gabyct/nuttcp
+# Measurement time (seconds)
+time=30
```
jodh-intel: How was this value chosen? What is the impact of reducing/increasing it? Please can you add a comment to provide some explanation and also point out that this variable effectively defines the duration of this test.

network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
```
jodh-intel: It looks like this script is in the wrong location - shouldn't it live below `tests/metrics/`?

network-metrics-nuttcp.sh
grahamwhaley: +1 I like that idea. Can we do that for https://hub.docker.com/r/gabyct/network/ as well, to go with the other iperf based metrics?

network-metrics-nuttcp.sh
grahamwhaley: It does seem to be in the wrong place (in the root?). I concur, it should probably live in `tests/metrics/network`

grahamwhaley: In fact, I can't run it from root, but can from under `tests/metrics/network` - due to the `source "${SCRIPT_PATH}/../../lib/test-common.bash"`. @GabyCT I guess you tested this in a different tree before a copy/merge/push cycle? :-(

grahamwhaley:
Just to note, once I'd copied the script into `tests/metrics/network`, I could run the test.
The UDP packet loss numbers made me twitch (24 and 37% under my ciao-down on my i7 lappy) - but, this is UDP... Can I ask @GabyCT and @mcastelino if we think that packet drop rate is correct/expected/acceptible? :-)
I'll re-run the test once the feedback has been addressed.

GabyCT:
@jodh-intel we can not use the default nuttcp as it has a bug in the measurement of udp so I need it to build an ubuntu container where I am compiling nuttcp in order to avoid the bug

GabyCT:
@grahamwhaley I do not know about ciao down

chavafg:
qa-failed

chavafg:
qa-passed

jodh-intel:
@GabyCT - have we raised an issue / proposed a fix for the default `nuttcp` image? It would be highly preferable to use the default image if possible.

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,78 @@
...
+# Description:
+# Test inter (docker<->docker) udp network bandwidth
+# using nuttcp
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+set -e
+# Image name
```
jodh-intel: I think a comment explaining why we aren't using the default nuttcp image is required here (ideally with a link to a bug for the upstream nuttcp if appropriate).

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,78 @@
...
+	$DOCKER_EXE run -ti --rm --name=nuttcp-client ${image} bash -c "${client_command}" > "$bandwidth_result"
+
+	total_bandwidth=$(cat "$bandwidth_result" | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat "$bandwidth_result" | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f "$bandwidth_result"
+}
+
+function udp_default_buffer_size {
+	bl=8972
```
jodh-intel: This still needs a comment: why `8972`?

GabyCT:
@jodh-intel changes applied :)

GabyCT:
@jodh-intel and no we can not use the default one as it has a bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=745051, I already put that in the script

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,81 @@
...
+# Test inter (docker<->docker) udp network bandwidth
+# using nuttcp
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+set -e
+# Currently default nuttcp has a bug
+# see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=745051
```
jodh-intel: Thanks @GabyCT - I must have been looking at a previous version as this wasn't showing for me when I reviewed it earlier.

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,81 @@
...
+
+	total_bandwidth=$(cat "$bandwidth_result" | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat "$bandwidth_result" | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f "$bandwidth_result"
+}
+
+function udp_default_buffer_size {
+	# Default size for UDP testing
+	bl=8972
```
jodh-intel: Thanks, but specifically *why* was this very unusual-looking number chosen? Please can you document why that value and not, say, `42` or `8192`, *et cetera* was chosen.

GabyCT:
@jodh-intel ohh no problem :) changes applied

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,82 @@
...
+
+	total_bandwidth=$(cat "$bandwidth_result" | tail -1 | cut -d'=' -f2 | awk '{print $(NF-11), $(NF-10)}')
+	total_loss=$(cat "$bandwidth_result" | tail -1 | awk '{print $(NF-1), $(NF)}')
+	$DOCKER_EXE rm -f nuttcp-server > /dev/null
+	rm -f "$bandwidth_result"
+}
+
+function udp_default_buffer_size {
+	# Default size for UDP high
+	# speed testing for a mtu of 9k
```
grahamwhaley: Hi @GabyCT, can we explain just a touch more why 9000 == 8972 - which, I believe is taking into account for the 28byte header overhead (ICMP??)
In which case, can I also ask why we use 1200 for the 1500 byte MTU (I'm sort of half guessing here) instead of 1472 for instance. Just need to check we clearly understand the what and whys here :-)

chavafg:
qa-passed

tests/metrics/network/network-metrics-nuttcp.sh
GabyCT: @grahamwhaley 1200 I think is the current number that we have, I do not remember all the details but I discussed that with Manohar and he agreed with that number and the first assumption is correct

GabyCT: @grahamwhaley I lost my notes where Manohar was telling me about the 1200 sorry :S

jodh-intel: I think it's important we capture why these values were chosen, so could you have a quick chat with @mcastelino and then update the comment in the script with some of these details so we can get this merged.

chavafg:
qa-failed

grahamwhaley:
@GabyCT (picking up old comment) - you don't need to know about ciao-down (but it is useful, so you might want a quick peek: https://github.com/01org/ciao/tree/master/testutil/ciao-down) - the question was more about what packet loss do you:
* see with your setup
* expect to see?

I just ran the test with runtime `runc` under ciao-down and I see much less packet loss - so, I think this is something I'd like to have some feedback on.
/cc @egernst

grahamwhaley:
General feedback:
 the test looks reasonably good. If we can get some feedback on the packet loss, and also clarify the use of the packet sizes (1200 and 8972) wrt the normal (1500) and jumbo (9000) sized packets then I think we will be fine to ack and merge.
Oh, one other thing - it would be good if the tests took a (maybe optional) argument to set the runtime to use, which then makes it easier when trying to compare things against each other.
Thanks.

mcastelino:
@grahamwhaley Running network tests or any performance tests comparing runc vs cc within ciao-down do not make sense. The VT-x EPT and TLB overhead will mask the real issues.

The only benchmarks we can perform runc vs cc inside ciao-down are the memory usage ones.

jcvenegas:
@jodh-intel  could please  you review last changes ?

GabyCT:
@grahamwhaley after talking to Manohar, here it is the reason of the different lengths

This nuttcp command transmits a 10 second stream of udp packets at 50Mbps. It
will present interval reports once a second. The interval and final reports
not only present throughput, but also packet loss. It is important to rate
limit udp transfers to avoid causing network problems with your testing. If
you know the bandwidth and utilization of the whole path you are testing then
you can choose a different number for the rate limit. The default is to use
8192 byte buffers and this may cause undesired fragmentation. If you are
still on a 1500 mtu network you should consider -l1200 to set the buffers
to 1200 bytes.]
source http://nuttcp.net/nuttcp/5.1.3/examples.txt

GabyCT:
@grahamwhaley and @jodh-intel a comment was added for the value of 1200 (reference to the nuttcp examples)

chavafg:
qa-passed

grahamwhaley:
@mcastelino We are not running performance comparisons inside ciao-down - I use ciao-down to build and run test the PRs before we merge them, and during that testing I noticed the packet losses and then checked if that was cor specific or not. The question was - did we see and/or expect those packet losses during the original implementation of these test. To that I think I still don't have an answer.

grahamwhaley:
@GabyCT Thanks for the pointer to the nuttcp notes - that helps a lot.
With that addition here and the comment added to the code:
lgtm

grahamwhaley:
@jodh-intel - OK now?

jodh-intel:
@grahamwhaley - not quite: the comment in the file still hasn't been updated to explain the discrepancy between `9000` and `8972`.

@GabyCT - could you add a comment explaining the details? The nuttcp doc introduces yet another number (`8192`), so that doesn't help expalins things for the 9k test.

GabyCT:
@jodh-intel done :)

GabyCT:
nuttcp is definitely the best tool for high-speed UDP testing. To get a full 10Gbps using UDP requires the right MTU size (9K), the right packet size (8972), and an larger buffer size.

GabyCT:
@grahamwhaley yes that is the main point of this test to see the packet losses of cor (that behaviour is also in baremetal). The main reason of why we are implementing two buffer sizes is that in the default size baremetal containers are showing almost no loss, however with a 1200 buffer size we have some loss so we want to compare our loss with baremetal containers.

chavafg:
qa-passed

GabyCT:
@grahamwhaley sure :)

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,84 @@
...
+	# https://fasterdata.es.net/performance-testing/network-troubleshooting-tools/nuttcp/
```
grahamwhaley: OK, let's see if I can supply some expanded comments for the file here so we can get this merged.
For the 9000/jumbo test I suggest:
```
# Test UDP Jumbo (9000 byte MTU) packets
# Packet header (ICMP+IP) is 28bytes, so maximum payload is 8972 bytes
# See the nuttcp documentation for more hints:
# https://fasterdata.es.net/performance-testing/network-troubleshooting-tools/nuttcp/
```

tests/metrics/network/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,84 @@
...
+	bl=8972
+	udp_bandwidth
+}
+
+function udp_specific_buffer_size {
+	# Recommended size for mtu of 1500
+	# http://nuttcp.net/nuttcp/5.1.3/examples.txt
```
grahamwhaley: And for the 'normal' MTU 1500 packets, let's try changing the comment
to this:
```
# Test UDP standard (1500 byte MTU) packets
# Even though the packet header (ICMP+IP) is 28 bytes, which would
# result in 1472 byte frames, the nuttcp documentation recommends
# use of 1200 byte payloads, so let's run with that for now.
# See the nuttcp examples.txt for more information:
# http://nuttcp.net/nuttcp/5.1.3/examples.txt
```

tests/metrics/network/network-metrics-nuttcp.sh
jodh-intel: Hi @grahamwhaley - that looks perfect to me: a nice simple yet informative comment explaining why 8972 != 9000. There can be no confusion here now.

tests/metrics/network/network-metrics-nuttcp.sh
jodh-intel: Again, +1. @GabyCT - if you are happy with these words, please apply so we can close this PR off.

GabyCT:
@jodh-intel done :)

GabyCT:
@grahamwhaley thanks for the comments

jodh-intel:
lgtm

chavafg:
qa-passed

Issue #578, opened 2017-01-16
Not possible to retrieve the hostname of a replica (swarm) with Docker 1.12.5


GabyCT:
Currently in docker 1.12.5 it is not possible to obtain the hostname of any replicas, as curl hangs all the time. However, this behavior is already fixed in Docker version 1.13.0-rc5 using runc.

1.  docker swarm init
2. docker service create --name testswarm --replicas 4 --publish 8080:80 nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \daemon off;\" 2> /dev/null
3. unset http_proxy
4. curl http://127.0.0.1:8080/hostname 
(here it hangs)

GabyCT:
Using runc with docker 1.12.5 is working perfectly.
1. url=http://127.0.0.1:8080/hostname
2. curl $url
7ca4aa689c16

This was tested on using clearlinux.

GabyCT:
Swarm should be tested using 2.1

devimc:
Port forwarding using ```docker --publish``` works
```
$ docker run --publish 8080:80 -dti nginx bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\""
4fc876a718329d74ff066a8dba895f7558218cfe5a206265f91b972766af0974
$ unset http_proxy
$ curl http://127.0.0.1:8080/hostname
4fc876a71832
```
Port forwarding with swarm is NOT working using ```http://127.0.0.1:8080/``` but it works using container's IP
```
$ docker swarm init --advertise-addr enp0s25
$ docker service create --name testswarm --replicas 1 --publish 8080:80 nginx /bin/bash -c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\"" 2> /dev/null
$ docker exec af9baed6d0ac ip a
...
3: eth1: <BROADCAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 02:42:ac:12:00:03 brd ff:ff:ff:ff:ff:ff
    inet 172.18.0.3/16 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe12:3/64 scope link 
       valid_lft forever preferred_lft forever
....
$ curl http://172.18.0.3/hostname
af9baed6d0ac
```
notice that the IP of this container is **172.18.0.3**
I executed ```iptables``` to get more information about port forwarding
```
[...]

PR #825, opened 2017-04-12
Save and flush iptables in network namespace

amshinde:


src/networking.c
```diff
@@ -539,6 +539,122 @@ get_mac_address(const gchar *const ifname)
...
+
+	/* Set the default policies for the built-in chains to ACCEPT,
+	 * flush nat and mangle tables, flush all chains(-F) and
+	 * delete all user chains(-X).
+	 */
+	char *iptables_drop_command = "iptables -P INPUT ACCEPT "
+			"&& iptables -P FORWARD ACCEPT "
+			"&& iptables -P OUTPUT ACCEPT "
+			"&& iptables -t nat -F "
+			"&& sudo iptables -t mangle -F "
```
mcastelino: Why is there a sudo here

src/networking.c
```diff
@@ -539,6 +539,122 @@ get_mac_address(const gchar *const ifname)
...
+
+		if (! g_spawn_check_exit_status(exit_status, &error)) {
+			g_critical("Error dropping iptable rules: %s",
+				error->message);
+			g_clear_error(&error);
+		}
+	}
+
+	// Free rules early on.
+	g_free_if_set(config->net.iptable_rules);
```
mcastelino: We may want to retain what we had scanned initially. So may be best to free it outside explicitly.

amshinde: Ok, I thought we could free memory early on, I'll change this.

src/networking.c
amshinde: Removed the sudo.

src/networking.c
mcastelino: @amshinde I was thinking it may be useful if we save this information in our state file or elsewhere. That way we will be able to detect any issues. Otherwise you can free it here.

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+ *
+ * \return \c true on success, else \c false.
+ */
+gboolean
+get_iptable_rules(struct cc_oci_config *config)
+{
+	gboolean ret;
+	GError *error = NULL;
+	gint exit_status;
+	char *args[2] = { "iptables-save", NULL };
```
jodh-intel: Why not just drop the `2` and let the compiler deal with it? Also, `const` maybe?

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+              NULL,
+              NULL,
+              &(config->net.iptable_rules),
+              NULL,
+              &exit_status,
+              &error);
+
+	if ( !ret || exit_status) {
+		if (! ret) {
+			g_critical("Error spawning process for iptables-save:%s",
```
jodh-intel: Why not define a variable for `iptables-save` to avoid hard-coding this.

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+	gboolean ret;
+	GError *error = NULL;
+	gint exit_status;
+	char *args[2] = { "iptables-save", NULL };
+
+	if (! config) {
+		return false;
+	}
+
+	ret = g_spawn_sync (NULL,
```
jodh-intel: This is a blocking call, so you should add a `g_debug()` listing the command being run to give a clue to an admin in case it blocks indefinitely.

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+	 */
+	char *iptables_drop_command = "iptables -P INPUT ACCEPT "
+			"&& iptables -P FORWARD ACCEPT "
+			"&& iptables -P OUTPUT ACCEPT "
+			"&& iptables -t nat -F "
+			"&& iptables -t mangle -F "
+			"&& iptables -F "
+			"&& iptables -X";
+
+	char *args[4] = { "/bin/bash", "-c", iptables_drop_command, NULL};
```
jodh-intel: - Again, I'd drop the `4`.
- I don't think you should rely on `bash` (added to which it's huge and slow), so `/bin/sh` (`_PATH_BSHELL`) would be preferable.
- If we really have to call out to iptables, it would be very useful to add a couple of `g_debug()` calls (before and after) so we can see the impact in the log.
- Shouldn't we be using `ip rule ...` rather than `iptables` anyway?

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+ * Flush all iptable rules set in network namespace.
+ * We need to do this as rules get applied to the
+ * packets crossing the tap bridge.
+ *
+ * \param config \ref cc_oci_config
+ *
+ * \return \c true on success, else \c false.
+ */
+gboolean
+purge_iptable_rules(struct cc_oci_config *config)
```
jodh-intel: `static`? Also `config` isn't used.

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+ * Get and store the iptable rules set in the
+ * network namespace. These need to be re-applied
+ * within the container.
+ *
+ * \param config \ref cc_oci_config
+ *
+ * \return \c true on success, else \c false.
+ */
+gboolean
+get_iptable_rules(struct cc_oci_config *config)
```
jodh-intel: `static`?

src/networking.c
```diff
@@ -539,6 +539,119 @@ get_mac_address(const gchar *const ifname)
...
+              &(config->net.iptable_rules),
+              NULL,
+              &exit_status,
+              &error);
+
+	if ( !ret || exit_status) {
+		if (! ret) {
+			g_critical("Error spawning process for iptables-save:%s",
+				 error->message);
+			g_clear_error(&error);
```
jodh-intel: Do you want to `goto out` here? (same question for `purge_iptable_rules()`).

src/networking.c
amshinde: Fixing

src/networking.c
amshinde: afaik ip rule is meant for policy rules for routing table, here we need to do deal with iptables.

src/networking.c
amshinde: No, I  had intended it this way.

amshinde:
@jodh-intel Addressed your comments. Please have a look.

jodh-intel:
lgtm

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+
+	if (! config) {
+		return false;
+	}
+
+	g_debug("Running command : %s", cmd);
+	ret = g_spawn_sync (NULL,
+              args,
+              NULL,
+              G_SPAWN_SEARCH_PATH | G_SPAWN_STDERR_TO_DEV_NULL,
```
devimc: I'd rather use full path 
according with glib doc
```
If the program name is not a full path and G_SPAWN_SEARCH_PATH flag is not used, then the program will be run from the current directory (or working_directory , if specified); this might be unexpected or even dangerous in some cases when the current directory is world-writable.
```

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+	 * flush nat and mangle tables, flush all chains(-F) and
+	 * delete all user chains(-X).
+	 */
+	char *iptables_drop_command = "iptables -P INPUT ACCEPT "
+			"&& iptables -P FORWARD ACCEPT "
+			"&& iptables -P OUTPUT ACCEPT "
+			"&& iptables -t nat -F "
+			"&& iptables -t mangle -F "
+			"&& iptables -F "
+			"&& iptables -X";
```
devimc: probably you should specify the fullpath

src/networking.c
amshinde: Different distros may have this under different path.
eg Ubuntu has this under /sbin/iptables, while Clear has this under /usr/bin.
Docker and hyper codebases do not provide the full path as well.

Also, since G_SPAWN_SEARCH_PATH is provided, the program will not be run from the current directory.

amshinde: @devimc Does this answer your comment?

devimc: @amshinde yes, it was just a comment

devimc:
lgtm

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+ *
+ * \return \c true on success, else \c false.
+ */
+static gboolean
+get_iptable_rules(struct cc_oci_config *config)
+{
+	gboolean ret;
+	GError *error = NULL;
+	gint exit_status = 0;
+	char *cmd = "iptables-save";
```
jodh-intel: `const`.

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+	char *args[] = { cmd, NULL };
```
jodh-intel: `const`.

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+{
+	gboolean ret;
+	GError *error = NULL;
+	gint exit_status = 0;
+
+	/* Set the default policies for the built-in chains to ACCEPT,
+	 * flush nat and mangle tables, flush all chains(-F) and
+	 * delete all user chains(-X).
+	 */
+	char *iptables_drop_command = "iptables -P INPUT ACCEPT "
```
jodh-intel: `const`.

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+			"&& iptables -P FORWARD ACCEPT "
+			"&& iptables -P OUTPUT ACCEPT "
+			"&& iptables -t nat -F "
+			"&& iptables -t mangle -F "
+			"&& iptables -F "
+			"&& iptables -X";
+
+	char *args[] = { "/bin/sh", "-c", iptables_drop_command, NULL};
```
jodh-intel: `const`.

src/networking.c
```diff
@@ -539,6 +539,116 @@ get_mac_address(const gchar *const ifname)
...
+
+	g_debug("Running commands for dropping iptables");
```
jodh-intel: Can you add the commands that are run to this message? For multi-argument command-lines, here's the boilerplate convention we use:

- https://github.com/01org/cc-oci-runtime/blob/master/src/process.c#L698..L701

src/networking.c
amshinde: @jodh-intel I avoided this as g_spawn_async expects non-const arguments, passing const gives a bunch of warnings

jodh-intel: ack.

src/networking.c
amshinde: I am not so keen on logging here, as the commands run are hardcoded, we do not really pass any variable arguments here.

jodh-intel: `iptables-save` is hard-coded too... We can't assume that all users will have the source code or want to wade through it to find out what is happening at this point. Since this is a blocking call, we need to log what the runtime is doing.

src/networking.c
amshinde: ditto for all others.

src/networking.c
amshinde: Added this.

jodh-intel:
Thanks.

lgtm

chavafg:
already tested the port forwarding  (#578)  with image 14700 locally and works just fine.

amshinde:
PR #831 needs to be merged first, before this can go in.

chavafg:
qa-failed

chavafg:
qa-passed

PR #832, opened 2017-04-21
networking: Disable udev from assigning predictable interface names

amshinde:
This kernel command line option disables systemd/udev from
assigning predictable interface names.

Now that we do not pass interface names to qemu and do not rely
on systemd/udev assigning them interface names, we can safely
disable this. This was  causing a race condition
wherein we were querying for interface names within hyperstart
with the mac addresses in the middle of systemd/udev renaming these.

We were dealing with the race conditions earlier, by waiting for
the changed interface name, but this was broken with introducing
configuration based on mac addresses in hyperstart and omitting
interface names passed to qemu altogether.

Fixes: https://github.com/01org/cc-oci-runtime/issues/815


chavafg:
qa-passed

dlespiau:
Awesome find! lgtm

dlespiau:
lgtm

jodh-intel:
lgtm

Issue #828, opened 2017-04-17
synchronization bug (create -> start -> kill)


devimc:
running container with runc
```
for i in `seq 1 15`; do docker run --runtime=runc -dti --name bla$i ubuntu bash -c "trap \"exit 15\" SIGUSR1 ; while : ; do sleep 1; done" && docker kill -s SIGUSR1 bla$i ; done
```
the status of all container will be ```Exited (15)```

but using cor
```
for i in `seq 1 15`; do docker run --runtime=cor -dti --name bla$i ubuntu bash -c "trap \"exit 15\" SIGUSR1 ; while : ; do sleep 1; done" && docker kill -s SIGUSR1 bla$i ; done
```
the status of some containers will be 0

grahamwhaley:
Hi @devimc Do you know if this exists still for CC3.x https://github.com/clearcontainers/runtime ?  I just did a test run, and did not see the issue - would be good if somebody else could verify?
btw, @chavafg - do we test this in the QA tests? Maybe we can add a test if we don't have one already?

I ask as we still reference this issue in https://github.com/clearcontainers/tests/pull/381

chavafg:
Still fails for me using 2nd test of: https://github.com/01org/cc-oci-runtime/blob/master/tests/integration/docker/kill.bats 

We still do not have this test on the `clearcontainers/tests` repo.

I tested this under a VM... maybe a race condition? @grahamwhaley are you running on baremetal?

grahamwhaley:
@chavafg yep, tested on a bare metal Ubuntu, just one run of the simple script at the top of this Issue.
If we have no record of this in CC3.x repos, can we at minimum open an Issue there pls?

chavafg:
@grahamwhaley, created https://github.com/clearcontainers/runtime/issues/430

devimc:
Hi @grahamwhaley , I don't think so

PR #841, opened 2017-04-24
Add autoconf-archive as dependency

devimc:
ax_code_coverage.m4 and ax_valgrind_check.m4 are upstream
files maintained by GNU Autoconf Archive we should use
them instead of maintaining own/local versions


dlespiau:
There is no comment on the reason(s) you'd like to do that. Mind adding those to the PR and commit messages?

grahamwhaley:
@jodh-intel Can you verify for me this is a good thing to do away with our direct pulling in of those two modules. I think we can go either way (keep the files, or do a dep) - do you have a preference?

chavafg:
qa-failed

jodh-intel:
@grahamwhaley - I'd rather we use the newest versions possible.

One point to consider is whether the local changes we have made to these files (cf3e64823981b30d6a40dd57f583eb2786dc9aec and c4660de7672ee62a1bb63760c4928913afbc7edd) need to be upstreamed before we switch.

devimc:
@jodh-intel sounds good, could you send a email to autoconf archive maintainers (autoconf-archive-maintainers@gnu.org) ? might be they will accept your changes

jodh-intel:
@devimc - I haven't checked to see if our changes are already fixed upstream yet. But looking at those commits, if this branch can be made to run under Travis, we prolly don't need to upstream.

chavafg:
qa-failed

chavafg:
qa-passed

chavafg:
qa-failed

chavafg:
got this random failure: 
```
not ok 2 Kill container with SIGUSR1 
```
which is already reported on #828

chavafg:
qa-passed

jcvenegas:
lgtm

PR #845, opened 2017-04-25
version: Update versions of linux kernel and qemu-lite

gorozco1:
qemu-lite and linux-kernel used to run clear-containers were pointing to
the wrong versions. Currently we are using newer versions of linux
kernel and qemu-lite.


versions.txt
```diff
@@ -7,10 +7,10 @@ docker_ubuntu_version=16.04
 docker_engine_fedora_version=1.12.1
 docker_engine_ubuntu_version=1.12.1-0~xenial
 clear_vm_image_version=14700
-clear_vm_kernel_version=4.5-57
+clear_vm_kernel_version=4.9.4-57
```
jcvenegas: looks good, btw we need to update kernel version to the last minior version (4.9.24)

jcvenegas:
lgtm

versions.txt
gorozco1: At the moment the kernel we use and we build is 4.9.4 no 4.9.24, when we update the kernel we can change this line ;-)

chavafg:
qa-passed

chavafg:
qa-passed

jodh-intel:
FYI - I've raised #847 for an idea I have to improve our process and avoid such oversights.

lgtm.

versions.txt
grahamwhaley: Note, I have recently asked @miguelinux if/when/should we update the container kernel to the latest longerm (4.9.24)... I think that is now overdue.

PR #846, opened 2017-04-26
networking: Disable multicast snooping on the bridge we create

### LINKED ISSUE #581: CI tests: kernel panic in the DUT randomly
Recently I have seen that sometimes the machines that run the CI tests are getting hanged. I have experienced the same issue on ClearLinux and Ubuntu machines. I have not seen this behavior in the Fedora machine. 

When trying to connect the monitor to the hanged machine, I cannot see any output. On the test execution log I can see that the last tests that were being executed:

 ✓ HostName is passed to the container
 ✓ Verify connectivity between 2 containers
 ✓ Port a container
 ✓ Restart a container
 ✓ Run shell echo
 ✓ stdout using run
 ✓ stderr using run
 ✓ stdin from pipe
 ✓ check that the replicas' names are different

The last test belong to the swarm test suite. 

I am trying to reproduce the error and will document further as I get more information.

amshinde:
This may possibly resolve #581. Multicast snooping may be
turned on the bridge depending on the distro kernel config.
Disable this explicitly.

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -213,6 +213,7 @@ netlink_link_add_bridge(struct netlink_handle *const hndl,
 	struct nlmsghdr *nlh = NULL;
 	struct ifinfomsg *ifm = NULL;
 	struct nlattr* link_attr = NULL;
+	bool disable_snooping = false;
 
 	if ((hndl == NULL) || (name == NULL)) {
 		g_critical("%s NULL parameter", __func__);
@@ -230,8 +231,36 @@ netlink_link_add_bridge(struct netlink_handle *const hndl,
 	mnl_attr_put_str(nlh, IFLA_IFNAME, name);
 	link_attr = mnl_attr_nest_start(nlh, IFLA_LINKINFO);
 	mnl_attr_put_str(nlh, IFLA_INFO_KIND, "bridge");
-	mnl_attr_nest_end(nlh, link_attr);
 
+	/* Disable multicast snooping on the bridge.
+	 * Netlink support for this was added in kernel 4.4 where
+	 * IFLA_BR_MCAST_SNOOPING was defined with value 23 in an enum.
+	 * Reference:
+	 * https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_link.h
+	 *
+	 * So check if the last value of the enum is greater than 23, before
+	 * adding the value to the netlink nested attributes to be compatible
+	 * with earlier kernels.
+	 * If CONFIG_BRIDGE_IGMP_SNOOPING is disabled in a newer kernel,this
+	 * attribute is simply ignored.
+	 */
+	#ifdef IFLA_BR_MAX
+		if (IFLA_BR_MAX > 23) {
+			disable_snooping = true;
+			g_debug("Turning off multicast snooping for bridge %s",
+				name);
+			struct nlattr *link_data = mnl_attr_nest_start(nlh,
+						IFLA_INFO_DATA);
+			mnl_attr_put_u8(nlh, IFLA_BR_MCAST_SNOOPING, 0);
+			mnl_attr_nest_end(nlh, link_data);
+		}
+	#endif
+
+	if (! disable_snooping) {
+		g_debug("Not possible to turn off bridge multicast snooping");
+	}
+
+	mnl_attr_nest_end(nlh, link_attr);
 	return netlink_execute(hndl, nlh);
 }
 


chavafg:
qa-passed

amshinde:
Travis logs say that identifier IFLA_BR_MCAST_SNOOPING not declared. Looks like netlink support for multicast snooping was added recently in kernel 4.4.

dlespiau:
So, on older kernels or if snooping isn't compiled in, the call will fail. That's totally fine but, just in case we check for success of the operation, we're happy to fail here, just means unsupported.

For the compilation issue, we'll have to define the constant ourselves to compile on older systems. https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_link.h has the definition (it's an enum) and if I'm not mistaken its value is `23`. We can do be a bit creative with:

```
#if IFLA_BR_MAX < 23
#define IFLA_BR_MCAST_SNOOPING 23
#endif
```

jodh-intel:
It might even make sense to put that `#if` test in a new `compat.h` along with our existing check for `CLONE_NEWCGROUP`. Just an idea.

Either way, please document clearly in the code why we are defining `IFLA_BR_MCAST_SNOOPING`.

@chavafg - regarding https://github.com/01org/cc-oci-runtime/issues/581#issuecomment-273484961, approximately how many iterations of `test-campaign.sh` are required to trigger the crash? Will 100 iterations almost(?) guarantee it?

Once we know that, @amshinde please could you run the test in question so we can say in the commit message with a little more confidence that #581 appears to be fixed (in other words drop the "may").

chavafg:
100 iterations of `test-campaign.sh` should be enough. Or to make it fail quicker just run the swarm tests:  `prove -v swarm.bats` with the same iterations. 
I just broke it on iteration 47 today. It took around 3 hrs.

chavafg:
@amshinde I can run the tests with your PR, but not sure if comments need to be addressed before? Please let me know

amshinde:
@dlespiau Nice suggestion! I'll use the IFLA_BR_MAX value. How about 
```
#if IFLA_BR_MAX > 23
    //Add to netlink attributes
#endif
```
That way we can avoid the extra call

amshinde:
@chavafg Thanks, I'll push changes and let you know. On what distribution did you run the tests?
I tried running it on ubuntu yesterday, but could not reproduce it all.

chavafg:
I was able to reproduce it on Fedora 25 and ubuntu 16.04

chavafg:
qa-passed

amshinde:
@chavafg Pushed an update. Can you run the tests and let me know if you still see the crash?

chavafg:
qa-passed

chavafg:
@amshinde, sure, will run in a loop and let you know

chavafg:
qa-passed

src/netlink.c
```diff
@@ -230,8 +230,29 @@ netlink_link_add_bridge(struct netlink_handle *const hndl,
...
+	 * Netlink support for this was added in kernel 4.4 where
+	 * IFLA_BR_MCAST_SNOOPING was defined with value 23 in an enum.
+	 * Reference:
+	 * https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_link.h
+	 *
+	 * So check if the last value of the enum is greater than 23, before
+	 * adding the value to the netlink nested attributes to be compatible
+	 * with earlier kernels.
+	 */
+	#ifdef IFLA_BR_MAX
```
dlespiau: I'm afraid that may not work. One could have a recent enough kernel and the snooping code not compiled in. Probably need to double check that but, conceptually, the kernel may/should reject that value if the code to handle it isn't loaded/compiled in. 

A couple of things you may find interesting:
- you don't need that ifdef, the preprocessor will evaluate it to 0 if undefined.
- should it have been needed, there's a more elegant construct that can be nice to use in the future, eg.
```
#if defined(FOO) && FOO > 23
```

chavafg:
@amshinde, tried to reproduce the kernel panic on Ubuntu 16.10 and Fedora 25 with no luck. 

So seems that this PR fixes #581. :)

amshinde:
@chavafg Thanks!

src/netlink.c
amshinde: @dlespiau I'll check the case where the module is disabled altogether, I was thinking the kernel would just ignore the value.
For the second case, travis had failed earlier because IFLA_BR_MAX itself was not defined, either #ifdef or the defined operator are required for really old kernels

amshinde: @dlespiau Checked this with a kernel with CONFIG_BRIDGE_IGMP_SNOOPING disabled. In that case the attribute related to multicast snooping is ignored.
If the bridge or br_netfilter module is disabled, then we would not be able to create the bridge in the first place.
Also note, I am checking for the value of IFLA_BR_MAX at runtime, since this macro is not evaluated at preprocessor time as it depends on an enum value.

chavafg:
qa-passed

amshinde:
@dlespiau @jodh-intel Any feedback on this?

jodh-intel:
lgtm

dlespiau:
Thanks for double checking everything, seems like it's all good now.

lgtm

dlespiau:
Argh! this commit was missing the Fixes: tag. Next time!

dlespiau:
Oh, somehow github manged to notice anyway.

grahamwhaley:
heh, I think there was a relevant fixes comment in the thread :-)

PR #855, opened 2017-04-28
docs: Require every PR to include a "Fixes #XXX" comment.

jodh-intel:
Previously adding a "Fixes #XXX" commit message was optional, but we
have found that "stale" issues are being left open. Require such a
comment to have those issues automatically closed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,8 @@ For more information on github 'force push' workflows see [here](http://blog.ada
 
 It is perfectly fine for your Pull Request to contain more than one patch - use as many patches as you need to implement the Request (see the previously mentioned 'small patch' thoughts). Each Pull Request should only cover one topic - if you mix up different items in your patches or pull requests then you will most likely be asked to rework them.
 
+All pull requests must include a reference to the issue or issues they solve, see [Closing Issues](#Closing-issues).
+
 ## Reviews
 
 Before your Pull Requests are merged into the main code base, they will be reviewed. Anybody can review any Pull Request and leave feedback (in fact, it is encouraged), but this project runs a rotational GateKeeper schedule for who is ultimately responsible for merging the Pull Requests into the main code base. See [here](https://github.com/01org/cc-oci-runtime/wiki/GateKeeper-Schedule).
@@ -120,16 +122,21 @@ visibility on the problem and work toward resolution.
 
 ## Closing issues
 
-You can either close issues manually by adding the fixing commit SHA1 to the issue
-comments or by adding the `Fixes` keyword to your commit message:
+You must add a `Fixes` keyword along with the issues fixed in the body
+of one of your commits:
 
 ```
-Fix handling of semvers with only a single pre-release field
+install: Remove RHEL-specific code to handle missing chronic(1).
 
-Fixes #121
+Don't conditionally call chronic(1) if not running under RHEL - just
+check if chronic is available and use it if it is.
 
-Signed-off-by: James Hunt <james.o.hunt@intel.com>
+Fixes #789.
+
+Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>
 ```
 
 Github will then automatically close that issue when parsing the
 [commit message](https://help.github.com/articles/closing-issues-via-commit-messages/).
+
+If there is no an associated issue number, please create one.


grahamwhaley:
Sorry to be a pain - but can we have some mention of 'all PRs must container an associated <Fixes>' in the 'Pull requests' section as well? I suspect many folks will read the 'Pull requests' section, but may never get past that and hence miss the Fixes requirement - so, calling it out in the 'Pull requests' section itself I think would be good.

jodh-intel:
Done. Having taken another read of that doc, I think it really could do with a refresh. Two immediate points:

- "patches" should probably be "commits".
- The structure could do with a rework.

CONTRIBUTING.md
```diff
@@ -88,6 +88,8 @@ For more information on github 'force push' workflows see [here] (http://blog.ad
 
 It is perfectly fine for your Pull Request to contain more than one patch - use as many patches as you need to implement the Request (see the previously mentioned 'small patch' thoughts). Each Pull Request should only cover one topic - if you mix up different items in your patches or pull requests then you will most likely be asked to rework them.
 
+All PRs must contain a reference to the issue(s) they resolve (see #Closing-issues).
```
grahamwhaley: I think you can use [Closing issues](CONTRIBUTING.md#closing-issues) here for a hotlink - but, I'm not going to let that stop a merge - your call.

grahamwhaley:
With a minor comment,
lgtm

CONTRIBUTING.md
jodh-intel: That should be a hotlink (an doc internal one), but doesn't seem to render in the viewer.

dlespiau:
Somehow my idea to generate the list of bugs fixed from Fixes: tags transformed into "every single PR needs to have an associated issue and Fixes: tag". That's not quite the same thing. I hope we haven't just added a new barrier to contribution and some "process" one would have to go through even if it doesn't make sense.

That may be fine for very stable branches, but for general development, I don't think that's reasonable.

dlespiau:
Maybe a bit less confrontational: for stable branches, yes! otherwise, no :)

It's just that 2.1.x hasn't really felt like a stable branch, we're adding many features which isn't something we should have been doing, I believe.

CONTRIBUTING.md
```diff
@@ -120,8 +122,7 @@ visibility on the problem and work toward resolution.
 
 ## Closing issues
 
-You can either close issues manually by adding the fixing commit SHA1 to the issue
-comments or by adding the `Fixes` keyword to your commit message:
+You must add a `Fixes` keyword to one of your commits:
```
rcaballeromx: Consider changing it to:
Start the first line of at least one commit message with 'Fix'.
My issue with your version is that the example below does not have the word Fixes but Fix in it.
It is also important to state where in the commit message you are expecting the keyword.

CONTRIBUTING.md
```diff
@@ -88,6 +88,8 @@ For more information on github 'force push' workflows see [here] (http://blog.ad
 
 It is perfectly fine for your Pull Request to contain more than one patch - use as many patches as you need to implement the Request (see the previously mentioned 'small patch' thoughts). Each Pull Request should only cover one topic - if you mix up different items in your patches or pull requests then you will most likely be asked to rework them.
 
+All PRs must contain a reference to the issue(s) they resolve (see [Closing Issues](#Closing-issues)).
```
rcaballeromx: Consider:
All pull requests must include a reference to the issue or issues they solve, see [Closing Issues](#Closing-issues).

jodh-intel:
@dlespiau - This PR was the result of a combination of:

- Your thoughts on auto-generating some of the release-notes.
- The fact that this project is now becoming mature.
- We have introduced the `Fixes #XXX` requirement into the release process so why not just have a consistent process?

I don't see this as an onerous step nor a barrier to contribution: creating an issue it the matter of a few seconds work: click "New Issue" and fill out two text boxes.

Yes, it's a little annoying to have to create an issue if you don't already have one, but (and this is important), **having an issue assigned to you is a clear indication to the whole team what you are currently working on**. The alternative is that suddenly a PR is raised and nobody had any idea it was being worked on. Or worse, *multiple* PRs to solve the same problem get raised. At that point, unfortunately someone is going to be upset. Much better to be able to have an issue assigned to you so that you have "author rights" on PRs to resolve that issue.

Hence, the way we should all be working is:

1. Raise an issue
1. Assign it to yourself
1. Raise a PR and reference the issue.

jodh-intel:
@rcaballeromx - PR updated. Could you take another look?

rcaballeromx:
@jodh-intel The wording on fixes has not changed. Could correct it? It is confusing to read the instructions just to have an example which does not follow them.

CONTRIBUTING.md
jodh-intel: If you look further down that example, you'll see `Fixes #121`. The title start with "Fix handling".

I've updated that example with an actual commit entry.

jodh-intel:
@rcaballeromx - updated.

chavafg:
qa-failed

chavafg:
qa-failed

jcvenegas:
lgtm

Issue #848, opened 2017-04-26
Add dockerfiles for the gabyct/network docker network image


GabyCT:
Add dockerfile for the image that is using the network metrics

jodh-intel:
Related to #843.

jodh-intel:
The  custom `gabyct/network` docker image is used by:

- `tests/metrics/network/network-metrics.sh`
- `tests/metrics/network/network-metrics-iperf3.sh`

And soon to also be used by:

- `tests/metrics/network/network-metrics-cpu-consumption.sh.in` (#843).

jodh-intel:
Hi @GabyCT - any update on this?

GabyCT:
@jodh-intel, I need some time to do this...kind of busy these days

GabyCT:
@jodh-intel PR #895

GabyCT:
Closing this issue as PR is already merged

PR #865, opened 2017-04-28
Metrics: Measure PSS memory while running network

GabyCT:
Measure Proportional Set Size Memory using smem while
running a bandwidth measurement using iperf2

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -68,6 +68,7 @@ GENERATED_FILES = \
 	tests/metrics/density/docker_memory_usage.sh \
 	tests/metrics/workload_time/cor_create_time.sh \
 	tests/metrics/network/network-metrics-cpu-consumption.sh \
+	tests/metrics/network/network-metrics-memory-pss.sh \
 	tests/lib/test-common.bash
 
 $(GENERATED_FILES): %: %.in Makefile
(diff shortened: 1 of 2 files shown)


chavafg:
qa-passed

amshinde:
lgtm

jodh-intel:
lgtm

jodh-intel:
See #848.

mcastelino:
This PR measures the PSS at max bandwidth
-	Measure absolute numbers for runc and cor
         - Max bandwidth
         - Latency at max
         - Jitter at max
- CPU/Memory utilization at max

But we know that the max is different for runc and cor.

So we should run a second set of tests where cor and runc are doing the “same amount of work”

So we know for example both cor and runc can easily handle 10Gbps of TCP traffic. And the test tools supports bandwidth capping.

So we should measure the 
-	CPU/Memory utilization @ particular load level (say 10Gbps)

What this does it, let us really figure out relative “extra” overhead of cor vs runc.
We can run multiple load levels like 1Gbps, 5 Gbps and 10Gbps to see if the scaling is non linear.

This is not needed as part of this test, but a separate PnP test.

jodh-intel:
@GabyCT - please could you raise an issue to cover the new test @mcastelino has outlined?

GabyCT:
@jodh-intel , I added #874 for the memory first and then I will add another one for the CPU when I figured out what is the best way to do it :) thanks

Issue #792, opened 2017-04-03
Add a autoconf variable to allow the specification of a "small" docker image


jodh-intel:
*Lots* of files currently hardcode the docker image `busybox` as it is a very small but useful docker image.

Rather than hard-coding it everywhere we should define it in `configure.ac` to allow all files in the repo to make use of the variable.

We could also provide a `--with-small-docker-image-name=` option to `configure.ac` to allow the default to be changed. This would be useful to shake out any bugs we have where a busybox environment is assumed.

PR #867, opened 2017-05-02
Metrics: Latency measurement

GabyCT:
This will measure latency when we do a ping
from one container to another (docker <-> docker)

### DIFF
--- a/tests/metrics/network/network-latency.sh
+++ b/tests/metrics/network/network-latency.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Description:
+# This will measure latency when we do a ping
+# from one container to another (docker <-> docker)
+
+set -e
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Image name (ping installed by default)
+image=busybox
+# Number of packets (sent)
+number=10
+
+function setup {
+	runtime_docker
+	kill_processes_before_start
+}
+
+# This script will perform all the measurements using a local setup
+
+# Test latency docker<->docker using ping
+
+function latency {
+	setup
+	latency_result=$(mktemp)
+	container_name="first_container"
+	second_container_name="second_container"
+	$DOCKER_EXE run -tid --name ${container_name} ${image} sh > /dev/null
+	ip_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" ${container_name})
+	$DOCKER_EXE run -ti --rm --name ${second_container_name} ${image} sh -c "ping -c ${number} ${ip_address}" > "$latency_result"
+	latency_average=$(cat $latency_result | grep avg | tail -1 | awk '{print $4}' | cut -d '/' -f 2)
+	$DOCKER_EXE rm -f ${container_name} > /dev/null
+	rm -f "$latency_result"
+}
+
+latency
+echo "The average latency is : $latency_average ms"
+


tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,56 @@
...
+}
+
+# This script will perform all the measurements using a local setup
+
+# Test latency docker<->docker using ping
+
+function latency {
+	setup
+	latency_result=$(mktemp)
+	$DOCKER_EXE run -tid --name first_container ${image} sh > /dev/null
```
jcvenegas: why send output to /dev/null ?

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,56 @@
...
+# Description:
+# This will measure latency when we do a ping
+# from one container to another (docker <-> docker)
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Image name
+image=gabyct/network
```
jcvenegas: I wonder if we could add this images to clearcontainers dockerhub organization ?

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,56 @@
...
+# This script will perform all the measurements using a local setup
+
+# Test latency docker<->docker using ping
+
+function latency {
+	setup
+	latency_result=$(mktemp)
+	$DOCKER_EXE run -tid --name first_container ${image} sh > /dev/null
+	ip_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
+	$DOCKER_EXE run -ti --rm --name second_container ${image} bash -c "ping -c ${number} ${ip_address}" > "$latency_result"
```
jcvenegas: Instead of send the ping output to a tmpfile , I wonder if the ping output could be stored in a variable ?

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,56 @@
...
+	latency_average=$(cat $latency_result | tail -1 | awk '{print $4}' | cut -d '/' -f 2)
```
jcvenegas: Here it gets the last line, what if the last line does not have the information you are looking for ? I'd suggest to use a regex if possible

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Description:
+# This will measure latency when we do a ping
+# from one container to another (docker <-> docker)
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
```
jcvenegas: I would add set -e  to the script in order to catch any issue and exit.

tests/metrics/network/network-latency.sh
GabyCT: @jcvenegas we already had a discussion about that, #848

tests/metrics/network/network-latency.sh
GabyCT: @jcvenegas so it does not display the container id when I am running the test

tests/metrics/network/network-latency.sh
jcvenegas: @GabyCT  thank  you for refer to the issue

GabyCT:
@jcvenegas changes were applied

tests/metrics/network/network-latency.sh
GabyCT: @jcvenegas, I like to direct the output to a tmpfile

chavafg:
qa-passed

chavafg:
qa-failed

chavafg:
qa-passed

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,58 @@
...
+}
+
+# This script will perform all the measurements using a local setup
+
+# Test latency docker<->docker using ping
+
+function latency {
+	setup
+	latency_result=$(mktemp)
+	$DOCKER_EXE run -tid --name first_container ${image} sh > /dev/null
```
jodh-intel: Can you use a variable for the container name please?

GabyCT:
@jodh-intel changes applied

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,60 @@
...
+# from one container to another (docker <-> docker)
+
+set -e
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
+source "${SCRIPT_PATH}/../../lib/test-common.bash"
+
+# Image name
+image=debian
```
jodh-intel: Could you put a comment explaining why this image was chosen? (Basically why not `busybox` ;-)

GabyCT:
@jodh-intel, I used this one because it has ping installed, however, now that I am checking busybox also have it by default should I change it?

jodh-intel:
Most of the tests use the `busybox` image (see #792), but we're not limited to that image if you need to use others. But if you do use another one, I think a comment would be helpful to understand why it was chosen.

GabyCT:
@jodh-intel got it :) I change it to use busybox and put a comment :) thanks

chavafg:
qa-passed

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# This script will perform all the measurements using a local setup
+
+# Test latency docker<->docker using ping
+
+function latency {
+	setup
+	latency_result=$(mktemp)
+	container_name="first_container"
+	second_container_name="second container"
```
jodh-intel: I think this name is invalid (no spaces allowed).

grahamwhaley: Agreed. Let's make the names consistent. In fact, the test does not run for me with the space in the name.

tests/metrics/network/network-latency.sh
```diff
@@ -0,0 +1,60 @@
...
+	$DOCKER_EXE run -tid --name ${container_name} ${image} sh > /dev/null
+	ip_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" $($DOCKER_EXE ps -ql))
```
grahamwhaley: Rather than using the `docker ps -qa` here, can you not use the `${container_name} - which feels more correct and robust.
A local tests showed it to work for me.

grahamwhaley:
Just a couple of notes, not blockers to the PR necessarily:
 - the SD on the test is fairly high during my testing (nature of the beast), at around 20% variability. I bumped the 'number' from 10 to 100 and that helped a bit, but not vastly. Just something to consider (and maybe do some more number crunching and analysis on)
 - If we do bump the number of iterations and the test becomes laboriously long then we can get ping to work quicker by setting its '-i' interval (default 1s I think - we can set it to significantly less than that, as the ping times we are seeing are in the sub-ms range). Note though, busybox ping has limited functionality and is missing '-i', so we would have to move back to another image like the Debian one for instance if we want to add the feature.

chavafg:
qa-passed

GabyCT:
@jodh-intel and @grahamwhaley thanks for the feedback, changes applied

jodh-intel:
lgtm

chavafg:
qa-passed

grahamwhaley:
I'd like to run-test this before I merge, but won't find time today... if somebody wants to merge it, please give it a twirl. You'll need something like:
```
# make metrics-tests
# #stop the tests after they have built the code and pre-processed the metrics .in files
# cd tests/metrics
# bash ./network/network-latency.sh
```

grahamwhaley:
OK, did a run test - fine. the SD is still a bit high for my liking, but let's see how it pans out, and we can try to adjust in the future if we find we need to.
lgtm

PR #871, opened 2017-05-03
Metrics: Remove iperf2 tests

GabyCT:
Remove iperf2 metrics as they are already implemented
using iperf3 (see #781)

### DIFF
--- a/tests/metrics/network/network-metrics.sh
+++ b/tests/metrics/network/network-metrics.sh
@@ -19,8 +19,7 @@
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 # Description:
-#  Test inter (docker<->docker) network bandwidths and jitters
-# using iperf2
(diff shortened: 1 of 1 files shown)


tests/metrics/network/network-metrics.sh
```diff
@@ -19,8 +19,7 @@
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 # Description:
-#  Test inter (docker<->docker) network bandwidths and jitters
-# using iperf2
+#  Test inter (docker<->docker) bidirectional network bandwidth using iperf2
```
jcvenegas: is iperf2 ok  in the description ?

jcvenegas:
just one question but
lgtm

tests/metrics/network/network-metrics.sh
GabyCT: @jcvenegas yes it is ok, we are removing the rests of the test keeping only the bidirectional one :)

chavafg:
qa-passed

PR #873, opened 2017-05-04
CI: Add script to perform basic checks on commits.

### LINKED ISSUE #872: Create CI pre-checks script to check git commits
Create a script that performs basic checks on the new commits to be merged.

Script can check for things like a `Fixes #XXX` tag, subsystem in title, *etc*.

Script will be called as part of the CI workflow crucially before any setup/build/tests run.

See #856.

jodh-intel:
The script is called before the CI runs any setup/build/tests and
performs the following checks on all commits:

- Subject
  - Contains a subsystem.
  - Length is within prescribed limits.

- Body
  - Contains a "Fixes #XXX" comment.
  - Line lengths are within prescribed limits.

Fixes #872.

### DIFF
--- a/.ci/ci-pre-checks.sh
+++ b/.ci/ci-pre-checks.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#---------------------------------------------------------------------
+# Description: Perform basic checks on the branch before attempting
+#   to run the build and test suites. If this script fails, the CI run
+#   should be aborted.
+#---------------------------------------------------------------------
+
+typeset script_name=${0##*/}
+
+# Limits defined in CONTRIBUTING.md
+typeset -i max_subject_length=75
+typeset -i max_body_line_length=72
+
+# Latest commit (HEAD) in current branch
+typeset commit
+
+# "Destination" git branch that the new commit(s) in the
+# current branch want to be merged into.
+typeset branch
+
+# Set to 'yes' when a commit is found that specifies an
+# issue this PR fixes.
+typeset found_fixes=no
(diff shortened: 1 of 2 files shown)


.ci/ci-pre-checks.sh
```diff
@@ -0,0 +1,146 @@
...
+
+    [ -n "$body" ] || die "no body"
+
+    # Load the body text into an array
+    mapfile -t lines < <( echo "$body" )
+
+    local found_fixes=no
+    for line in "${lines[@]}"
+    do
+        echo "$line"|egrep -iq "\<fixes\> *#[0-9][0-9]*"
```
grahamwhaley: Just a note/clarification that we are fixing our requirement here to 'fixes', whereas in fact github has a slew of words that can be used to close a commit: https://help.github.com/articles/closing-issues-via-commit-messages/
So, some acks that this is our intention would be good.

.ci/ci-pre-checks.sh
```diff
@@ -0,0 +1,146 @@
...
+        if [ $? -eq 0 ]
+        then
+            found_fixes=yes
```
grahamwhaley: I guess you could `break` here as well once you have spotted an occurrence.

jodh-intel: Actually no as that would mean the line length check below wouldn't be performed.

grahamwhaley: ah, indeed - mea culpa

.ci/ci-pre-checks.sh
jodh-intel: Yes I thought about this as I wrote that line. Happy to change to accept all possibilities if that's what we want. But in some ways its probably easier to standardise on a single one. Added to which we do specify "Fixes #XXX" (only) in `CONTRIBUTING.md`.

chavafg:
qa-passed

amshinde:
Having a strict requirement of having "Fixes" in the commit message, and having to create an issue for every small thing such as a typo in a readme seems like an overkill to me. I feel this is something that should be part of gatekeeping for significant changes.

chavafg:
qa-passed

jodh-intel:
@amshinde - this script simply automates checking the requirements we have already specified in `CONTRIBUTING.md`.

The point of this script is twofold:

- To ensure a more consistent and useful git history.
- To help the Gatekeepers buy saving them from having to check things that a program can do for them! :smile:

As explained on #855, if you have an issue assigned to you, it is clear what you are working on which will minimise the chance that we end up with multiple person attempting to submit a PR on the same topic.

Yes, it's mildly annoying having to raise an issue for a 1-line fix, but that is the small price we pay for improving our process. After all, it is the matter of a merely a few seconds to raise an issue (probably less if you use a tool like https://github.com/node-gh/gh#1-create).

If you disagree with our documented process, please raise an issue so we can discuss it there.

grahamwhaley:
I agree that, particularly in this more stable phase of the project, I'd rather have a little more process (sometimes that being a little annoying) to ensure quality and we don't miss things, rather than the opposite. I say we ack this PR, and see how it goes - if it becomes an issue that we can see is slowing development (as I doubt it is going to harm quality), then we can always later modify or revert it.

chavafg:
qa-passed

jcvenegas:
lgtm

Issue #875, opened 2017-05-04
Make it easier to install autoconf-archive m4 macros


grahamwhaley:
Now we have removed the m4 macros we had extracted from the autoconf-archive, and rely on either the host system having them pre-installed or for an installation script to install them by hand if necessary (such as seen here https://github.com/01org/cc-oci-runtime/blob/master/installation/rhel-setup.sh#L161), there are times during development when you'd just like to do that curl hand install 'by hand'. Thus, it would be good to extract that 'curl magic' into a separate script that can be both utilised by the install scripts as per above, or run by hand before a configure to get the relevant macros installed in place.

dlespiau:
What about making installing that new build dependency as part of autogen.sh? that's the script supposed to bootstrap the build system after all.

jodh-intel:
That's an interesting point. I've always seen `autogen.sh` as something that attempts to make the source buildable, crucially without changing the system state. It might be rather unexpected if that script starts installing packages.

Maybe a compromise could be to add a `--make-it-so`-type option. If not specified, `autogen.sh` will be have like it does now and fail if dependencies are not installed (but also make mention of the new option). If that new option is specified, the script will attempt to install missing packages.

Note that in the latter case, we'd have to add in distro-specific commands to install the packaged `autoconf-archive`, or download if not available. In fact, we'd also need to rework the `.ci/` and the `installation/` scripts since they call `autogen.sh` and have `autogen.sh` source files as appropriate.
Finally, the docs will need to be updated to take account of such changes (including detailing the new `--make-it-so`-type option).

PR #876, opened 2017-05-04
install: Extract autoconf-archive install to separate script

grahamwhaley:
Under some circumstances it is needed to 'hand install' the m4
scripts. This was done in the rhel install script, but sometimes
it is really useful to be able to do that by hand, so extract the
relevant commands into their own script (to allow hand running),
and get the rhel script to invoke that new script at the appropriate
time.

Fixes: #875

### DIFF
--- a/installation/curl-autoconf-archive.sh
+++ b/installation/curl-autoconf-archive.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# Description: This script downloads and installs the autoconf-archive macros
+#   necessary to configure this project. This is invoked for distro builds where
+#   the necessary package is not available to install, and is also useful to
+#   hand run under certain circumstances.
+#   Should be run from the top level directory
+#
+
+# autoconf-archive url
+autoconf_archive_url="http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4"
+mkdir -p m4/
+# curl the required autoconf archive files into the correct place
+curl -L "${autoconf_archive_url}/ax_code_coverage.m4" -o m4/ax_code_coverage.m4
+curl -L "${autoconf_archive_url}/ax_valgrind_check.m4" -o m4/ax_valgrind_check.m4
--- a/installation/rhel-setup.sh
+++ b/installation/rhel-setup.sh
@@ -152,15 +152,12 @@ cor=github.com/01org/cc-oci-runtime
 # Currently it is the latest version for cc-oci-runtime
 commit=tags/2.1.9
 release=${commit##*/}
-# autoconf-archive url
-autoconf_archive_url="http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4"
 go get "$cor"
 pushd "$GOPATH/src/$cor"
 git checkout -b "$release" "$commit"
-mkdir -p m4/
-# autoconf-archive package does not exist in RHEL we need to download all m4 files required
-curl -L "${autoconf_archive_url}/ax_code_coverage.m4" -o m4/ax_code_coverage.m4
-curl -L "${autoconf_archive_url}/ax_valgrind_check.m4" -o m4/ax_valgrind_check.m4
+# autoconf-archive package does not exist in RHEL we need to download all m4
+# files required
+source "${SCRIPT_PATH}/curl-autoconf-archive.sh"
 ./autogen.sh \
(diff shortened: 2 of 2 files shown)


grahamwhaley:
Marked this as WIP, as I am not set up to re-test the modified rhel install scripts. @devimc or @chavafg - are either of you in a position to try this out for me?

installation/curl-autoconf-archive.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Description: This script downloads and installs the autoconf-archive macros
+#   necessary to configure this project. This is invoked for distro builds where
+#   the necessary package is not available to install, and is also useful to
+#   hand run under certain circumstances.
+#   Should be run from the top level directory
+#
+
+# autoconf-archive url
+autoconf_archive_url="http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4"
```
jcvenegas: I recently added autoconf_archive to Clear Linux, what do you think to chanage that url with something like this
https://ftp.gnu.org/gnu/autoconf-archive/autoconf-archive-2017.03.21.tar.xz

jcvenegas:
Looks good looks very helpful !  I think it could be nice to have some documentation about it somewhere or add an option to configure.ac to do something like:
```
./configure --with-download-autoconf-archive
```

installation/curl-autoconf-archive.sh
grahamwhaley: I think I like the versioned link better than the gitweb one, yes.
What do you think - we can just pull the tar file and extract just the files we need into the m4 directory with a single piped tar command maybe?

jodh-intel: I think this decision effectively summarises to:

- do we want to remain on the development tip of autoconf-archive (gitweb)? or,
- do we want stability at the expense of using old code (ftp)?

I'd leaning slightly towards the former since otherwise we have something else to maintain (if we go with the latter we should add the package version to `versions.txt`). If the GNU tip does break, we can raise a PR to revert to the last good commit. That said, I can't imagine that the two m4 files we're using are going to change significantly enough to warrant tracking the top of development. However, by remaining current, we'll ensure RHEL/CentOS are more in line with the other distros (which presumably will update the `autoconf-archive` package as new versions are released upstream.

grahamwhaley: Nice breakdown @jodh-intel  With that info, I think I agree that longer term the development tip is safe enough (these are very low churn files)...
@jcvenegas , what say you?

jcvenegas: @grahamwhaley @jodh-intel , not problem here, I was curious to  be a  consistent with the versions.txt, it already includes a version for "autoconf_archive_version", and in case of that happen,  we potentially could know about it and  fix it quickly

grahamwhaley:
@jcvenegas I mocked up some docs locally that lived as an indented paragraph in here: https://github.com/01org/cc-oci-runtime/blame/master/README.rst#L209 but, it feels odd, and I'm not sure that is the correct or best place. @jodh-intel any thoughts on this?
We have issues that contain the trail with error and workaround, so I'm not unhappy to proceed without an explicit document being present.

grahamwhaley:
@jcvenegas @devimc @jodh-intel For the autoconf addition... two things:
* I'll be interested if somebody has a view here - I have no objection, just want to collect a collective view
* I don't have the hands on auto-magic skills to easily locate where and how to implement this - if somebody wants to give some sample code/patch then please feel free - I could either then merge that into my PR, or we can merge this PR and you can build on top of it.

jcvenegas:
@grahamwhaley  sure lets make the autoconf thing not a blocker, we can added it latter if needed

grahamwhaley:
Just to nudge this forwards, here is the doc snippet (in a temporary branch/commit) I tried out.
https://github.com/grahamwhaley/cc-oci-runtime/commit/92a9ef715312ab89fbe3cfbe544e818a7a2dc2df
I'm not too happy with that myself, so;
 - if somebody can come up with a better place or way to doc it, please suggest
 - or, given the details of the problem and workaround are in the github history where folks might search for such things, we could just go ahead and commit as is
 - or, if this patch set is being outdated by the pending autoconf archive versions PR #911 , then once that is sorted out and merged let me know and we can close this one down.

grahamwhaley:
@gorozco1 @jodh-intel with #911 , is this PR now redundant - shall I shut it down?

jodh-intel:
I think it is still useful for RHEL and CentOS which don't provide an `autoconf-archive` package.

jodh-intel:
Due to the problems we're experiencing with the http://savannah.gnu.org/ site (see #946), I think this script is definitely needed, but should download the whole archive and extract just the files we need. A mirror like `kernel.org` (http://mirrors.kernel.org/gnu/autoconf-archive/) would seem like a good bet for reliability but we'll have to add "some" version of `autoconf-archive` to `versions.txt`.

gorozco1:
@grahamwhaley I think this script is still needed for RHEL/CentOS

jodh-intel:
Update: savannah is back so we have some thinking time now ;)

grahamwhaley:
This has been hanging here for some time - it is 'self documenting' in the Issue, and mostly we have moved on past this issue, but the script is still useful. Drop the DNM tag, and let's merge it.

jodh-intel:
lgtm

chavafg:
qa-failed

jcvenegas:
@chavafg any information why  tests failed ?

chavafg:
Seems that PR needs a rebase, so it runs the `swarm` tests with the `nginx` image that we currently support.

jcvenegas:
@grahamwhaley  could you rebase the PR ?

grahamwhaley:
Rebase done and pushed. I am not set up to easily verify this patch at present, but a visual check shows the changes appear identical after the rebase. If somebody feels they are set up and would like to do a test run of the rhel installer, please do.

chavafg:
qa-failed

chavafg:
this is a qa-passed, but for some reason old logs didn't get cleared and they make this failed. Will look into the ci issue.

chavafg:
qa-passed

devimc:
lgtm

Issue #877, opened 2017-05-05
Pass debug flag to the shim when the runtime is invoked with --debug


amshinde:
We should be able to see debug logs from the shim as well, when the runtime in invoked with --debug option.

PR #878, opened 2017-05-05
Pass debug flag to shim

amshinde:


chavafg:
qa-passed

jodh-intel:
Looks good, but please can you update the commit message to include `Fixes #877` (it's currently `Fixes 877` without the hash symbol, so won't auto-close the issue afaik).

dlespiau:
The only question I have is: are we happy with not printing errors to stderr anymore? it was quite handy and helped us notice errors that may have been silenced otherwise.

jcvenegas:
@amshinde  @dlespiau  That  is a good  question, I agree that is a useful option, and if the user is running with debug options . In case we dont want it because could be a bit annoying we should add information  in: https://github.com/01org/cc-oci-runtime/wiki/Debugging or  https://github.com/01org/cc-oci-runtime/blob/master/shim/README.md

amshinde:
@jodh-intel Commit message has been updated.
@dlespiau @jcvenegas Error messages will still be printed to stderr

dlespiau:
oh!

lgtm

jodh-intel:
lgtm

chavafg:
qa-passed

PR #885, opened 2017-05-05
Vhostuser network enabling

egernst:


egernst:
This is for the POC branch.  It is just a PR from what I already reference in the VPP CNM plugin.  This was only tested on 2.1.2; for the short term I'd rather have this merged at this baseline, and keep OVS/VPP/SRIOV all on the same baseline.  Once there I will work to rebase /test on a regular cadence.  Thoughts?

PR #889, opened 2017-05-06
build: Update versions.txt to include cc-oci-runtime version

gorozco1:
versions.txt should add cc-oci-runtime version.

Automation needs an easy way to access the cc-oci-runtime version
when building packages for diferent distros.

This will require that in future releases versions.txt must be
updated along with configure.ac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,8 +20,9 @@
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 
+
 AC_PREREQ([2.69])
-AC_INIT([cc-oci-runtime], [2.1.8])
+AC_INIT([cc-oci-runtime], m4_esyscmd_s([cat versions.txt  | grep cc_oci_runtime_version | cut -d'=' -f2]))
 AC_CONFIG_MACRO_DIR([m4])
 AM_INIT_AUTOMAKE([foreign -Wall -Werror -Wno-portability silent-rules subdir-objects color-tests no-dist-gzip dist-xz])
 AM_SILENT_RULES([yes])
--- a/versions.txt
+++ b/versions.txt
@@ -15,3 +15,4 @@ mpfr_version=3.1.4
 gmp_version=6.1.0
 mpc_version=1.0.3
 autoconf_archive_version=2017.03.21
+cc_oci_runtime_version=2.1.8


chavafg:
qa-passed

versions.txt
```diff
@@ -15,3 +15,4 @@ mpfr_version=3.1.4
 gmp_version=6.1.0
 mpc_version=1.0.3
 autoconf_archive_version=2017.03.21
+cc-oci-runtime=2.1.7
```
jodh-intel: This will have to be `cc_oci_version=2.1.7` to be a valid shell variable.

jodh-intel:
Hi @gorozco1 - It sounds like updating `versions.txt` for the runtime should be added to the wiki page:

- https://github.com/01org/cc-oci-runtime/wiki/Release-Checklist
  (along with mentioning updating `configure.ac`)

gorozco1:
@jodh-intel updated commit and wiki

jodh-intel:
Thanks.

lgtm

jcvenegas:
Looks like configure.ac could user the that new variable to avoid change in 2 places the project version.

jcvenegas:
@gorozco1 
I wonder if se could do source of  versions.txt before define the version:
source $srcdir/versions.txt
AC_INIT([cc-oci-runtime], [$cc_oci_runtime])

chavafg:
qa-passed

gorozco1:
@jcvenegas where? in configure.ac ?

jcvenegas:
@gorozco1  yes, actually is move the line : https://github.com/01org/cc-oci-runtime/blob/2.1.7/configure.ac#L35 before AC_INIT this will load the variable cc-oci-runtime that you could use instead to hardcode the version in AC_INIT

chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
qa-passed

jcvenegas:
lgtm

gorozco1:
can we merge this?

configure.ac
```diff
@@ -20,8 +20,9 @@
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 
+
 AC_PREREQ([2.69])
-AC_INIT([cc-oci-runtime], [2.1.7])
+AC_INIT([cc-oci-runtime], m4_esyscmd_s([cat versions.txt  | grep cc_oci_runtime_version | cut -d'=' -f2]))
```
jamesodhunt: You'll have to use `$srcdir/versions.txt` for this to be reliable. Also you could drop the `cat` and just:

```
grep cc_oci_runtime_version "$srcdir/versions.txt" | ...
```

jamesodhunt:
On other point - https://github.com/01org/cc-oci-runtime/wiki/Release-Checklist will need to be updated when this lands.

configure.ac
gorozco1: I think in this specific part of the configure.ac there is no `$srcdir` variable defined, it's the very beginning of `configure.ac`

jodh-intel: I'm rather surprised this works, since conventionally `configure.ac` should be updating `versions.txt.in`. @dlespiau , @devimc - do you have any thoughts on this approach? Anything that could "bite us" later on?

gorozco1: @devimc @dlespiau any comment regarding this commit?

jodh-intel:
I don't know how reliable this is going to be given that no path is specified for `versions.txt`. However, I've tested this and it does seem to work, so presumably there is a `cd $srcdir` happening somewhere. Hence, unless anyone else speaks up, once'we you've resolved the conflict, let's see if we can get the requisite acks.

dlespiau:
That does look like a scary approach. No idea if it's be guaranteed to work. Played with sed a bit, it's possible to grab the version from configure.ac somewhat easily:

```
$ sed -n -e 's/AC_INIT(.*,[ ]*\[\(.*\)\])/\1/p' configure.ac 
2.1.5
```

gorozco1:
@dlespiau @jamesodhunt I will close this looks like is not the right approach :) thanks for your feedback.

PR #897, opened 2017-05-10
Prevent race condition with new client ID

gtkramer:
In case two requests for serving clients come in at the same time, using a mutex synchronization primitive to guarantee that both requests do not get the same client ID

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -333,20 +333,22 @@ func (proxy *proxy) init() error {
 	return nil
 }
 
-var nextClientID = uint64(1)
+var nextClientID = uint64(0)
 
 func (proxy *proxy) serveNewClient(proto *protocol, newConn net.Conn) {
+	// Unfortunately it's hard to find out information on the peer
+	// at the other end of a unix socket. We use a per-client ID to
+	// identify connections.  This ID needs to be unique.  To ensure
+	// this, the ID is first incremented and then assigned to prevent
+	// two peers from having the same ID.
+	atomic.AddUint64(&nextClientID, 1)
+
 	newClient := &client{
 		id:    nextClientID,
 		proxy: proxy,
 		conn:  newConn,
 	}
 
-	atomic.AddUint64(&nextClientID, 1)
-
-	// Unfortunately it's hard to find out information on the peer
-	// at the other end of a unix socket. We use a per-client ID to
-	// identify connections.
 	newClient.info(1, "client connected")
 
 	if err := proto.Serve(newConn, newClient); err != nil && err != io.EOF {


dlespiau:
I'm a bit perplex here, there's an atomic operation which should make sure that we increment that variable safely. Did you observe two clients with the same ID in the logs?

dlespiau:
Oh I know! we increment the clientID *after* assigning it to the client structure, so two clients can connect and get assigned the same number. Adding a lock won't help here :) adding 1 to clientID is already atomic. The fix is to initialize the counter to 0 and increment it *before* assigned it to the client struct.

amshinde:
Yes, we just came across this issue, two goroutines could potentially get the same clientid.  The increment does need to happen before the assignment.

gtkramer:
amshinde and I just synced on the fix.  I'll put in another commit with the correct patch.

dlespiau:
Could you leave the logic incrementing `nextClientID` in the `serveNewClient` function? I think it belongs to to same function creating a client object (eg. could be factored out into a NewClient() function later on).

dlespiau:
Just a couple of technicalities left: seems like a good idea to squash the two commits together in this case, there's no good reason to split this in two. Also we require signed-off-by tags for all contributions. With those two things fixed, this is:

lgtm

gtkramer:
I force pushed a rebase before this last commit to make sure that all commits in this pull request have the signed-off-by part.  I was hoping the pull request could be merged without closing and reopening.  But the check for this hasn't passed yet.  Is this a problem?

dlespiau:
The first commit uses a different email address for the Author than the one in the s-o-b tag. IIRC that's something that will make the test fail.

Still need to squash the two commits as well.

proxy/proxy.go
```diff
@@ -333,17 +333,17 @@ func (proxy *proxy) init() error {
 	return nil
 }
 
-var nextClientID = uint64(1)
+var nextClientID = uint64(0)
 
 func (proxy *proxy) serveNewClient(proto *protocol, newConn net.Conn) {
+	atomic.AddUint64(&nextClientID, 1)
```
jodh-intel: I think this line deserves a comment explaining why the increment is happening here.

jodh-intel:
I agree with @dlespiau - please squash the commits into one.

proxy/proxy.go
gtkramer: Yes, it would make it more clear for other readers.  I'll insert a comment about the potential for a race condition here and squash all of the commits into one.

jcvenegas:
Seems that this pr is  almost ready :D ! @gtkramer could you please squash the commits to merge thi PR ?

jodh-intel:
@gtkramer - any update on this?

gtkramer:
Looks like we're all set?  Checks are pending.

dlespiau:
lgtm

gtkramer:
Thank you @dlespiau @amshinde @jodh-intel @jcvenegas for helping me with my first commit to open source!  Much appreciated.

chavafg:
qa-passed

jcvenegas:
lgtm

dlespiau:
\o/ hopefully it wasn't too scary. I cherry-picked this commit for 3.0 as well. https://github.com/clearcontainers/proxy/pull/54

gtkramer:
It was exciting more than anything else!

gtkramer:
I was thinking about this patch last night.  Did it fully solve the problem?  Don't we want something like:
mutex.Lock
Increment ID
Assign ID
mutex.Unlock

Today it's:
mutex.Lock
Increment ID
mutex.Unlock
Assign ID

It looks like there's still a small gap before the ID is assigned where a race condition could occur that ends up with two clients having the same ID.

dlespiau:
Absolutely, in fact I realised that yesterday as well. That's the reason atomic.Add* operations return the value. I fixed it for 3.0 in a PR yet to come, mind sending a PR for the 2.x repo?

```diff
        // identify connections.  This ID needs to be unique.  To ensure
        // this, the ID is first incremented and then assigned to prevent
        // two peers from having the same ID.
-       atomic.AddUint64(&nextClientID, 1)
+       id := atomic.AddUint64(&nextClientID, 1)
 
        return &client{
-               id:    nextClientID,
+               id:    id,
                proxy: proxy,
                conn:  conn,
                kind:  clientKindRuntime,
        }
 }
```

gtkramer:
Sure thing.  Should a separate PR be opened?

dlespiau:
Yes please, separate PR.

gtkramer:
Will do.  I'll submit the PR tonight.

PR #905, opened 2017-05-15
RFC: srio-v enabling on 2.1.11

egernst:
Testing and debug of an issue with the CNM plugin for SRIO-V is still underway, but since this adds a substantial amount of code, I wanted to be more proactive on posting it for feedback.

This could be squashed later, but for now I wanted to preserve the history so you can see what was being taken from 1.x COR versus what needed to be cleaned up.  I also wanted to make sure original author is credited.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -24,6 +24,7 @@
 #include "config.h"
 #include "state.h"
 #include "oci-config.h"
+#include "hypervisor.h"
 
 #include <errno.h>
 #include <glib/gstdio.h>
@@ -155,6 +156,7 @@ handle_command_stop (const struct subcommand *sub,
 	gboolean           ret;
 	GNode*             root = NULL;
 	gchar             *cgroup_dir = NULL;
+	guint		  index;
 
 	g_assert (sub);
 	g_assert (config);
@@ -199,14 +201,53 @@ handle_command_stop (const struct subcommand *sub,
 
 	g_free_node(root);
 
+	/* Check for devices which were passed through to the clear container
+	 * If found, bind it back to the original host driver and change the
+	 * interface back to the child process' namespace
+	 */
+	if (state->devices) {
+
+		/* Flag to mark successful unbind/bind back to original device
+		 * driver configuraiton.  If none of the devices transfer correctly,
+		 * then return unsuccessfully
+		 */
+		gboolean device_bind_success = false;
+
+		for (index=0; index<g_slist_length(state->devices); index++) {
+			struct cc_oci_device* device = (struct cc_oci_device *)
+				g_slist_nth_data(state->devices, index);
+
+			if (!cc_oci_bind_host(device)) {
+				g_debug("failed to bind device %s back to host", device->bdf);
+				continue;
+			}
+
+			/* brute-force hacks for changing the interface to
+			 * child process network namespace
+			 */
+			if (!cc_oci_switch_iface_to_container(device, state->pid)) {
+				g_debug ("failed to switch device %s to container",device->bdf);
+				continue;
+			}
+
+			device_bind_success = true;
+		}
+
+		if (device_bind_success == false) {
+			goto out;
+		}
+	}
+
 	if (! cc_oci_config_update (config, state)) {
 		goto out;
 	}
 
 	ret = cc_oci_stop (config, state);
+
 	if (! ret) {
 		goto out;
 	}
+
 
 	ret = true;
 
--- a/src/hypervisor.c
+++ b/src/hypervisor.c
@@ -21,6 +21,7 @@
 #include <stdbool.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <unistd.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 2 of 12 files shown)


src/networking.c
```diff
@@ -412,6 +445,28 @@ cc_oci_network_create(const struct cc_oci_config *const config,
...
+        for (i=0; i < g_slist_length(config->net.interfaces); i++) {
+                if_cfg = (struct cc_oci_net_if_cfg *)
+                        g_slist_nth_data(config->net.interfaces, i);
+
+                if (if_cfg->vf_based) {
+                        g_debug("interface %s is vf based vf: bdf %s driver %s",if_cfg->ifname, 
+                               if_cfg->bdf, if_cfg->device_driver);  
+                        device = json_object_new ();
+                        json_object_set_string_member (device, "bdf", if_cfg->bdf);
+                        json_object_set_string_member (device, "driver", if_cfg->device_driver);
```
jodh-intel: Can't the `device` be returned here (and the one at the end just be `return NULL`)?

src/oci.c
```diff
@@ -1628,6 +1628,12 @@ cc_oci_config_update (struct cc_oci_config *config,
 		g_free_if_set (state->process);
 	}
 
+        if (state->devices) {
+               config->devices = state->devices;
+               state->devices = NULL;
+               g_debug ("cor non-null devices from state");
```
jodh-intel: Can this `g_debug()` call be dropped?

src/oci.h
```diff
@@ -359,11 +359,21 @@ struct cc_oci_net_if_cfg {
...
+
+	/** device driver associated with this VF device */
+        gchar *device_driver;
+
 	/** List of IPv4 addresses on the interface */
 	GSList  *ipv4_addrs;
 
 	/** List of IPv6 addresses on the interface */
 	GSList  *ipv6_addrs;
+
```
jodh-intel: Superfluous blank line?

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
...
+	gchar *sym_link = NULL;
+	GError *link_error = NULL;
+	gchar **tokens = NULL;
+	gchar **driver_tokens = NULL;
+	gchar **mac_from_file = NULL;
+
+	gboolean retval = false;
+	guint idx=0;
+
+	g_debug("\n=== get mac and bdf for device %s", dirname);
```
jodh-intel: I'd drop the `\n` as it will look very odd in the logs and confuse tools that parse the logs.

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
...
+GHashTable* mac_hash;
+
+void
+hash_mac_and_store_bdf(GHashTable* hash, gpointer mac, gpointer bdf)
+{
+       g_hash_table_insert(hash, mac, bdf);
+}
+
+static gboolean
+get_mac_and_bdf(gchar *dirname, struct cc_oci_device *dev_info, gchar **mac)
```
jodh-intel: Arg checks?

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
 	g_free(str);
 }
 
+GHashTable* mac_hash;
```
jodh-intel: Can you add a comment on this global?

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
...
+
+void
```
jodh-intel: Function header.

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
...
+			g_debug ("sym_link read failure. Path: %s", tmp_path);
+		}
+		goto out;
+	}
+
+	/*
+	 * parse the sym_link in order to grab the bdf and make sure it is a
+	 * pci device.  If it is a virtual device, skip out.
+	 */
+	tokens = g_strsplit (sym_link, "/", 8);
```
jodh-intel: That `8` deserves a comment. Also, it would be worth validating that you actually get `8` values back before using `tokens[3]` below.

src/state.c
```diff
@@ -285,6 +288,31 @@ handle_state_mounts_section(GNode* node, struct handler_data* data) {
 	}
 }
 
+static void
```
jodh-intel: Function header.

src/process.c
```diff
@@ -771,6 +771,174 @@ cc_free_pointer(gpointer str)
...
+
+void
+build_mac_to_bdf_hash (void)
+{
+	gchar *mac = NULL;
+	struct cc_oci_device *d_info = NULL;
+	DIR *d;
+	struct dirent *dir;
+
+	mac_hash = g_hash_table_new(g_str_hash, g_str_equal);
```
jodh-intel: There doesn't seem to be a corresponding call to `g_hash_table_destroy()`?

src/command.c
```diff
@@ -204,10 +206,30 @@ handle_command_stop (const struct subcommand *sub,
 	}
 
 	ret = cc_oci_stop (config, state);
+
+        if (state->devices) {
+                /* loading appropriate drivers on the devices */
+		g_debug ("loading drivers on the devices that pass-thru'ed");
```
egernst: Hahaha -- Yes, will do this, agreed - some of the wording bothered me before, but I wanted someone from England to point out the English :)

You have any favorite (or is it favourite) spell checker for source?

jodh-intel: For the message, how about simply `loading drivers for pass-through devices`?

I use vim's `:set spell spelllang=en_gb` which flags this word (as being from another region).

src/hypervisor.c
```diff
@@ -46,6 +47,228 @@
...
+/*!
+ * Unbind a given interface from the host and add it as
+ * a vfio pci device for the guest.
+ *
+ * \param config \ref cc_oci_config*
+ * \param index \ref guint
+ *
+ */
+gboolean
+cc_oci_unbind_host(struct cc_oci_config *config, guint index)
```
egernst: Agreed- will do a third patch set to take this and some of the functional changes into account.

egernst:
@jodh-intel I took some of the feedback into account and will continue this early next week.  I pushed now so there's a good baseline that can be used for demonstrative purposes (the old one had stability issues due to poor tear down handling).  Thanks for the help - no need to rereview this.

jodh-intel:
Hi @egernst - any update on this? Should this be marked `do-not-merge` or are you still updating it?

egernst:
Note, the sriov branch needs to be reset to 2.1.11 to facilitate this.

src/process.c
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+	device_dir_path = g_strdup_printf("/sys/class/net/%s/device", sym_link);
+	sym_link = g_file_read_link(device_dir_path, &link_error);
+	if (!sym_link) {
+		if (link_error) {
+			g_debug ("device relative-path sym_link read failure. Path: %s", tmp_path);
+		}
+		goto out;
+	}
+
+	tokens = g_strsplit (sym_link, "/", 4);
```
jodh-intel: This was already defined so it looks like you're missing a call to `g_strfreev()`.

src/process.c
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+		goto out;
+	}
+
+	/* Grab device ID */
+	tmp_path = g_strdup_printf("%s/device", device_dir_path);
+	if ( !(g_file_get_contents(tmp_path, &buffer, NULL, NULL)) ) {
+		g_debug("Reading file %s returned error", tmp_path);
+		goto out;
+	}
+	g_free_if_set(tmp_path);
```
jodh-intel: There are a lot of calls to `g_free()` and `g_free_if_set()` in this function which makes it look quite "busy". You might want to declare these variables like this to avoid having to constantly free their values:

```
g_autofree gchar  *foo = NULL;
```

src/process.c
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+
+	/* pending the device, the location of driver token will vary, but
+	 * will always be the last token
+	 */
+
+	if (!driver_tokens) {
+		g_debug ("unexpected driver-path format for %s", sym_link);
+		goto out;
+	}
+	while( driver_tokens[idx] )
```
jodh-intel: If you know there are 8 fields, it might be better to use a `for` loop to avoid any possibility of an infinite loop here.

src/process.c
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+ * associated with the device
+ *
+ * \param dirname \ref gchar*
+ * \param dev_info \ref cc_oci_device to return BDF and driver info
+ * \param mac \ ref gchar** mac address to return found MAC into
+ *
+ * \return \c true if succesfully found mac/bdf
+ */
+static gboolean
+get_mac_and_bdf(gchar *dirname, struct cc_oci_device *dev_info, gchar **mac)
```
jodh-intel: No check on `dirname`, `dev_info` or `mac`.

src/process.c
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+	g_free(key);
+	g_free(value);
+}
+
+	
+void
+mac_hash_free (void)
+{
+	/* free each bdf/mac entry */
+	g_hash_table_foreach(mac_hash, hash_table_free_entry, NULL);
```
jodh-intel: Might be prudent to check the hash before operating on it. Something like:

```

if (mac_hash != NULL) {
    g_hash_table_foreach(mac_hash, hash_table_free_entry, NULL);
    g_hash_table_destroy(mac_hash);
} else {
    g_critical("invalid hash table")
}

File: src/process.c [hunk 18]
```diff
@@ -778,6 +778,246 @@ cc_free_pointer(gpointer str)
...
+build_mac_to_bdf_hash (void)
+{
+	gchar *mac = NULL;
+	struct cc_oci_device *d_info = NULL;
+	DIR *d;
+	struct dirent *dir;
+
+	mac_hash = g_hash_table_new(g_str_hash, g_str_equal);
+
+	d = opendir("/sys/class/net");
```
jodh-intel: Again, I'd use the glib routines here to avoid having to use boiler-plate code.

src/state.c
```diff
@@ -804,6 +838,28 @@ cc_oci_state_free (struct oci_state *state)
 	g_free (state);
 }
 
+static JsonObject *
```
jodh-intel: Missing function header.

src/state.c
```diff
@@ -804,6 +838,28 @@ cc_oci_state_free (struct oci_state *state)
...
+cc_oci_devices_to_json (const struct cc_oci_config *config)
```
jodh-intel: Missing check on `config`.

src/state.c
```diff
@@ -285,6 +288,37 @@ handle_state_mounts_section(GNode* node, struct handler_data* data) {
 	}
 }
 
+static void
+handle_state_devices_section(GNode* node, struct handler_data* data) {
```
jodh-intel: No check on params - we need every function to check all its parameters to avoid crashes ;)

src/process.c
egernst: I wasn't aware of this.  Makes me nervous putting trust into the library -- do they handle even the case where you are doing multiple mallocs, though, or will it just do the cleanup once the scope ends for the most recent?

src/process.c
egernst: ACK

src/command.c
```diff
@@ -199,15 +201,37 @@ handle_command_stop (const struct subcommand *sub,
...
+				g_slist_nth_data(state->devices, index);
+
+                        if (!cc_oci_bind_host(device))
+				g_debug("failed to bind device %s back to host", device->bdf);
+
+			/* brute-force hacks for changing the interface to
+			 * child process network namespace
+			 */
+			if (!cc_oci_switch_iface_to_container(device, state->pid))
+				g_debug ("failed to switch device %s to container",device->bdf);
```
egernst: Similar logic to above -- I think its best to finish looping and handling the other devices.

src/process.c
jodh-intel: Sorry - yes, with glib you will still need to call `g_free*()` for all but the penultimate allocation as you'll only get auto-free when the variable goes out of scope (I was thinking about the NIH library, which is cleverer than glib ;-)

src/hypervisor.c
```diff
@@ -46,11 +47,267 @@
...
+		goto out;
+	}
+
+	/* call tear down script */
+	char * argv[4];
+	gchar *std_output = NULL;
+	gchar *std_err = NULL;
+	int exit_status=0;
+	pid_str = g_strdup_printf("%d", pid);
+	argv[0] = SRIOV_TEARDOWN_SCRIPT;
```
egernst: K - done!

egernst:
@jodh-intel thanks for the help.  I made those two changes requested.  If feasible, i'd like to merge this to the networking/sriov branch very soon, as I want to point other people to this, rather  than the MUCH worse code that doesn't work that exists there today.

Issue #893, opened 2017-05-09
cannot compile under Centos 7


chavafg:
when running `make` under Centos 7, I got this error:

```
src/netlink.c: In function 'netlink_link_add_bridge':
src/netlink.c:254:25: error: 'IFLA_BR_MCAST_SNOOPING' undeclared (first use in this function)
    mnl_attr_put_u8(nlh, IFLA_BR_MCAST_SNOOPING, 0);
                         ^~~~~~~~~~~~~~~~~~~~~~
src/netlink.c:254:25: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [src/cc_oci_runtime-netlink.o] Error 1
make[1]: Leaving directory `/home/cloud/go/src/github/01org/cc-oci-runtime'
make: *** [all] Error 2
```
Centos 7 kernel version:
```
[cloud@centos7-dm cc-oci-runtime]$ uname -r
3.10.0-514.10.2.el7.x86_64 
```

Related to https://github.com/01org/cc-oci-runtime/commit/001e2648f540ed7672ae67c7f3c369120a0c25fc

dlespiau:
Could you attach the file /usr/include/linux/if_link.h that Centos 7 has?

chavafg:
Sure, here it is: 
[if_link.h.txt](https://github.com/01org/cc-oci-runtime/files/987537/if_link.h.txt)

GabyCT:
@amshinde, we are still having issues in the compilation with centos 7.0
after doing the ./autogen.sh ... and doing the make -j, we have the following error

src/netlink.c: In function 'netlink_link_add_bridge':
src/netlink.c:269:25: error: 'IFLA_BR_MCAST_SNOOPING' undeclared (first use in this function)
    mnl_attr_put_u8(nlh, IFLA_BR_MCAST_SNOOPING, 0);
                         ^~~~~~~~~~~~~~~~~~~~~~
src/netlink.c:269:25: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [src/cc_oci_runtime-netlink.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory `/root/go/src/github.com/01org/cc-oci-runtime'
make: *** [all] Error 2

jcvenegas:
Fixed via #907

PR #907, opened 2017-05-15
networking: Fix compile for bridge networking code

amshinde:
Get rid of undef and rely on the definition we provide
for muticast snooping. This was an oversight in realizing that
the undef happens unconditionally at compile time.

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -253,14 +253,6 @@ netlink_link_add_bridge(struct netlink_handle *const hndl,
 	 */
 	#ifdef IFLA_BR_MAX
 		if (IFLA_BR_MAX > 23) {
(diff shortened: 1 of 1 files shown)


dlespiau:
The Story Of the Awful Netlink API. Is this the end? who knows!

lgtm

chavafg:
qa-passed

GabyCT:
@amshinde, I tested in Centos 7 and it is working! :) thanks

jcvenegas:
looks that works for gaby using CentOS, compiles in travis and our CI. So, looks that now compiles in all the places we want. 
lgtm

amshinde:
Fixes #893

PR #916, opened 2017-05-18
tests: updated user.bats to check group id with a regex

chavafg:
The test 'Running with additional groups' uses busybox image
which previously used id 99 for nogroup. In its latest
update, busybox now uses id 65534 as other distros do.

The test now uses a regex for verifying this id.


chavafg:
qa-passed

jcvenegas:
lgtm

jodh-intel:
Nice.

lgtm

PR #918, opened 2017-05-18
proxy: Avoid shadowing type "header" with variable "header"

amshinde:
Issue detected with vetshadow. Rename the variable to avoid
shadowing the type.


jcvenegas:
lgtm

dlespiau:
I wonder how is that a problem, any idea? the compiler doesn't get confused, we can do the same in C as well. Looks like a tooling problem rather than a real problem.

chavafg:
qa-passed

jodh-intel:
Agreed, shouldn't be a problem, but no harm in making the code clearer by using unique names.

lgtm

PR #929, opened 2017-05-23
Copy client ID locally to prevent race condition

gtkramer:
The previous pull request #897 is incomplete. The potential for a race
condition still exists because there is a gap between when the client ID
is incremented and when the client ID is later assigned. Copying the
return value from the atomic intermenting operation as a local variable
in the function to close the gap.


dlespiau:
lgtm

jodh-intel:
lgtm

chavafg:
qa-passed

Issue #944, opened 2017-06-05
Docs: network metrics: Run instructions have a typo, and some clarification


grahamwhaley:
On:
https://github.com/01org/cc-oci-runtime/blob/master/tests/metrics/network/README.md
The command:
`$ sudo -E make metrics tests`
should be:
`$ sudo -E make metrics-tests`

We should also note that that command not only builds the necessary test files, but will run them as well (which is not mentioned in the current documentation).

Ideally the `preprocess` and `run` actions would be separate steps in the `Makefile`, and then we could document them as such (obviously with the generate step being a dependence of the run step - so `run` would do a generate and then run) - that would also be really useful for developers (to be able to do the generate step separately from the run step), so they can then run individual tests by hand - but, that is a fix for another day.

PR #946, opened 2017-06-05
Docs: metrics: network: Clarify pre-process and running

grahamwhaley:
Correct the typo in the network metrics build command, and clarify
how the tests can be run by hand after pre-processing.

Fixes: #944

### DIFF
--- a/tests/metrics/network/README.md
+++ b/tests/metrics/network/README.md
@@ -63,16 +63,21 @@ RSS is a standard measure to monitor memory usage in a physical memory scheme.
 
 ## Running the networking tests
 
-1. Generate the necessary files to run the individual tests with the following command:
+Before the network tests can be run, some files must be pre-processed.
+The following command will both pre-process the required files and then run
+the tests:
 
 ```
-$ sudo -E make metrics tests
+$ sudo -E make metrics-tests
 
 ```
-2. Run the networking tests using `sudo` with the following commands:
+
+Once the necessary files have been pre-processed via the above command,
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

chavafg:
qa-passed

grahamwhaley:
Hi @jodh-intel @gorozco1 Looks like Travis is failing as the `curl` of the autoconf archive is giving a `502 Bad Gateway` error. Not sure if that is a temporary or permanent issue - any thoughts?
If you try the url:
`http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_valgrind_check.m4`
then what I see in the logs is a 193 byte download instead of a 8852 byte download - that was the clue ;-)

jodh-intel:
Hi @grahamwhaley - it does indeed look like that site is borked atm (you can't even download via cvs).

I think we might want to rethink our approach for obtaining `autoconf-archive` (https://github.com/01org/cc-oci-runtime/pull/876#pullrequestreview-36369774). It's "overkill" to download the whole archive when we only want 2 files, but if we're pulling a stable release, not only do we know which version of the code we're actually using, but we have more choices wrt a reliable download site due to large number of available mirrors.

jodh-intel:
... alternatively, we could just enable https://docs.travis-ci.com/user/caching/ so that once we have a good copy, we'll continue to use that.

GabyCT:
lgtm

PR #951, opened 2017-06-06
proxy/protocol: Add checks on header length and flags

dlespiau:
We can check better the header we receive from clients. This time, make
sure the payload length is within reasonable limits and that the flags
are 0 (we don't have any at the moment).

### DIFF
--- a/proxy/api/protocol.go
+++ b/proxy/api/protocol.go
@@ -18,6 +18,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 )
 
@@ -26,6 +27,18 @@ const headerLength = 8 // in bytes
 type header struct {
 	length uint32
 	flags  uint32
+}
+
+const maxPayloadLength = 1 * 1024 * 1024
+
+func (hdr *header) validate() error {
+	if hdr.length > maxPayloadLength {
+		return fmt.Errorf("payload size too big: %d (max: %d)", hdr.length, maxPayloadLength)
+	}
+	if hdr.flags != 0 {
+		return fmt.Errorf("unexpected flags: 0x%x", hdr.flags)
+	}
+	return nil
 }
 
 // A Request is a JSON message sent from a client to the proxy. This message
@@ -69,6 +82,10 @@ func ReadMessage(reader io.Reader, msg interface{}) error {
 		flags:  binary.BigEndian.Uint32(buf[4:8]),
 	}
 
+	if err := hdr.validate(); err != nil {
+		return err
+	}
+
 	received := 0
 	need := int(hdr.length)
 	data := make([]byte, need)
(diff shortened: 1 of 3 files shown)


chavafg:
qa-passed

proxy/api/protocol.go
```diff
@@ -28,6 +29,18 @@ type header struct {
 	flags  uint32
 }
 
+const maxPayloadLength = 1 * 1024 * 1024
```
jodh-intel: - What's the `1` for?
- Is there any value in making this a cli option?

proxy/api/protocol.go
```diff
@@ -28,6 +29,18 @@ type header struct {
...
+
+func (hdr *header) validate() error {
```
jodh-intel: Could you add a test for this?

proxy/api/protocol.go
```diff
@@ -28,6 +29,18 @@ type header struct {
...
+	if hdr.length > maxPayloadLength {
+		return fmt.Errorf("payload size too big: %d", hdr.length)
```
jodh-intel: It would be useful to include `maxPayloadLength` in the error message.

proxy/api/protocol.go
dlespiau: - that's for 1 MB
- I don't think so

jodh-intel: - I guess it does add a little clarity even though redundant (currently).
- Roger that.

proxy/api/protocol.go
dlespiau: ✓

proxy/api/protocol.go
dlespiau: ✓

jodh-intel:
lgtm

chavafg:
qa-passed

proxy/api/protocol.go
amshinde: The shim uses a different limit on the payload, namely the one specified by hyperstart.
https://github.com/01org/cc-oci-runtime/blob/master/shim/shim.h#L61.
We should be consistent here.

dlespiau: Except that this limit isn't the max length of a packet is it? just the length of the receive buffer. I'm hoping hyperstart can reconstruct bigger packets then the receive buffer size, but maybe that's too hopeful.

jcvenegas: @amshinde could you take a look to @dlespiau  comment ?  If you agreed give a approval :)

dlespiau: I was lazy yesterday, but it seems hyperstart will realloc the receive buffer (without any check one the new size, ooops) https://github.com/clearcontainers/hyperstart/blob/0.7.0-clearcontainers/src/init.c#L1193

amshinde: that needs to be fixed !

amshinde:
lgtm

PR #947, opened 2017-06-05
CI: Perform basic checks on PR commits.

jodh-intel:
Use the new "checkcommits" program in the "tests" repository
to perform basic checks on PR commits, such as:

- Subject
  - Contains a subsystem.
  - Length is within prescribed limits.

- Body
  - Contains a "Fixes #XXX" comment.
  - Line lengths are within prescribed limits.

Unlike the original implementation (#873), "checkcommits" should
be more robust.

Fixes #872.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #952, opened 2017-06-06
CI: Perform basic checks on PR commits.

### LINKED ISSUE #872: Create CI pre-checks script to check git commits

jodh-intel:
Use the new "checkcommits" program in the "tests" repository
to perform basic checks on PR commits, such as:

- Subject
  - Contains a subsystem.
  - Length is within prescribed limits.

- Body
  - Contains a "Fixes #XXX" comment.
  - Line lengths are within prescribed limits.

Unlike the original implementation (#873), "checkcommits" should
be more robust.

Fixes #872.

### DIFF
--- a/.ci/ci-pre-checks.sh
+++ b/.ci/ci-pre-checks.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#---------------------------------------------------------------------
+# Description: Perform basic checks on the branch before attempting
+#   to run the build and test suites. If this script fails, the CI run
+#   will be aborted.
+#---------------------------------------------------------------------
+
+repo="github.com/clearcontainers/tests/cmd/checkcommits"
+go get -d "$repo"
+(cd "$GOPATH/src/$repo" && make)
+checkcommits \
+    --need-fixes \
+    --need-sign-offs \
+    --body-length 72 \
+    --subject-length 75 \
+    --verbose
--- a/.ci/ci-setup.sh
+++ b/.ci/ci-setup.sh
@@ -22,6 +22,10 @@ set -e -x
 root=$(cd `dirname "$0"`/..; pwd -P)
 source "$root/versions.txt"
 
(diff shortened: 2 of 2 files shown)


jodh-intel:
Ahem... this replaces #947.

chavafg:
qa-passed

.ci/ci-pre-checks.sh
```diff
@@ -0,0 +1,29 @@
...
+
+#---------------------------------------------------------------------
+# Description: Perform basic checks on the branch before attempting
+#   to run the build and test suites. If this script fails, the CI run
+#   will be aborted.
+#---------------------------------------------------------------------
+
+repo="github.com/clearcontainers/tests/cmd/checkcommits"
+go get -d "$repo"
+(cd "$GOPATH/src/$repo" && make)
```
grahamwhaley: We may have covered this elsewhere already (possibly on the old ticket or irc), but why do we use `get -d` and then do a `make` by hand - is there a good reason we can't use the default `go get` action of download/build/install. the checkcommits readme.md even says to use 'got get' to build and install it?
Looking at the checkcommits Makefile - maybe it is the commit id extraction we are after or similar? Can you elaborate here a little just for clarity?

.ci/ci-pre-checks.sh
```diff
@@ -0,0 +1,29 @@
...
+checkcommits --verbose --need-fixes --need-sign-offs
```
grahamwhaley: Just to note - looks like we are taking the default line length and other settings from `checkcommits` here - that is probably fine today as the defaults of `checkcommits` match this project I believe - but - it might be 'safer' for the long term if we did wire the specific project requirements in here to insulate ourselves from any `checkcommits` changes in the future.

grahamwhaley:
lgtm - with feedback to take on board.

.ci/ci-pre-checks.sh
jodh-intel: We have to call make since we need to control the build options which are required to embed the version and commit id into the `checkcommits` binary:

- https://github.com/clearcontainers/tests/blob/master/cmd/checkcommits/Makefile#L12

.ci/ci-pre-checks.sh
jodh-intel: I can change that if you like? FTR, the defaults in `checkcommits` are our defaults from https://github.com/01org/cc-oci-runtime/blob/master/CONTRIBUTING.md.

.ci/ci-pre-checks.sh
grahamwhaley: Thanks. That is what I suspected. In which case, does this need updating?
https://github.com/clearcontainers/tests/tree/master/cmd/checkcommits#download-and-build

.ci/ci-pre-checks.sh
grahamwhaley: If you don't mind. Given `checkcommits` is in a different repo in a different org then I think it will be clearer and safer in the long term for little cost.

.ci/ci-pre-checks.sh
jodh-intel: Yes. I'll raise a PR on the project itself to correct the docs.

jodh-intel: PR https://github.com/clearcontainers/tests/pull/49 raised to fix the docs.

chavafg:
qa-passed

jcvenegas:
lgtm

Issue #945, opened 2017-06-05
tests: split pre-process and run into two Makefile targets


grahamwhaley:
Often during development it is useful to be able to pre-process the 'GENERATED' files without taking further actions (such as running the networking tests). There is no current single nice Makefile 'target' to do the pre-processing. It would be very nice to have a rule inserted between the run rule (such as `metrics-tests`, and the $(GENERATED_FILES) dependency, so one can do, for instance a 'make generated-files' etc.
As previously hinted at in #944

grahamwhaley:
@jodh-intel @jcvenegas can I ask for your thoughts on the best way to do this.
The simplest may be to just add:
```
generated_files: $(GENERATED_FILES)
```
at about https://github.com/01org/cc-oci-runtime/blob/master/Makefile.am#L98

I considered then changing the $(GENERATED_FILES) deps such as here:
https://github.com/01org/cc-oci-runtime/blob/master/Makefile.am#L420
to rely on that `generated_files` rule instead, but, I don't have to - just adding a single target line as above gives us a rule we can invoke if we just want to generate the files, and the other rules (like `metrics-tests`) continue to work as is.
It looks like I don't even need a PHONY for `generated_files` here either.

So, thoughts - any preference here - happy with the one liner?

jodh-intel:
+1 for the one-liner, but maybe call it `generate-files` (or even maybe just `generate`?)

PR #963, opened 2017-06-08
build: Makefile: Add generate-files rule

grahamwhaley:
Sometimes you just want to generate the generated files without
doing a full build or a further test run (for instance, if we want
to run the tests on an already installed runtime).
Add a rule to allow that.

(turns out we don't need to split the metrics-tests and other rules down,
just add a simple extra rule to nudge the file generation - nice and simple)

Fixes: #945


grahamwhaley:
@MarioCarrilloA you should probably see this - as this is for our use for test runs etc.

chavafg:
qa-passed

jodh-intel:
lgtm

jcvenegas:
lgtm

PR #966, opened 2017-06-08
pod: Support new CRI-O namespaces

### LINKED ISSUE #960: 2.1.9 does not work with the latest CRI-O
CRI-O renamed the pod annotations and thus Clear Containers do not run properly with it anymore.

jcvenegas:
The CRI-O code base went through renaming. All "ocid" instances
were replaced with "cri-o", including the annotation namespaces.
We fixed this and added proper annotations namespaces for CRI-O,
prefixed with io.kubernetes.cri-o.

This patch tries makes our current code base support both the new
and the legacy namespaces.

Fixes #960

Signed-off-by: Samuel Ortiz sameo@linux.intel.com

### DIFF
--- a/src/pod.c
+++ b/src/pod.c
@@ -18,19 +18,6 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-/* Sandbox rootfs */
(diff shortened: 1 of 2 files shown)


sameo:
LGTM

Issue #769, opened 2017-03-23
vm.json defaults don't align with installed cc defaults


egernst:
When I install clear containers via a repo such as [1], the dependent images end up going into /usr/share/clear-containers/ and the vm.json at /usr/share/defaults/cc-oci-runtime point to this accordingly.

Now, after convincing myself that I love clear containers so much that I want to move on and build it from source and run it, I go ahead to a make install and notice that the default updates to vm.json point to a non-existent location, /var/lib/cc-oci-runtime/...

I don't see any rationale one why we'd want to change the default location, especially if it is putting an immediate burden on folks who are trying to build and run as fast as possible.  We should update the default to use the same location, /usr/share/defaults/cc-oci-runtime/

[1] http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora\_25/home:clearlinux:preview:clear-containers-2.1.repo

dlespiau:
Juste a note that, actually, /usr/share/defaults/ isn't a thing outside of clear linux and if we had to package the runtime for other distros, they'd probably make us change that.

Your point still holds thought.

jodh-intel:
Clear Linux overrides the default locations as shown here: https://github.com/clearlinux-pkgs/cc-oci-runtime/blob/master/configure

It's debatable whether these files should live below `/var/lib/` or `/usr/share/`. The Linux FHS would seem to permit both: the latter is "more correct", although the former would allow for a read-only `/usr/`.

egernst:
Just following up because this got me again.  

While it may not be a bug, per se, I think this should be handled in a way that's a bit more kind to users/developers.   How're developers expected to handle this?  Since what the make install updates into vm.json is incorrect (on some distros, at the very least), can we instead of overwriting the installed vm.jsons's path, copy and reuse it?

Ie, make the "make install" a bit smarter in determining what path the vmlinux.container and clear-containers.img is?

dlespiau:
It's still somewhat of a bug I think. We need to decide where those files belong and be consistent across the board.

egernst:
I think it is a bug.  I'm going to open a second issue as well which would solve this.  Basically that we should be able to provide a build option that will build all of the dependent packages.  cc-oci-runtime is just a single binary.  For Clear Containers, we need to also provide mechanism to build from source all of the other components (namely things like container.img and vmlinux/z).   We need to be able to build all the artifacts that are necessary for Clear Containers without necessitating having a login to koji, imo.

PR #978, opened 2017-06-09
configure: Add warning when kernel or image is not found

jcvenegas:
This patch adds a warnings at configuration time, it will
show a warning message when either clear containers kernel
or image are not found.

It also check if a kernel or image was installed by clear containers
packages. If they are installed it will use them as default paths. This
is useful for people that want to use `cc-oci-runtime` compiled from
source but had a previous installation from packages.
Fixes: #769

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -321,6 +321,14 @@ AS_IF([test x"$with_tests_bundle_path" != x""],
 
 #Kernel path for Clear Containers
 DEFAULT_KERNEL_PATH=$localstatedir/lib/$PACKAGE_NAME/data/kernel/vmlinux.container
+
+# Verify container kernel is installed by package.
+OBS_PKG_KERNEL_PATH="/usr/share/clear-containers/vmlinux.container"
+if  [ ! -f "$DEFAULT_KERNEL_PATH" -a -f "$OBS_PKG_KERNEL_PATH" ] ; then
+	DEFAULT_KERNEL_PATH="$OBS_PKG_KERNEL_PATH"
+	AC_MSG_WARN($DEFAULT_KERNEL_PATH detected using as default kernel path)
+fi
+
 AC_ARG_WITH([cc-kernel],
     [AS_HELP_STRING([--with-cc-kernel=[[KERNEL-PATH]]],
         [clear container kernel path])
@@ -334,8 +342,24 @@ AC_ARG_WITH([cc-kernel],
     ]
 )
 
+if test -f "$CONTAINER_KERNEL"; then
+	KERNEL_FOUND=yes
+else
+	KERNEL_HELP="Could not find $CONTAINER_KERNEL kernel file. Use --with-cc-kernel to specify the correct path to container kernel."
+	AC_MSG_WARN([$KERNEL_HELP])
+	KERNEL_FOUND=no
+fi
+
+
 #Container image path for Clear Containers
 DEFAULT_CONTAINER_IMG=$localstatedir/lib/$PACKAGE_NAME/data/image/clear-containers.img
+
+# Verify container image is installed by package.
+OBS_PKG_IMAGE_PATH="/usr/share/clear-containers/clear-containers.img"
+if  [ ! -f "$DEFAULT_CONTAINER_IMG" -a -f "$OBS_PKG_IMAGE_PATH" ] ; then
+	DEFAULT_CONTAINER_IMG="$OBS_PKG_IMAGE_PATH"
+	AC_MSG_WARN($DEFAULT_CONTAINER_IMG detected using as default kernel path)
+fi
 AC_ARG_WITH([cc-image],
     [AS_HELP_STRING([--with-cc-image=[[IMAGE-PATH]]],
         [clear container image path])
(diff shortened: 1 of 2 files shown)


gorozco1:
The developer needs to know his options to get Clear Containers working if the kernel and image was not found. 

```
cc-oci-runtime - 2.1.11

 • General:

       QEMU                     : /usr/bin/qemu-lite-system-x86_64
       QEMU pc-lite             : yes
       Container kernel         : ${prefix}/var/lib/cc-oci-runtime/data/kernel/vmlinux.container   [NOT Found]
                                : Use make install-complete to download and install the kernel in the path defined above
                                  or use --with-cc-kernel to specify the correct path to container kernel.
       Container image          : ${prefix}/var/lib/cc-oci-runtime/data/image/clear-containers.img [NOT Found]
                                : Use make install-complete to download and install the image in the path defined above
                                  or use --with-cc-kernel to specify the correct path to container kernel.
       Container image services : no
       Debug build              : no
              

 • Proxy:

       Systemd unit directory   : /usr/lib/systemd/system

 • Tests:

       Unit tests               : yes
       Functional tests         : yes
       Docker tests             : yes
       CRI-O tests              : yes
       Cppcheck                 : no
       Valgrind                 : yes
       Code coverage            : no
       Bundle path              : /home/gorozco1/repos/clear-containers/cc-oci-runtime/tests/functional/bundle
       Create + delete bundle   : yes
```

chavafg:
qa-failed

chavafg:
qa-passed

configure.ac
```diff
@@ -330,6 +330,9 @@ AC_ARG_WITH([cc-kernel],
         AC_SUBST([CONTAINER_KERNEL], [$DEFAULT_KERNEL_PATH])
     ]
 )
+AC_CHECK_FILE("$CONTAINER_KERNEL",,
+	      AC_MSG_WARN([Could not find $CONTAINER_KERNEL kernel file Use --with-cc-kernel to specify the correct path to container kernel.]))
```
jodh-intel: s/file Use/file. Use/g

configure.ac
```diff
@@ -345,6 +348,8 @@ AC_ARG_WITH([cc-image],
         AC_SUBST([CONTAINERS_IMG],[$DEFAULT_CONTAINER_IMG])
     ]
 )
+AC_CHECK_FILE("$CONTAINERS_IMG",,
+	      AC_MSG_WARN([Could not find $CONTAINERS_IMG image file Use --with-cc-image to specify the correct path to container image.]))
```
jodh-intel: s/file Use/file. Use/g

egernst:
Aside from James' comments, I think this looks good.  This will be a major step up for usability.  It looks like the installation path will be matching what is used in the packages right?

Did you test to make sure that the output of make install (what exists today) provides a vm.json which aligns with the paths that this installs to?

jcvenegas:
@egernst I wont take use the paths we use in our packages today. In my opinion that is a distribution choose. 
It will check in `var/.../...` but fail if the report it if it does not exist. As far as I know #979 will install in the path in `var/...` . 
1 - So an user will try to build cc-oci-runtime. 
2 - he will be notified that the kernel is not installed will be given 2 options:
  a. Use #979 to install kernel an image.
  b. Use `--with-cc-kernel` `--with-cc-image`

Given `/usr/share/clearcontainers` is being used as a common directory we could verify if the kernel and image are installed in that path as well. That would simplify if the user has the clear containers components installed but want to build and use  a new runtime.

@jodh-intel  @gorozco1 comments?

egernst:
@dlespiau  @sameo  ^ thoughts on above?

gorozco1:
@jcvenegas @egernst I agree with @jcvenegas the paths on a package are pretty much a linux distribution desition. 

However when installing from source it is known that everything or almost everything will be placed under ``/usr/local``


Any thoughts @jodh-intel ?

gorozco1:
and yes #979 will add options to install kernel and image as well on a given path defined on ``--with-cc-kernel``  and ``--with-cc-image`` or use the defaults.

In this case ``/var/lib/clearcontainers``


We can argue also the posibility to change the default to `${prefix}/usr/share/clearcontainers`

jodh-intel:
I'm slightly confused about the above. I think (?) we're talking about *not* defaulting to the path values we currently use in OBS and clr (https://github.com/clearlinux-pkgs/cc-oci-runtime/blob/master/configure)?

But I'm certainly happy for us to continue to have `${prefix}` as `/usr/local` and to store kernels + images below `${prefix}/var/lib/cc-oci-runtime/*`.

chavafg:
qa-passed

jcvenegas:
@jodh-intel  I think Eric's main concern is when you want to compile and install your own installed version with OBS packages. The OBS packages provide kernel and image in `/usr/share` then when the cc-oci-runtime is configured `vm.json` it will be configured to use `/var`.  So what it is wanted is an easy way to not use `--with-cc-kernel`.

So I think we can check the image and kernel are not installed in obs paths ( not because is a default path but a common path used in our packaging). If it exist we use those paths in `vm.json`.

jodh-intel:
Hi @jcvenegas - thanks for explaining. Sounds good.

jcvenegas:
@gorozco1  @egernst  @jodh-intel  can you review this PR again
updated to detect kernel and image installed by packages and use it as default path.
We still honor configure option --with-cc-kernel --with-cc-image

chavafg:
qa-failed

chavafg:
qa-failed

jodh-intel:
lgtm

jcvenegas:
@chavafg  can we re-test this PR?

chavafg:
qa-passed

chavafg:
qa-passed

Issue #912, opened 2017-05-16
Need to be able to build and install Clear Containers from source


egernst:
If a user  pulls down cc-oci-runtime and does a make install, Clear Containers will not work.  It is assumed that binary artifacts area already installed.

We need to create an option to build all of the relevant parts for Clear Containers from source.  This build will initially take a long time, but without this usability for Clear Containers is very poor.  We need to allow developers to be able to rebuild a given release in its entirety from source, including things like the kernel and container image.

This will solve problems like: https://github.com/01org/cc-oci-runtime/issues/769

gorozco1:
Building the kernel and container image from source is out of scope to fix this issue, what we are going to do is to provide the user in the doc the dependencies needed to run clear-containers. 

We'll provide the scripts to do so, and will configure will fail if the "dependencies" are not meet.

egernst:
Ultimately, people will need the ability to easily modify the kernel as well.  I can call that a P3, perhaps, though.

Having said that, I think it's still a high priority to added a make option called something like "make install-complete" which would do the pull and installation of .img/kernel.

gorozco1:
@egernst  agreed 👍 

just created two other items to build from source the components.

jcvenegas:
@egernst  what about qemu-lite this is something we should not care in `make install-complete`

PR #979, opened 2017-06-09
build: Adds option to install kernel and image

gorozco1:
If a user pulls down cc-oci-runtime and does a make install, Clear
Containers will not work. It is assumed that binary artifacts are
already installed.

This adds `make artifacts` as an option to download and install `clear-containers-kernel` and `clear-containers-image.`

Fixes: #912

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -818,6 +818,26 @@ endif
 
 check: $(CHECK_DEPS)
 
+artifacts: | install-image install-kernel
+
+# Get clear containers image path
+IMAGE_PATH := $(shell dirname $(CONTAINERS_IMG))
+install-image:
+	@if [ $(shell id -u) -ne 0 ]; then \
+		$(srcdir)/installation/install_clear_image.sh $(CLEAR_VM_IMAGE_VERSION) $(IMAGE_PATH) ; \
+	else \
+		>&2 echo "ERROR: do not run $@ as root" ; \
+	fi
+
+# Get clear containers kernel path
+KERNEL_PATH := $(shell dirname $(CONTAINER_KERNEL))
+install-kernel:
+	@if [ $(shell id -u) -ne 0 ]; then \
+		$(srcdir)/installation/install_clear_kernel.sh $(CLEAR_VM_IMAGE_VERSION) $(CLEAR_VM_KERNEL_VERSION) $(KERNEL_PATH) ; \
+	else \
+		>&2 echo "ERROR: do not run $@ as root" ; \
+	fi
+
 clean-local:
 	rm -f commit_id
 if AUTOGOPATH
--- a/README.rst
+++ b/README.rst
@@ -15,7 +15,7 @@ Overview
 --------
 
 ``cc-oci-runtime`` is an Open Containers Initiative (OCI_) "runtime"
-that launches an Intel_ VT-x secured Clear Containers 2.0 hypervisor,
+that launches an Intel_ VT-x secured Clear Containers 2.1 hypervisor,
 rather than a standard Linux container. It leverages the highly
 optimised `Clear Linux`_ technology to achieve this goal.
 
@@ -214,6 +214,16 @@ If you have specific requirements, run::
 ... then add the extra "``configure``" flags you want to use::
 
   $ ./autogen.sh --enable-foo --disable-bar && make
+
+Artifacts
+---------
+
+- clear-containers-image
+- clear-containers-kernel
+
+To download and install artifacts for Clear Containers, run::
+
+  $ make artifacts
 
 Tests
 -----
--- a/configure.ac
+++ b/configure.ac
@@ -155,6 +155,9 @@ AM_CONDITIONAL([CRIO_TESTS], [test x"$run_crio_tests" = x"yes"])
 # Checks for libraries.
 source $srcdir/versions.txt
 
+AC_SUBST([CLEAR_VM_IMAGE_VERSION], [$clear_vm_image_version])
+AC_SUBST([CLEAR_VM_KERNEL_VERSION], [$clear_vm_kernel_version])
+
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= "$glib_version"], have_required_glib=yes, have_required_glib=no)
 AS_IF([test "$have_required_glib" != "yes"], AC_MSG_ERROR([ERROR: need glib ${glib_version}+ for the `G_FILE_MONITOR_WATCH_MOVES' flag]))
 
--- a/installation/install_clear_image.sh
+++ b/installation/install_clear_image.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+
+set -x
+set -e
+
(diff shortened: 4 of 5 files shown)


chavafg:
qa-failed

chavafg:
qa-failed

chavafg:
qa-passed

Makefile.am
```diff
@@ -813,6 +813,20 @@ endif
 
 check: $(CHECK_DEPS)
 
+install-all: install-kernel |  install-image |  install
+
+# Get clear containers image path
+IMAGE_PATH := $(shell dirname $(CONTAINERS_IMG))
+install-image:
+	curl https://raw.githubusercontent.com/clearcontainers/tests/master/.ci/install_clear_image.sh \
```
jodh-intel: I like the simplicity of this code, but this is a potential security risk; if the user is running `sudo make`, they are executing arbitrary shell code downloaded from a remote site as the superuser (`install_clear_kernel.sh` calls `rm -r` too ;)

I'm wondering if it makes sense to rework the two scripts in the `tests` repo slightly so that they have to be passed a parameter (`download`, or `install` or `show-install-locations`). That way, this `Makefile` could:

- Download the two scripts.
- Add a couple of rules (`download-kernel` and `download-image`).
- Ensure these two rules run as a non-`root` user:
  ```
  download-kernel:
  ifeq ($(shell id -u), 0)
    $(error ERROR: do not run this rule as root!)
  endif
  ...
  ```
- Make the new rules call the scripts with the `download` parameter.
- Change the `install-*` rules to run each script with the `show-install-locations` parameter. This would show 1 or more lines in the following format (which would map directory to `install(1)` arguments):
  ```
  <from> <to> <user> <group> <perms>
  ```
   ... where
   - `<from>` is the full path of the file to install,
  - `<to>` is the full path of the directory to install the file to.
- This `Makefile` could then call `install` itself (using `sudo`). This would be the only phase run as `root` so would limit the possibility of problems.

What do others think?

grahamwhaley: I think reducing the necessity to run as `root` to a minimum makes sense as a general plan.

Running a dynamically downloaded script as root will raise a security flag, so we should probably address it. I will note that I think the 'level of trust' of that shell script is currently probably the same as that of the `Makefile` that is downloading and running it - they both come from github repos managed by the same group - but, reducing the attack surface is always a good thing.

egernst:
This uses the same path that is used in the packaging setup ci per distro, correct?

Can you verify this works appropriately when you do a make install-kernel and then a make install?  I want to make sure that the vm.json paths after the second install aligns with the path that is used in install-kernel.  Thanks @gorozco1 , this patch helps our project become a lot more usable.

gorozco1:
@egernst yes it works...  However i need to address @jodh-intel suggested changes first.

Makefile.am
gorozco1: Well the reason I did this was to avoid the duplication of the code, and I wanted to get feedback from the devs of this project first. 

The main problem is that yes, at some point the scripts downloaded from the other repo need root to install the artifacts ``kernel`` and ``image`` needed to run clear containers.

IMHO the best option is to  adapt the scripts  to our own needs with some minnor tweaks. And add them to our repo. After all, the scripts in the other repo pursue a different goal. 

The feedback @jodh-intel gave is great but I think we are overcomplicating things with a fix that can be simple as this.

chavafg:
qa-passed

chavafg:
qa-passed

gorozco1:
@jodh-intel updated with what I think is the best approach.

Makefile.am
```diff
@@ -813,6 +813,18 @@ endif
 
 check: $(CHECK_DEPS)
 
+install-all: install-kernel |  install-image |  install
+
+# Get clear containers image path
+IMAGE_PATH := $(shell dirname $(CONTAINERS_IMG))
+install-image:
+	./installation/install_clear_image.sh $(IMAGE_PATH)
```
jcvenegas: $(srcdir)/installation/install_clear_image.sh $(IMAGE_PATH)

Makefile.am
```diff
@@ -813,6 +813,18 @@ endif
...
+
+# Get clear containers kernel path
+KERNEL_PATH := $(shell dirname $(CONTAINER_KERNEL))
+install-kernel:
+	./installation/install_clear_kernel.sh $(KERNEL_PATH)
```
jcvenegas: $(srcdir)/installation/install_clear_kernel.sh $(KERNEL_PATH)

jcvenegas:
@egernst `This uses the same path that is used in the packaging setup ci per distro, correct? `
It does not use the paths to used in obs packages.

gorozco1:
@egernst  @jcvenegas This configuration uses the path defined by the user at the compilation time or the defaults if the user did not specify other.

gorozco1:
@jodh-intel @jcvenegas updated again take a look please

chavafg:
qa-failed

chavafg:
qa-failed

gorozco1:
@chavafg could you run the tests again please?

chavafg:
qa-passed

Makefile.am
```diff
@@ -813,6 +813,18 @@ endif
 
 check: $(CHECK_DEPS)
 
+install-all: install-kernel |  install-image |  install
```
jodh-intel: Shouldn't this be:

```
install-all: | install-kernel install-image install
```

Makefile.am
```diff
@@ -813,6 +813,18 @@ endif
...
+
+# Get clear containers image path
+IMAGE_PATH := $(shell dirname $(CONTAINERS_IMG))
+install-image:
+	$(srcdir)/installation/install_clear_image.sh $(CLEAR_VM_IMAGE_VERSION) $(IMAGE_PATH)
```
jodh-intel: This is still potentially going to run `curl` as root if the user is running `sudo make`.

Since the scripts call `sudo` when needed (good), we should be able to stop `install-kernel` and `install-image` running as `root` with something like:

```
ifeq ($(shell id -u), 0)
running_as_root=1
endif

artifacts: | install-image install-kernel

install-kernel:
ifneq ($(running_as_root),)
	$(error ERROR: do not run $@ as root)
endif
	# ...

install-image:
ifneq ($(running_as_root),)
	$(error ERROR: do not run $@ as root)
endif
	# ...
```

This does mean that `sudo make install` won't do everything - we'll have to add a new rule for downloading and installing the artifacts, but the user will then run:

```
make artifacts && sudo make install
```

... which doesn't seem to arduous. We'll need to document how to do this of course.

gorozco1: oh ok, so you said two steps to instal the kernel and the image? one for just downloading other for installing (last one will be run as root)

jodh-intel: I think that would be safer. Any thoughts @grahamwhaley, @jcvenegas ?

gorozco1: @egernst

jcvenegas: sounds good, they are more steps but safer, how this is handled in cc3.0?

jodh-intel: It isn't handled at all in CC 3.0 yet as the `Makefile` currently only installs the runtime binary and config file.

gorozco1:
@jodh-intel @egernst @jcvenegas Updated commit could you review?

chavafg:
qa-passed

chavafg:
qa-passed

chavafg:
qa-passed

Makefile.am
gorozco1: done

Makefile.am
gorozco1: done

Makefile.am
```diff
@@ -818,6 +818,26 @@ endif
 
 check: $(CHECK_DEPS)
 
+artifacts: | install-image install-kernel
+
+# Get clear containers image path
+IMAGE_PATH := $(shell dirname $(CONTAINERS_IMG))
+install-image:
+	@if [ $(shell id -u) != 0 ]; then \
```
jodh-intel: Nit: this is using the string comparison operator rather than the numeric (`-ne`), but it still works so not really a problem.

jodh-intel:
lgtm

Makefile.am
gorozco1: updated @jodh-intel

jodh-intel:
@gorozco1 - thanks!

lgtm

chavafg:
qa-passed

chavafg:
qa-passed

gorozco1:
@jcvenegas changes made can you review this PR?

jcvenegas:
@gorozco1 looks good for me

chavafg:
qa-passed

jodh-intel:
Hi @iphutch - could you take a look at the doc files on this PR?

PR #996, opened 2017-06-17
proxy: Better error message for too long socket path

### LINKED ISSUE #995: Proxy does not provide useful information when socket path is too long
When `cc-proxy` is used with the option `-socket-path` and the path is too long it only say 

```
init: couldn't create AF_UNIX socket: listen unix $LONG_PATH bind: invalid argument
```
In linux the max length of a  unix socket path is 108 including null character (http://man7.org/linux/man-pages/man7/unix.7.html). 

The message  `bind: invalid argument` is not obvious.

jcvenegas:
This verify the socket path length according to http://man7.org/linux/man-pages/man7/unix.7.html.

The proxy fails correctly when the socket path is too long, but the message is too generic: `bind: invalid argument`.

Fixes #995

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -32,6 +32,10 @@ import (
 	"github.com/golang/glog"
 )
 
+// In linux the max socket path is 108 including null character
+// see http://man7.org/linux/man-pages/man7/unix.7.html
+const socketPathMaxLength = 107
+
 // Main struct holding the proxy state
 type proxy struct {
 	// Protect concurrent accesses from separate client goroutines to this
@@ -308,6 +312,11 @@ func (proxy *proxy) init() error {
 		socketPath := DefaultSocketPath
 		if len(*ArgSocketPath) != 0 {
 			socketPath = *ArgSocketPath
+		}
+
+		if len(socketPath) > socketPathMaxLength {
+			return fmt.Errorf("socket path too long: %d (max %d)",
+				len(socketPath), socketPathMaxLength)
 		}
 
 		socketDir := filepath.Dir(socketPath)


proxy/proxy.go
```diff
@@ -32,6 +32,10 @@ import (
 	"github.com/golang/glog"
 )
 
+// In linux the max socket path is 108 inclding null character
```
dlespiau: including

proxy/proxy.go
```diff
@@ -310,6 +314,11 @@ func (proxy *proxy) init() error {
 			socketPath = *ArgSocketPath
 		}
 
+		if len(socketPath) > socketPathMaxLength {
+			return fmt.Errorf("Socket path too long(length is %d): max length is %d ",
```
dlespiau: Every other error message starts with a lower case character :)

dlespiau: "socket path too long: %d (max %d)"

chavafg:
qa-passed

proxy/proxy.go
jodh-intel: Could you add a space after `long` in: `too long(length` ?

jodh-intel:
Hi @jcvenegas - good catch! Could you also raise a PR to fix the same issue in:

- https://github.com/clearcontainers/proxy/blob/master/proxy.go#L554..L569 (that function should probably return a `string` *and* an `error`).

dlespiau:
(Of course, I had already opened the corresponding bug https://github.com/clearcontainers/proxy/issues/78)

jcvenegas:
@dlespiau PR updated , 
@jodh-intel @dlespiau  sure will send to `proxy` repository as well.

jodh-intel:
lgtm

dlespiau:
lgtm

chavafg:
qa-failed

dlespiau:
@chavafg anything scary in the CI logs?

chavafg:
@dlespiau actually no, not sure why it send the failed state. I am checking my CI as I added some changes before. This should be ok

dlespiau:
Thanks!

PR #999, opened 2017-06-20
cc-oci-runtime.sh: Enable debug by default

sameo:
We can assume that most of the time using
cc-oci-runtime.sh means we're looking for some debug information.


chavafg:
qa-passed

egernst:
should we enable the persistent log in the case of .sh as well?

jodh-intel:
This PR enables the global log, which is the persistent log.

lgtm.

dlespiau:
lgtm

PR #1012, opened 2017-06-26
tests: Use nginx:1.13.0 for swarm tests

chavafg:
Newest nginx docker image (1.13.1) does not contain `ip`
command by default, which we use in our swarm tests. This
commit specifies to use 1.13.0 which contains `ip`.

### DIFF
--- a/tests/integration/docker/dns.bats
+++ b/tests/integration/docker/dns.bats
@@ -54,14 +54,14 @@ setup() {
 	$DOCKER_EXE service create \
 	--name "${SERVICE1_NAME}" \
 	--replicas $number_of_replicas \
-	--publish 8080:80 nginx /bin/bash \
+	--publish 8080:80 "${nginx_image}" /bin/bash \
 	-c "hostname > /usr/share/nginx/html/hostname; nginx -g \"daemon off;\""
 
 	info "create service ${SERVICE2_NAME}"
 	$DOCKER_EXE service create \
 	--name "${SERVICE2_NAME}" \
 	--replicas $number_of_replicas \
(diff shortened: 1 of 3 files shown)


grahamwhaley:
This feels like an interim fix to me? I don't think we can stick to 1.13.0 for ever. Do we have a plan to make our own Dockerfile/image overlayed with `ip` on top of the standard nginx one perhaps?

chavafg:
Well, we can create a custom image for that, but I remember @jodh-intel on a PR saying that we should work with upstream images, instead of creating and maintaining our own.  But here maybe is not an option.
I'll create another issue to track that and in the meanwhile I would like to have this merged so the CI comes back to normal.

chavafg:
qa-passed

jcvenegas:
lgtm

chavafg:
@grahamwhaley I have created #1014 for creating our own image.

grahamwhaley:
Great, thanks. I think making our own image that builds on the latest upstream is acceptable - we are still 'using the upstream image', just adding extras to it as we need. Thus, for this:
lgtm

jodh-intel:
I agree with @grahamwhaley here: if we use a hard-coded version, we'll forget about it until bad things happen in the future. Yes, I think we need an issue to investigate an overlay as Graham suggests. Ideally, that issue would be referenced in this PR in file `test-common.bash.in`, explaining why we are using a fixed version.

Do we know why the `ip` command was removed in the upstream image? Presumably by design and not by accident?

chavafg:
qa-passed

chavafg:
Added comment in `test-common.bash.in`

PR #1035, opened 2017-07-06
versions: Update kernel and image versions

### LINKED ISSUE #1034: Update ```versions.txt``` with package versions used by the release and the cc-oci-runtime.
The kernel had a minor update last week lest update versions.txt before release #1033

jcvenegas:
Update kernel to version 4.9.35.

The Clear Containers image is updated as well to sync the version where the kernel was updated:

https://download.clearlinux.org/releases/16180/clear/RELEASENOTES

Fixes #1034


grahamwhaley:
in the diff you update the kernel and the image version - just to check, are the two intrinsically linked, or is this really a dual update (in which case we should in put that in the subject and commit to make it clear :-) )

chavafg:
qa-passed

jcvenegas:
@grahamwhaley we update both in order to reach easily the kernel that was released in an specif clear linux version, so yes they are linked, but I think I can improve the commit message

jcvenegas:
@grahamwhaley commit message was changed, please take a look.

chavafg:
qa-passed

miguelinux:
+1

